PR #56, opened 2015-01-05
Add getUptimeMillis() to the BrokerView MBean

PaulGale:
I need to be able to get the up-time of the broker in milliseconds, rather than as a formatted string.


hzbarcea:
Do you mind opening a jira issue?

PaulGale:
Done. https://issues.apache.org/jira/browse/AMQ-5505

PR #59, opened 2015-02-11
https://issues.apache.org/jira/browse/AMQ-4848 - Karaf update added Transport Connectors...

zmhassan:
This was a fix that I sent pull request #28 which was from Jun 17, 2014. I fixed activemq:list command and it now displays the transport connectors available for the broker on karaf. Can anyone review this patch. If the fix is good then can this be merged? 

screen shot 2015-02-11 at 4 14 05 pm 2

Also I've updated the pax exam unit test so the integration test does not fail.

screen shot 2015-02-11 at 4 26 39 pm 2


davsclaus:
Its IMHO better to create a new command to list the transport connectors, as the existing is for listing only brokers.

PR #62, opened 2015-02-17
AMQ-4598 purge enhancement

metatechbe:
Reset queue counter during "purge" to the number of inflight messages, so that they are not "hidden" while still being processed.


tabish121:
This issue was fixed in the upstream, mind closing the PR ?

PR #63, opened 2015-02-25
revise set and get methods name of Activation Spec properties "Redeliver...

soul2zimate:
...yBackOffMultiplier" and "RedeliveryUseExponentialBackOff" to allow configuration through @ActivationConfigProperty.

https://issues.apache.org/jira/browse/AMQ-5613

Activation Spec properties[1] "RedeliveryBackOffMultiplier" and "RedeliveryUseExponentialBackOff" can not be used within @ActivationConfigProperty, in contrast, it accepts "backOffMultiplier" and "useExponentialBackOff".

[1] http://activemq.apache.org/resource-adapter-properties.html

### DIFF
--- a/activemq-ra/src/main/java/org/apache/activemq/ra/ActiveMQActivationSpec.java
+++ b/activemq-ra/src/main/java/org/apache/activemq/ra/ActiveMQActivationSpec.java
@@ -593,7 +593,15 @@ public class ActiveMQActivationSpec implements MessageActivationSpec, Serializab
         }
     }
 
+    @Deprecated
     public double getBackOffMultiplier() {
+        if (redeliveryPolicy == null) {
+            return 0;
+        }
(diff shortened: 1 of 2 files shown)


davsclaus:
This is an API breaking. Is this good to do? @gtully @tabish121 

Is it maybe the documentation that is wrong and need to be changed?

tabish121:
the documented names probably make more sense given they have the most context into what is being configured, unfortunately the code is expecting the stripped down options.  Changing it now would indeed break existing users.  Perhaps addition of methods that match the documented values while retaining the older ones is an option

davsclaus:
Thanks @tabish121.

Yeah we could add new methods and mark the old ones as deprecated and they can use the new methods.

soul2zimate:
rebased as per proposed change.

lucastetreault:

cshannon:
Closing stale PR

PR #64, opened 2015-02-25
Fix to only include the activemq opts for memory when start is called.  ...

burtonator:
...If we don't do this we end up potentially allocating too much memory if -Xms is a high value.


tbain98:
Kevin, this is a good start but it still uses the memory options when you invoke the status target.  Can we update the script to avoid using the memory options in that case?

PR #69, opened 2015-03-09
AMQ-5645 - Updated DayOfMonth cron to roll to next month properly.

mwarhaftig:
For the Minutes and Hours tokens in CronParser.buildCronEntries() the CronEntry.end value was correctly +1 of actually allowed end values (ex. a time of 10:60 is not possible).  Updated DayOfMonth to also be +1 (32 instead of 31).  Expanded unit tests included.


tabish121:
This issue was resolved at Apache and can be closed.

tabish121:
Mind closing this one out, changes have been applied upstream

PR #71, opened 2015-03-12
AMQ-5659 Safety measure against infinite loop in "purge"

metatechbe:


metatechbe:
Obsoleted by submitted patch V2

PR #73, opened 2015-03-15
AMQ-5662 Improve web console retry to support all DLQ names

northlander:
https://issues.apache.org/jira/browse/AMQ-5662

Retry link in webconsole is changed from using move and pattern based DLQ identification to use AMQ built in DLQ query and retry mechanism.


northlander:
Replaced by #176

PR #74, opened 2015-03-16
AMQ-5082 unit test and patch

jimrobinson:
A proposed unit test and patch to address the issues raised in AMQ-5082 (the quorum never recovered after a broker's zookeeper session expires)


christian-posta:
I've cleaned up the PR history, applied with thanks here: https://github.com/apache/activemq/commit/a39e51e0519852332f7779443c3db09b6e691d49

Can close this PR.

PR #75, opened 2015-03-17
Lichtin patch 1

lichtin:
AMQ-5669: Update Karaf feature files to add required namespace

### DIFF
--- a/activemq-karaf/src/main/resources/features-core.xml
+++ b/activemq-karaf/src/main/resources/features-core.xml
@@ -18,7 +18,7 @@
 
 -->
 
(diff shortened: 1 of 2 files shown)


davsclaus:
@jbonofre is there a Karaf version which requires to use that xmlns as in this PR?

The code is using karaf 4 for testing, and its maybe okay to assume its karaf 4.0.x onwards for 5.15 release.

hzbarcea:
@lichtin good patch, too bad it wasn't applied at the right time. I updated the code using the proper newer versions in the targetNamespace. You get the credit for the contribution though. Many thanks!

PR #80, opened 2015-04-02
Ensure JMX is started on XBeanBrokerService

mtaylor:
Using the XBeanBrokerService with JMX and Scheduler Support enabled.
The ManagementContext is never started when in slave mode, since
afterPropertiesSet() blocks trying to get the scheduler shared file
lock.  This patch updates the XBeanBrokerService to start the
BrokerService ManagementContext (when isJMX==true) before trying
to get the job scheduler shared file lock..

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -2487,11 +2487,14 @@ public class BrokerService implements Service {
     }
 
     protected void startManagementContext() throws Exception {
-        getManagementContext().setBrokerName(brokerName);
-        getManagementContext().start();
-        adminView = new BrokerView(this, null);
-        ObjectName objectName = getBrokerObjectName();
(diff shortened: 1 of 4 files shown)


gtully:
thanks for the contribution, great to see a test :-) however I removed the mods to afterPropertiesSet and the test still works!
One other thought  - the call to startManagementContext needs to be gated with shouldAutostart,
there are use cases where a spring context creates the broker bean for later start up.: 

if (shouldAutostart() && isUseJmx() && !getManagementContext().isStarted()) {
      startManagementContext();
}

mtaylor:
@gtully Somehow I removed the key statement from the test when I was cleaning up:

```java
  broker.setSchedulerSupport(true); 
```

It is actually blocking during jobScheduler.start(), whilst the scheduler is trying to get hold of shared file lock.

SharedFileLocker.doStart() line 52:

```java
  while ((!isStopped()) && (!isStopping())) {
```

I'll add back in this statement and make the suggested change.  I wonder though, whether I've added the fix in the correct place?

PR #84, opened 2015-04-08
Modified StoreQueueCursor to prevent a broker deadlock

cshannon:
Modified StoreQueueCursor to properly implement tryAddMessageLast so that the message lock can be released during a timeout when the temporary store is full.  Updated Queue to call tryAddMessageLast instead of addMessageLast in the cursorAdd method.

    This resolves https://issues.apache.org/jira/browse/AMQ-5712


cshannon:
I just updated this request to include a unit test showing the deadlock occurring when the memory store size is larger than the temporary store size.

cshannon:
Closing this as Tim Bish has posted a better version of this patch.

PR #85, opened 2015-04-11
Add support for suffix destination filtering.

jwcarman:


activemq-client/src/main/java/org/apache/activemq/filter/SuffixDestinationFilter.java
```diff
@@ -0,0 +1,80 @@
...
+
+
+/**
+ * Matches messages which match a suffix like "<.A.B"
+ *
+ *
+ */
+public class SuffixDestinationFilter extends DestinationFilter {
+
+    private String[] suffixes;
```
peteryhwong: Can these not be final?

jwcarman: Yes. I was mimicking the existing code I believe

hzbarcea:
Wow, 5.11... @jwcarman do you still want this patch applied? If so I'll try to test and port as necessary. Please say no :).

jwcarman:
@hzbarcea well, if it hasn't been merged by now (2.5 years later), I don't guess I really need it at all. :)

hzbarcea:
@jwcarman fair enough. If you want us to be more responsive next time ping me. You know how. Happy Holidays!

PR #88, opened 2015-04-15
Do not log passwords on MBean method calls.

mtaylor:
Previous to this patch the AnnotatedMBean class would simply dump
any arguments passed in via JMX call to the log (when audit is enabled).
Method parameters can sometimes contain sensitive information such as
the password field on QueueView.sendTextMessage.

This patch adds a @Sensitive annotation to the JMX module allowing
implementations of MBean interfaces to mark method parameters as sensitive
preventing values from being logged.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/AnnotatedMBean.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/AnnotatedMBean.java
@@ -191,10 +191,38 @@ public class AnnotatedMBean extends StandardMBean {
             entry.setUser(caller);
             entry.setTimestamp(System.currentTimeMillis());
             entry.setOperation(this.getMBeanInfo().getClassName() + "." + s);
-            entry.getParameters().put("arguments", objects);
+
+            try
+            {
+               if (objects.length == strings.length)
+               {
+                  Method m = getMBeanMethod(this.getImplementationClass(), s, strings);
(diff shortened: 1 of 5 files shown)


gtully:
like it, nice patch and test. I guess it would be good to track this enhancement/improvement in a jira entry so that it will make its way into the release notes for the next release.

mtaylor:
Created JIRA here: https://issues.apache.org/jira/browse/AMQ-5729 and updated the Commit message to suit.

Merging.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/AnnotatedMBean.java
```diff
@@ -191,10 +191,38 @@ public Object invoke(String s, Object[] objects, String[] strings) throws MBeanE
             entry.setUser(caller);
             entry.setTimestamp(System.currentTimeMillis());
             entry.setOperation(this.getMBeanInfo().getClassName() + "." + s);
-            entry.getParameters().put("arguments", objects);
+
+            try
+            {
+               if (objects.length == strings.length)
```
peteryhwong: Should we ensure the brackets conform to Egyptian Brackets as the style used by the rest of the source file?

hzbarcea:
Merged by @mtaylor a long time ago. Thanks for the patch. Closing.

PR #89, opened 2015-04-15
Set eclipse plugin version in leveldb-store pom

mtaylor:


gtully:
Think you should reference the root pom -  <version>${maven-eclipse-plugin-version}</version>

mtaylor:
This seems to be working fine now in IntelliJ 14.  I suspect the version issue was environmental so closing this.  We can reopen later should others experience the same.

PR #95, opened 2015-04-29
Add use case with 2 consumers where one is exclusive

cmoulliard:
Example where we start 2 consumers where one is exclusive and not the second and we control/check that all the messages are consumed by the exclusive


hzbarcea:
@cmoulliard this seems very old. Do you want to keep it or close it?

PR #97, opened 2015-05-01
Adding a configuration option set max destinations on the broker

cshannon:
Adding a configuration option to PolicyEntry to enable setting the maximum number of created destinations by policy on the broker.

This resolves https://issues.apache.org/jira/browse/AMQ-5751


cshannon:
I just updated the pull request slightly to make sure the validation only happens if the destination doesn't exist and I fixed formatting (replaced tabs with spaces since that seems to be the standard)

PR #99, opened 2015-05-18
Adding maxFrameSize to the Stomp Protocol

cshannon:
Adding maxFrameSize to the Stomp Protocol so that a maximum frame size can be validated to be consistent with OpenWire.

This applies to https://issues.apache.org/jira/browse/AMQ-5776


cshannon:
Updated pull request based on feedback on Jira.

cshannon:
This pull request was merged in so I am closing it as complete.

PR #101, opened 2015-05-20
Fixing missing volatile on references in VMTransport

cshannon:
Fixing missing volatile on references in VMTransport to prevent a synchronization bug.

This resolves https://issues.apache.org/jira/browse/AMQ-5787


cshannon:
I'm closing this request because it was merged into master.

PR #109, opened 2015-06-08
Renames env script to avoid possible POSIX env shadowing

bric3:
Got an issue with other shell scripts, turns out that when one prepends `$PATH` by `activemq/bin`, `env` is shadowed by ActiveMQ's `env`.

That means that script that want to run commands in customized environment will simple don't be executed like : 

``` sh
env A=aaa B=bbb mv ...
```

Worse such lines won't fail, because this will return the `0` exit status. The workaround is simple, i.e. append in `%PATH` instead of prepend. But a proper fix is needed. This PR proposes to rename the shell script from `env` to `activemq-env`.


bric3:
There's other executable that may be renamed, like `diag`, I can extends this PR if this makes sense.

bric3:
Another matter why this PR should be accepted, `env` is actually part of the [POSIX](http://www.unix.org/) specification : http://pubs.opengroup.org/onlinepubs/9699919799/utilities/env.html

Dridi:
While I agree that `env` should not be shadowed, you probably don't need it. Your example works fine without env:

```bash
A=aaa B=bbb mv ...
```

The `env` command is more useful when you need :
- to print your shell script/prompt environment (`env`)
- to run a command with a clean environment (`env -i A=aaa B=bbb mv ...`)
- an absolute path for a command, like in scripts shebangs (`#!/usr/bin/env python`)

bric3:
@Dridi The problem is it's a POSIX command. And any script ha a right rely on that command. It's even about scripts I own, it actually broke scripts written by others. And they shouldn't have to rewrote them because of this.

Dridi:
@bric3 As I said, I agree with you. I highlighted the fact that you don't need `env` in the example you give, hoping that would help you until it's fixed in activemq.

bric3:
@Dridi Yeah sure, but still

scoopex:
Does this happen because the file "env" has exec permissions and the activemq bin folder is part of the PATH variable?

bric3:
@scoopex Yes, just like any developer add `path/to/maven/bin` to `$PATH`

scoopex:
If "activemmq/bin/env" is not an executable, you should not have the problem.
("env" is just a configuration for the initscript - so it might be better to move it to the "conf" folder)

bric3:
> If "activemmq/bin/env" is not an executable, you should not have the problem.

It is executable otherwise I wouldn't have encountered the issue, agreed that is another solution, but with that solution the problem can come back in the future, any developer of activemq may want to `chmod +x` the `bin` folder.

lucastetreault:

jbonofre:
Superseeded by #1074 1074

PR #113, opened 2015-06-12
https://issues.apache.org/jira/browse/AMQ-5837

cshannon:
This commit adds an Inflight message size statistic to SubscriptionStatistics so we can know the size of all the messages that are inflight, besides just the count.


johnou:
@cshannon unfortunately the current implementation of this feature simply doesn't scale, with a large number of consumers and producers the broker is prone to running out of memory. See https://issues.apache.org/jira/browse/AMQ-6940

PR #114, opened 2015-06-14
https://issues.apache.org/jira/browse/AMQ-5265

Hauenstein:
This is a revision of AMQ-5265 -- the original fix didn't quite clean up all of the mbeans that MBeanBridgeDestination created, as [noted in the original issue by Tarek Hammoud](https://issues.apache.org/jira/browse/AMQ-5265?focusedCommentId=14527137&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-14527137).  The key for the map used to track the ObjectNames was an ActiveMQDestination (which uses the destination's physical name for its equals() implementation) caused the inbound and outbound entries in the map to randomly replace each other.

Like Tarek, I could reproduce this issue using 5.11.1 with two networked brokers (full duplex) with a client publishing to a topic.  After I restart a broker, the logs on the non-restarted broker show the MBeanBridgeDestination reporting many InstanceAlreadyExistsExceptions when trying to register the unintentionally-duplicated mbeans.

Please let me know if this makes sense or if there are any changes you would suggest (ex: is there a reasonable way or an example I could follow to add a unit test for something like this?).  Thanks!

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/PublishedAddressPolicy.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/PublishedAddressPolicy.java
@@ -22,6 +22,7 @@ import java.net.UnknownHostException;
 import java.util.HashMap;
 import java.util.Locale;
 
+import org.apache.activemq.transport.vm.VMTransport;
 import org.apache.activemq.util.InetAddressUtil;
 
 /**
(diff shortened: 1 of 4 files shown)


gtully:
peek at https://github.com/apache/activemq/blob/138e52b08c2f49b730817932a6e63f2a135854f1/activemq-unit-tests/src/test/java/org/apache/activemq/network/DuplexNetworkMBeanTest.java for some unit test inspiration, or other tests in that package

Hauenstein:
@gtully - I added a test to the DuplexNetworkMbeanTest that provides coverage for this issue and a bit of MBeanBridgeDestination in general.  I updated my branch from Apache before pushing and now this pull request looks messier than I expected - I chalk it up to my own inexperience with the process.  Let me know if you see anything odd and I'll be happy to try to clean it up.

lucastetreault:

jbonofre:
Closing old PRs.

PR #119, opened 2015-06-23
AMQ-5821

cschneider:
Improved fix based on current master


jbonofre:
As discussed on IRC, this PR is not fully correct. We are waiting for a fix from Christian. Thanks !

PR #125, opened 2015-07-02
https://issues.apache.org/jira/browse/AMQ-5814

cshannon:
This fixes an issue where a producer's connection context was being
used to create a subscription instead of the consumer's context
which could lead to permission check failures.

Thanks to hongphu8790@gmail.com for providing the test case.


cshannon:
Closing PR as another patch is being applied with an OpenWire test instead.

PR #126, opened 2015-07-03
Fix class loading issues for SSL/nonSSL connectors

cjnygard:
- This resolves issues loading SSL and nonSSL RAR connectors in Glassfish 3.x
- proposed fix for AMQ-5873

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
+++ b/activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
@@ -69,18 +69,7 @@ public class FactoryFinder {
             if (className == null) {
                 throw new IOException("Expected property is missing: class");
             }
-            Class clazz = null;
-            ClassLoader loader = Thread.currentThread().getContextClassLoader();
-            if (loader != null) {
-                try {
-                    clazz = loader.loadClass(className);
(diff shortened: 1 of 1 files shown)


cjnygard:
Not sure if this is the correct way to solve the problem.  The patch incorporates what we've done internally to resolve the issue, which has existed since at least 5.8.x.  We mentioned this issue to Johan at one point. 

Let me know if the proper way to fix this is to derive our own StandaloneObjectFactory-variant and install via the FactoryFinder.setObjectFactory().  It seems as though the thread-local class loader is succeeding but causing cast exception in downstream code instead of failing and dropping through to the normal class loader.  

See Jira issue AMQ-5873 (https://issues.apache.org/jira/browse/AMQ-5873) for stack trace as was reported internally.

davsclaus:
Oh classloading is not fun, if this is merged then it may break something on other systems.

mattrpav:
Agree with Claus, this shouldn't be merged without a better understanding of the root cause. I'm not following how this would fix the problem. Perhaps in Glassfish an undesirable class is being loaded first via the TCCL? Or perhaps a service locator issue.

The current code falls back to the class's classloader if the TCCL does not find the class.

lucastetreault:

PR #134, opened 2015-07-13
https://issues.apache.org/jira/browse/AMQ-5857

cshannon:
Changing ActiveMQTextMessage to clear out the text field on marshal to
a ByteSequence to prevent the data from being stored in memory twice.


cshannon:
fix already applied

PR #135, opened 2015-07-14
https://issues.apache.org/jira/browse/AMQ-5879

cshannon:
Adding full exception to logs for tansport failures when the exception
is unexpected to help debugging.


cshannon:
Closing as turning on debug will already log the exception

PR #136, opened 2015-07-27
https://issues.apache.org/jira/browse/AMQ-5875

cshannon:
Reworking DurableConduitBridge to look up subscriptions from the
msesage store instead of the RegionBroker since inactive subscriptions
need to be looked at as well when dynamicOnly is false for a
network bridge.


cshannon:
fix has been merged

PR #142, opened 2015-09-14
Changed Readme from txt to rst, links added

vivekanand1101:


vivekanand1101:
I would like to know what the problem is. Presently, i am not able to  figure that out myself.

PR #150, opened 2015-10-06
AMQ-5931 Updated jodatime to 2.8.2

czobrisky:
https://issues.apache.org/jira/browse/AMQ-5931


czobrisky:
pulled in under 149 instead of 150 in comment.

PR #160, opened 2015-11-24
TransportConnection does not synchronize iteration on synchronized list

emopers:
In TransportConnection.java:861, the synchronized list returned by cs.getTempDestinations() is iterated
in an unsynchronized manner, but according to the [Oracle Java 7 API specification](http://docs.oracle.com/javase/7/docs/api/java/util/Collections.html#synchronizedList%28java.util.List%29),
this is not thread-safe and can lead to non-deterministic behavior. This pull request adds a fix by synchronizing the iteration on the list returned by cs.getTempDestinations().


lucastetreault:

jbonofre (approved):

PR #161, opened 2015-11-25
BDEV-10624

pbruski:


pbruski:
Sorry about this, it was meant to be opened only against our fork for now.

clebertsuconic:
You guys are doing openwire over HTTP? why not send the PR upsstream? :)

pbruski:
We will send it upstream, but first, we need to go through the legal process, fix backward compatibility with old XStream serialisation and prepare a PR against master (currently it's 5.9.0).

PR #162, opened 2015-12-04
https://issues.apache.org/jira/browse/AMQ-6071 - Log info about corru…

tmielke:
…pted journal records at WARN level


davsclaus:
@gtully @dejanb seems like a good idea to have WARN level for a corruption detected? Any thoughts?

hzbarcea:
Hi @tmielke , this looks good, a bit late, but I will merge it now. As a matter of style, I would prefer curlies and params to Logger methods as the full message string is not built unless if not at the appropriate log level. I will make that minor change myself. Thanks!

hzbarcea:
Looks like this was merged long time back, just the PR wasn't closed.

PR #165, opened 2016-01-22
AMQ-6135 ActiveMQ HTTP module should support Openwire serialisation

pbruski:
I am submitting this on behalf of Atlassian, the appropriate CCLAs are in place.

We have that code running in production and we didn't have any problems reports.
I assumed rather strict backward compatibility and interoperability requirements, not sure if that's needed.  Well, at least the interoperability should be there.
In any case, it's easier to have them removed than to implement them as an afterthought.

### DIFF
--- a/activemq-http/src/main/java/org/apache/activemq/transport/http/HttpClientTransport.java
+++ b/activemq-http/src/main/java/org/apache/activemq/transport/http/HttpClientTransport.java
@@ -16,8 +16,8 @@
  */
 package org.apache.activemq.transport.http;
 
-import java.io.DataInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InterruptedIOException;
 import java.net.URI;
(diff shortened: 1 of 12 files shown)


lucastetreault:

jbonofre:
IMHO, it should be better to have an http option to support openwire format. Something like:

```
  <transportConnector name="http" uri="http://0.0.0.0:1234?openwireSerializatrionEnabled=true" />
```

It would allow to support current marshalling and openwire marshalling.

NB: if you don't reply to this PR quickly, this one will be closed.

mattrpav:
> ` <transportConnector name="http" uri="http://0.0.0.0:1234?openwireSerializatrionEnabled=true" />`

or perhaps a wireformat.type=[xstream | openwire], etc.

I think it would be good to deprecate xstream and either update the openwire generator to include XML marshaling, or just favor openwire over http.

PR #167, opened 2016-02-05
[AMQ-6160] fix NPE; use messageId.getFutureOrSequenceLong() as sequen…

jenshadlich:
…ceId if entryLocator is NULL


davsclaus:
@gtully @tabish121 does this PR look okay to you guys? I am not familiar with that future thingy

tabish121:
There doesn't appear to be any tests here to validate that this is still an issue, many fixes have gone into the codebase since this was opened.

lucastetreault:

jbonofre:
I'm not able to reproduce it. I think it has been already resolved as `entryLocator` doesn't seem to be null.

PR #171, opened 2016-03-01
Delete redundant persistenceAdapter.getTransactionContext

alien11689:


davsclaus:
@gtully @dejanb must this code line be called? If so we should add a code comment about it must be called and that its returned value is not in use.

hzbarcea:
@alien11689 , good spot. A bit late, but merging. Thanks for the patch!

PR #172, opened 2016-03-03
Move directory variables up

TomyLobo:
Setting ACTIVEMQ_CONF to a different value and not moving it up will result in login.config being taken from the old location.
This commit fixes this by moving these variables above all places they are used in.

Fixes AMQ-6189
https://issues.apache.org/jira/browse/AMQ-6189


tabish121:
Generally a good practice to include the JIRA issue data in your PRs.  If you include a reference to the JIRA issue (AMQ-6189) the PR is related to the tooling will add a comment in the JIRA so that people can see where this comes from.

TomyLobo:
I didn't know there was a JIRA issue for this, I just thought I'd PR a fix :)

TomyLobo:
Thanks

PR #173, opened 2016-03-11
Update activemq.bat

samuelt1:
Added quotes so that when the path is expanded if there is a space in the path it will not see the path as two words, but one word


tabish121:
Please open a Jira Issue for this at https://issues.apache.org/jira/browse/AMQ

PR #195, opened 2016-08-04
Activemq 5.11.x

sundarmr:


hzbarcea:
This is not a change that could be made for a few reasons. One would be that the master deviated too much from the 5.11 branch, which is no longer actively maintained by the community. @sundarmr , could you please close this PR?

PR #197, opened 2016-08-10
https://issues.apache.org/jira/browse/AMQ-6391

pdudits:
Test case for multiple connections participating in one phase commit,
causing leak in ConnectionStateTracker.

Keep all the connections participating in an transaction in a static multimap
similar to the one in TransactionContext.

Even though Garry  says that static fields may cause issues in OSGi, I am offering this as a solution. This static map has same lifecycle, classloading and  source bundle as the one already present in TransactionContext, therefore I assume it cannot really make things worse.


lucastetreault:

jbonofre:
Disabling transaction does the trick. Closing old PR.

PR #199, opened 2016-09-23
Make start script work on busybox

fjollberg:
The current version of the activemq start script uses options to ps which are unavailable 
in busybox.

Suggested change refactors the start script to remove duplicated code to find a java process 
given a pidfile and makes it work on busybox.

I have limited possibilities of verifying the change on different platforms but it is confirmed to
work on at least Mac OS X in addition to busybox and should work on platforms where 
previous "ps -p" command worked.

### DIFF
--- a/assembly/src/release/bin/activemq
+++ b/assembly/src/release/bin/activemq
@@ -366,24 +366,27 @@ invokeJar(){
    return $RET
 }
 
-# Check if ActiveMQ is running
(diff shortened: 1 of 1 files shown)


lucastetreault:

jbonofre (approved):
Tested successfully on MacOS, Linux Debian (Docker and bare metal).

PR #200, opened 2016-09-26
https://issues.apache.org/jira/browse/AMQ-6442

sobkowiak:
Config file org.apache.activemq.server-default.cfg points config to ${karaf.base}/etc


sobkowiak:
Merged in 132840b09fbf0a071c51d357d12e13999925df25

PR #201, opened 2016-09-28
Activemq 5.6

wgqazll:


hzbarcea:
@wgqazll what is the purpose of this PR? It looks like it was submitted by accident and should not be merged.

PR #202, opened 2016-09-29
Fixes AMQ-6441 where a negative value can be returned with large AWS EFS files systems when calling java.io.File.getTotalSpace()

wcrowell:
Fixes AMQ-6441

A test, testLargeFileSystem(), was added to org.apache.activemq.broker.BrokerServiceTest for checking if the total space on a volume was too large to fit into a primitive long.  If this happens, then totalSpace is set to java.lang.Long.MAX_VALUE.


davsclaus:
According to the AMQ-6441 we need some improvement on this PR with a flag to turn this check on.

PR #203, opened 2016-10-03
fix a wrong word

yangwei71:
the wrong word in this sentence "with Hadoop based file systems to achive HA of your stored messages" .
“achive” change to "achieve".


hzbarcea:
This was a trivial one. Thank you for the correction.

PR #211, opened 2016-11-15
AMQ-6057: Add performance patch to JDBC statements definition

graben:


davsclaus:
@gtully @tabish121 this seems fine to me, but want to double check by some of the AMQ guys whom knows the JDBC datastore better than me.

graben:
The patch is almost one year old. Running fine for almost the same amount of time.

davsclaus:
Have you run with a patched version in production for a year? What JDBC database are you using?

And the SQL syntax changes are standard SQL which is supported on all major databases?

graben:
> Have you run with a patched version in production for a year?

Yes

> What JDBC database are you using?

Oracle

>And the SQL syntax changes are standard SQL which is supported on all major databases?

Yes

davsclaus:
Thanks for the PR it has been merged. Do you mind closing this?

PR #224, opened 2017-01-25
 AMQ-6577: honour usePrefetchExtension in TopicSubscription.

vveloso:
The current implementation of TopicSubscription always uses the prefetch extension feature, effectively ignoring any usePrefetchExtension option set in the configuration for the topic destination.

See: [AMQ-6577](https://issues.apache.org/jira/browse/AMQ-6577)


vveloso:
It appears that the unit test that failed is also failing on master.

PR #225, opened 2017-01-31
[AMQ-6586] Compilation of activemq-unit-test fails

sobkowiak:


sobkowiak:
Should be merged only in activemq-5.14.x branch

PR #226, opened 2017-02-16
Javadoc typo fix and cleanup

ecki:
This mostly fixes the url1/url2 copy and paste problem. It uses a bit more readable property syntax and introduces a single sentence (plain text) summary for class Javadoc and adds some links.


hzbarcea:
This should have been accepted a long time ago. Looks good, testing and merging shortly.

PR #227, opened 2017-02-25
SSL Connection leaks

thodimi1:
There are possibilities to create a denial of service attack just by opening TCP connection to a SSL port and keep it open.


hzbarcea:
Sounds like a valid patch, or at least a valid concern. I cannot find your actual patch in the haystack of other commits. Can you please help with a bit of cleanup or point me in the right direction?

hzbarcea:
Hi, can you please point us to the relevant commits? The list above is getting out of hand :). Many thanks for the contribution!

thodimi1:
qwe

PR #228, opened 2017-03-04
[AMQ-6601] AMQ with master and slave broker, shutting down slave causes a dead-lock

catull:
This PR addresses the issue https://issues.apache.org/jira/browse/AMQ-6601.

The proposed solution is to start the broker asynchronously, but only wait for its starting up if it is the master (i.e. non-slave).

Closes / fixes issue #6601.

Credit goes to @lichtin for coming up with the solution.


hzbarcea:
I think the contribution is good, but I am conflicted because I dread the if/then/else programming that crept up in parts of activemq. I am not referring to your if, but to conditional logic that's topology aware (i.e. master-slave). What if it's a full-mesh or star topology? On the other hand, your patch addresses an issue with code from the pre-nob era.

I think I'll give it a whirl tomorrow and apply the patch from 5.14.x up (reluctantly).

PR #232, opened 2017-04-25
set name for primary key on create table

sturman:


lucastetreault:

jbonofre (approved):

PR #238, opened 2017-06-01
add ClientIdFilterDispatchPolicy and testcase

kimmking:
ClientIdFilterDispatchPolicy dispatches messages in a topic to a given client. Then the message with a "PTP_CLIENTID" property, can be received by a mqtt client with the same clientId.

Many developer have been insterested to use this feature:
http://blog.csdn.net/kimmking/article/details/17449019


lucastetreault:

jbonofre (approved):
Tested successfully. I will just add documentation about that later (with refactoring of website).

PR #250, opened 2017-06-08
Fix STOMP header escaping when using WebSocket.

thezbyg:
Use wire format class to marshal frames, because method
StompFrame.format() does not escape disallowed characters in STOMP
headers and can cause invalid STOMP messages to be sent, as newline and
colon characters in headers are sent as is.

### DIFF
--- a/activemq-http/src/main/java/org/apache/activemq/transport/ws/jetty9/StompSocket.java
+++ b/activemq-http/src/main/java/org/apache/activemq/transport/ws/jetty9/StompSocket.java
@@ -16,12 +16,15 @@
  */
 package org.apache.activemq.transport.ws.jetty9;
 
+import java.io.DataOutputStream;
 import java.io.IOException;
+import java.nio.ByteBuffer;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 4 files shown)


tabish121:
I see no tests to validate, and you'd need to create a JIRA issue to cover this case.

activemq-http/src/main/java/org/apache/activemq/transport/ws/jetty9/StompSocket.java
```diff
@@ -46,8 +49,13 @@ public StompSocket(String remoteAddress) {
     @Override
     public void sendToStomp(StompFrame command) throws IOException {
         try {
+            ByteArrayOutputStream baos = new ByteArrayOutputStream();
```
clebertsuconic: if that is an issue.. this would be dead slow for the server...

if there's any issues around this.. this fix is not good anyways.

As Tim said.. you need a test.

thezbyg:
JIRA issue created: AMQ-6699
Test added.
ByteArrayOutputStream usage removed.

PR #254, opened 2017-06-16
AMQ-6706 Allow for hooks before broker stop occurs

zregvart:
When Camel is run embedded within ActiveMQ it's shutdown is called at the point the broker is already stopped, this can lead to data loss as there could be exchanges that have not ended.

This adds the ability to specify preStopHook(s) that can be invoked during ActiveMQ shutdown but before the broker is shutdown.

With this Camel can gracefully shutdown while the broker is still started.


tabish121:
Don't see any tests here that confirm this works or ensure it remains working into the future.

zregvart:
@tabish121 I've added tests

PR #256, opened 2017-07-20
AMQ-6775: Race condition makes messages queued in the session be delivered before messages queued in consumers.

fsgonz:


gtully:
would it be possible to add a unit test?

fsgonz:
I've added a test using Mockito to exemplify the conditions under which the out of order issue is verified.

PR #259, opened 2017-09-12
AMQ-6808: preserve destination for browsed scheduled messages

erik-wramner:
Save original destination as JMS property before overwriting it when browsing scheduled messages.


tabish121:
The changes seem fine, need to squash this commit into one commit to make it less confusing.  Also it might be good to remove the colon following the AMQ-6808 as that seems like it has broken the JIRA integration so the updates to the PR aren't showing up in ActiveMQ JIRA.

tabish121:
We're looking to release 5.15.1 relatively soon (next week or so) if you want this in the next release can you either A. Squash the commits, or B. Close and create a new clean version with the final changes?  Otherwise it won't see a release for awhile.

erik-wramner:
I've squashed the commits, removed the colon and pushed again, hope it works out!

PR #263, opened 2017-10-17
Race condition makes messages queued in the session be delivered before messages queued in consumers. #256

fsgonz:
Create new PR to trigger CI.


lucastetreault:

PR #264, opened 2017-11-13
AMQ-6861 Allow customisation of network bridge creation logic.

splatch:


hzbarcea:
I wouldn't hold my breath for Jenkins to validate this.

I like your patch. Chris gave another solution of achieving the same result. There is another one, less direct that what Chris pointed at, but your solution is cleaner. What I don't like about your patch is that it doesn't address the by far most common way of configuring a broker, which is xml. Without that this patch is not very useful.

Maybe you can add that too, or I will try to find some time for that. I do get the usefulness of your patch, if would allow for different forwarding strategies in larger nobs, such as different paths for small vs large messages, or priorities, or whatever.

splatch:
I've reorganised logic and now it is possible to set up bridge factory per network connector. It is injectable property so it can be created as regular bean within spring/blueprint xml and reused across multiple connectors giving opportunity to have more advanced configurations of bridge factory implementations.

hzbarcea:
I expected something easier to configure, with less xml elements using FactoryFinder, not injection, not a big IoC fan.

However it works, ran tests. Merging for now, we can improve later if necessary.

PR #265, opened 2017-11-30
AMQ-5669 Update Karaf feature files to add required namespace (minimal

amarkevich:
available version)


hzbarcea:
Good patch, except that the version is too old. Fixed the code using a more recent version. This PR is similar to #75, you get the credit though too. Many thanks for the contribution.

PR #267, opened 2017-12-20
AMQ-6327 - getNextScheduledTime() returns incorrect time when working…

snurmine:
… with day of month

PR #268, opened 2017-12-21
AMQ-6327 - Remove System.out.println

snurmine:
Accidently added System.out.println in https://github.com/apache/activemq/pull/267


clebertsuconic:
Wouldn’t be better to replace by a log.debig or Info instead ?

snurmine:
It left there accidently. If logging will be added, then it could be using proper logger

yinian:
I want ask a question? how to run the test case in activemq-leveldb-store.
Or can you tell me how to run it. Thank you

snurmine:
I am not familiar with this project that much, but tried with
mvn -Dtest=IndexRebuildTest test 
and it ran single test case. It seems to be maven project so instruction in http://maven.apache.org/surefire/maven-surefire-plugin/examples/single-test.html might work.

hzbarcea:
I have an idea who didn't do a proper review of #267. Many thanks for pointing it out! I'll take care of it tomorrow if somebody doesn't beat me to it.

PR #270, opened 2017-12-25
Update README.md

javahongxi:
only test


gaohoward:
What is it that you are actually testing?

PR #272, opened 2018-01-05
replicated LevelDB fix and debugging output

superawesome:
I know LevelDB is now deprecated, and this may not get merged because of that. I certainly don't want to become its maintainer. I had an unhealthy cluster, and did not want to try and migrate it while in that state. This is just a small change to get it healthy again after a slave encountered this error:

Unexpected session error: java.net.ProtocolException: Maximum protocol buffer length exeeded

In my case I believe this is due to having too many LevelDB log files to replicate (a separate LevelDB bug, which I intend to investigate now that this cluster is healthy again).

2 commits in this PR:

1) debugging output to track that down and make it more ... debuggable.

2) a 2-character change to increase a buffer size by 4x, to "fix" the problem.


lucastetreault:

jbonofre:
LevelDB will be removed from `master` soon.

PR #273, opened 2018-01-26
[AMQ-6892]: Test failing unexpectedly: CamelVMTransportRoutingTest.testSendReceiveWithCamelRouteIntercepting

ReedOei:
When I run the test CamelVMTransportRoutingTest.testSendReceiveWithCamelRouteIntercepting, it passes, but when I run all of the tests in activemq-camel, it fails. I am running this in Intellij.

It appears the problem is unclosed connections in some of the other tests.


lucastetreault:

jbonofre:
I tried several times but it works fine in maven. Maybe only IntelliJ is impacted.

PR #274, opened 2018-02-07
AMQ-6894: limit poison exception message to 1024

ashakirin:
AMQ-6894: Limit poison exception message to 1024 to avoid IOException be encoding


lucastetreault:

jbonofre (approved):
LGTM, it prevents to "saturate" the broker with bunch of reconnection due to large messages (including lot of exception).

PR #281, opened 2018-04-10
AMQ-6930 Expose an environment variable to allow redirect stdout/stderr to a file

alvinlin123:
This change does not modify the default behavior, the default behavior is still to pipe stdout/stderr to /dev/null. User can specify a file using an environment variable called `ACTIVEMQ_OUT` to  redirect stdout/stderr to the file. 

This change also added and fixed some tests in `assembly/src/test/scripts/init-script-testsuite`.


artnaseef:
The build check is faulty.  Local build passed.  Commit merged with thanks to Alvin Lin.

PR #284, opened 2018-05-23
AMQ-6974 Avoid overflow errors with timestamps

thisthat:
Some comparisons with timestamp values are not safe. This comparisons can trigger errors that were found in other Apache projects, e.g. KAFKA-4290. 
I changed those comparisons according to what the Java documentation recommends to help preventing such errors.


thisthat:
@franz1981  I add some tests that cover some of those classes ;)

PR #285, opened 2018-07-02
[AMQ-7005] Thin client OSGI bundle

vladimirfx:


dkulp:
Umm... I understand some of the reasons in the features.xml, but I don't understand why you are creating a new activemq-client-osgi jar.    The only difference between it and the current activemq-client jar seems to be the embedding of the hawtbuf classes.   All the other imports and everything seem to be the same (ish, some of the ranges are missing).      I don't see the point of that so would be -1 to creating another bundle for that.   HOWEVER, I would definitely be ok with updating the OSGi manifest in the current client jar to mark things as optional if they are optional and updating the features.xml to reflect that.

lucastetreault:

PR #287, opened 2018-07-15
[AMQ-7011] Override dispatch to always set dispatchAsync true

jgoodyear:
AMQ-7011 Override dispatch to always set dispatchAsync true.


jgenender:
@alvinlin123 - The doc will need to get updated.

jbonofre:
@jgoodyear according to the Jira, it seems to have been merged already. Can you double check if we can close this PR ? Thanks.

jgoodyear:
Yes, we can close it.

PR #288, opened 2018-07-18
Amq7013

jgoodyear:


activemq-client/src/main/java/org/apache/activemq/util/JenkinsHash.java
```diff
@@ -0,0 +1,258 @@
...
+ * limitations under the License.
+ */
+package org.apache.activemq.util;
+
+public class JenkinsHash {
+
+    private static long INT_MASK = 0x00000000ffffffffL;
+    private static long BYTE_MASK = 0x00000000000000ffL;
+
+    private static JenkinsHash _instance = new JenkinsHash();
```
martel: Would suggest adding a final there to make sure it initializes only once under multithreaded environment:

https://stackoverflow.com/questions/8865086/why-is-this-static-final-variable-in-a-singleton-thread-safe

jgoodyear: Thanks, we've updated the patch :)

PR #294, opened 2018-07-28
AMQ-7023

jgoodyear:
Add build profile to execute owasp dependency check plugin on ActiveMQ build.


jgoodyear:
The Jenkins PR job can not fetch the repo to perform a test. 
Whom has perms / access to fix the Jenkins job?

PR #296, opened 2018-08-05
[AMQ-7032] Update Shiro to 1.2.6

jgoodyear:


jgoodyear:
Jenkins error is the git reset hard issue, not the patch contents itself.

PR #297, opened 2018-09-09
AMQ-7050: Allow alternate persistence mechanism with SubQueueSelectorCacheBrokerPlugin

alechenninger:
More details in [JIRA](https://issues.apache.org/jira/browse/AMQ-7050). Had a go at this by pulling out an interface for cache interactions and including default, File-based implementation that behaves just as before.

This flexibility is designed to allow future, alternative implementations such as a JDBC-based cache (e.g. for folks using JDBC persistence).

Thanks for review!

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/plugin/PeriodicallyFlushedFileSubSelectorCache.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/plugin/PeriodicallyFlushedFileSubSelectorCache.java
@@ -126,14 +126,11 @@ public class PeriodicallyFlushedFileSubSelectorCache implements SubSelectorCache
     private void readCache() {
         if (persistFile != null && persistFile.exists()) {
             try {
-                try (FileInputStream fis = new FileInputStream(persistFile);) {
-                    ObjectInputStream in = new ObjectInputStream(fis);
-                    try {
+                try (FileInputStream fis = new FileInputStream(persistFile)) {
+                    try (ObjectInputStream in = new ObjectInputStream(fis)) {
                         subSelectorCache = (ConcurrentHashMap<String, Set<String>>) in.readObject();
                     } catch (ClassNotFoundException ex) {
                         LOG.error("Invalid selector cache data found. Please remove file.", ex);
-                    } finally {
-                        in.close();
                     }
                 }
             } catch (IOException ex) {
--- a/activemq-broker/src/main/java/org/apache/activemq/plugin/PeriodicallyFlushedFileSubSelectorCache.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/plugin/PeriodicallyFlushedFileSubSelectorCache.java
@@ -18,6 +18,12 @@ import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
 
+/**
+ * Simple {@link SubSelectorCache} which uses a {@link File} for persistence.
+ *
+ * <p>Cache is flushed to file periodically based on {@link #getPersistInterval()}.
+ * Failures to flush are only logged.
+ */
 public class PeriodicallyFlushedFileSubSelectorCache implements SubSelectorCache, Runnable {
     public static final long MAX_PERSIST_INTERVAL = 600000;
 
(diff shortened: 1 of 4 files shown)


sheri21:
https://github.com/apache/activemq/pull/297#issue-214146120

lucastetreault:

activemq-broker/src/main/java/org/apache/activemq/plugin/SubQueueSelectorCacheBroker.java
```diff
@@ -58,39 +49,27 @@
...
-    private final Thread persistThread;
-    private long persistInterval = MAX_PERSIST_INTERVAL;
-    public static final long MAX_PERSIST_INTERVAL = 600000;
-    private static final String SELECTOR_CACHE_PERSIST_THREAD_NAME = "SelectorCachePersistThread";
-
     /**
      * Constructor
      */
-    public SubQueueSelectorCacheBroker(Broker next, final File persistFile) {
+    public SubQueueSelectorCacheBroker(Broker next, final SubSelectorCache subSelectorCache) {
```
jbonofre, line 69: Instead of changing the default behavior, I would rather create a totally new plugin, extending `SubQueueSelectorCacheBroker`. I think the configuration would be easier and less confusing for the users.

jbonofre (changes requested):

jbonofre:
@alechenninger do you have time to refactore this PR as a complete new plugin or do you want me to do it ? Thanks !

activemq-broker/src/main/java/org/apache/activemq/plugin/SubQueueSelectorCacheBroker.java
mattrpav: +1 JB

alechenninger: I can do that if you'd like, but I lean towards this approach for the usual composition-over-inheritance reasons. 

I'm interested in better understanding how this breaks previous users or changes the default behavior? The behavior is the same I believe, just factored out to a separate class.

If you're concerned about changing the constructor signature, I could add the original signature back and keep the new one.

ehossack-aws:
> @alechenninger do you have time to refactore this PR as a complete new plugin or do you want me to do it ? Thanks !

@jbonofre @alechenninger Since this is an old PR and your availability might have changed, I can offer to also refactor this PR to help move it forward.

alechenninger:
@jbonofre @ehossack-aws I'm available, I just commented above. Trying to better understand what we're really after with the refactoring recommendation before moving forward.

activemq-broker/src/main/java/org/apache/activemq/plugin/SubQueueSelectorCacheBroker.java
alechenninger: @jbonofre @mattrpav Thoughts appreciated! Again, happy to move forward, just wanted to take the opportunity to understand a little bit more what you meant. Thanks!

activemq-broker/src/main/java/org/apache/activemq/plugin/PeriodicallyFlushedFileSubSelectorCache.java
```diff
@@ -0,0 +1,190 @@
+package org.apache.activemq.plugin;
```
jbonofre, line 1: Can you please add the ASF header here ? Thanks.

activemq-broker/src/main/java/org/apache/activemq/plugin/SubQueueSelectorCacheBroker.java
jbonofre: Yes, the default behavior is the same as you "provide" the plugin already (using file persist). My point is more in term of configuration for users. But I guess it's acceptable.

activemq-broker/src/main/java/org/apache/activemq/plugin/SubSelectorCache.java
```diff
@@ -0,0 +1,51 @@
+package org.apache.activemq.plugin;
```
jbonofre, line 1: Can you please add the ASF header here ?

jbonofre (changes requested):

jbonofre:
I just did a new pass. I agree with the change however, this PR needs:
1. fix the missing ASF headers in couple of files
2. rebase on `master`
I will do a new final pass and merge as soon as it's OK.

jbonofre:
Non active PR

PR #300, opened 2018-09-19
AMQ-7055 - Optimization on SequenceSet to prevent iterating through t…

alanprot:
When the index file has a huge number of free pages and the broker is starting up (loading the index) we end up in a O(n2) loop.

This was causing the broker to use 100% cpu and not being able to start up even after a long time (as i remember we had around 3 millions free page in this case)

https://github.com/apache/activemq/blob/master/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java#L428

https://github.com/apache/activemq/blob/master/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/util/SequenceSet.java#L118

I noticed that almost all the free pages was being added to the end of the sequenceSet... So for any free page the broker had to necessity iterate through  all the Set (and after doing that for nothing add . the value to the tail).

With this small change, the broker started up in less than 5 minutes.

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/util/Sequence.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/util/Sequence.java
@@ -38,6 +38,10 @@ public class Sequence extends LinkedNode<Sequence> {
         return last + 1 == value;
     }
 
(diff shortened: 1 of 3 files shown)


jbonofre:
This change has been already applied.

PR #302, opened 2018-10-05
[AMQ-7067] KahaDB Recovery can experience a dangling transaction when…

jgoodyear:
[AMQ-7067] KahaDB Recovery can experience a dangling transaction when prepare and commit occur on different pagefiles.

Test cases added for Commit and Rollback on XA and Non-XA transactions.

Executed kahadb-store tests, activemq-unit-tests, and ran full build.


jgoodyear:
I need to add a null guard on the rollback command.
Will send in new PR.

PR #303, opened 2018-10-06
Use varargs for log parameters instead of `new Object[]`

FSchumacher:
Some log statements where called with a combination of `new Object[] {...}` and
an exception. That lead to a misinterpretation of those parameters. The first
object array is considered to be the first argument and the exception the second
parameter of the format string.
It should have been interpreted as the members of the object array as the parameters
for the format string and the exception as the exception itself.

So for example:

```
  LOG.info("User {} has done {}", new Object[] {"A", "something"}, e);
```

would log something like:

```
  User [A, something] has done java.lang.RuntimeException: OOps
```


lucastetreault:

jbonofre:
Superseeded by PR #631

PR #304, opened 2018-10-06
Add missing log param

FSchumacher:
Found two places, where either a placeholder or a parameter was missing.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnection.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnection.java
@@ -581,7 +581,7 @@ public class TransportConnection implements Connection, Task, CommandVisitor {
         if (consumerExchange != null) {
             broker.acknowledge(consumerExchange, ack);
         } else if (ack.isInTransaction()) {
-            LOG.warn("no matching consumer, ignoring ack {}", consumerExchange, ack);
+            LOG.warn("no matching consumer for {}, gnoring ack {}", consumerExchange, ack);
         }
         return null;
     }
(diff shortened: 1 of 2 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -2100,7 +2100,7 @@ private PendingList doPageInForDispatch(boolean force, boolean processExpired, i
                                 ConnectionContext connectionContext = createConnectionContext();
                                 dropMessage(ref);
                                 if (gotToTheStore(ref.getMessage())) {
-                                    LOG.debug("Duplicate message {} from cursor, removing from store", this, ref.getMessage());
+                                    LOG.debug("{}, duplicate message {} from cursor, removing from store", this, ref.getMessage());
```
coheigea: I think here we could just remove logging this from the log statement

jbonofre: Can you please use:

```
LOG.debug("Duplicate message {] from cursor, removing from store", ref.getMessage());
```

`this` is useless here.

FSchumacher: Right, the problem was, that `this` was used as first parameter in the log format string that had only one parameter. So the second  parameter `ref.getMessage()` was silently dropped.
As I didn't know, if `this` was needed for your analysis of the log messages, I added a second parameter placeholder.
I will try to update this PR.

activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnection.java
```diff
@@ -581,7 +581,7 @@ public Response processMessageAck(MessageAck ack) throws Exception {
         if (consumerExchange != null) {
             broker.acknowledge(consumerExchange, ack);
         } else if (ack.isInTransaction()) {
-            LOG.warn("no matching consumer, ignoring ack {}", consumerExchange, ack);
+            LOG.warn("no matching consumer for {}, gnoring ack {}", consumerExchange, ack);
```
coheigea: gnoring -> ignoring

coheigea:
@jbonofre This PR can be closed, as it's superceded by https://github.com/apache/activemq/pull/304

PR #308, opened 2018-10-13
[AMQ-7072] moving to JSON-B API instead of jackson

rmannibucau:
Goal is to let users switch the impl on need + using johnzon as default
Issue was identified on TomEE where jackson is not desired in the distribution cause there is already johnzon and some


jbonofre:
R: @jbonofre

lucastetreault:

mattrpav:
Thoughts on moving the requirement to json-b spec api, but staying with Jackson as impl?

jbonofre:
Johnzon is fine using jsonb. I've started the review and rebase anyway.

rmannibucau:
Overall the goal is to enable to use json-b and use the json-b contract (ie neither johnzon nor jackson must appear in compile scope in theory - until it is to be transitive but I'm speaking on a pure build perspective). I assume it should be tested with both impl - and maybe yasson - to ensure some portability - at least manually for the first round.
After the question of the default in the distro can be discussed. The least changing is jackson but the most apache is johnzon and for AMQ it should be very close once fully migrated to json-b api so at the end I'd say 50-50 ;).

mattrpav:
I think we should do this in 5.19.x and go straight to Jakarta. 

I don’t think it makes sense changing out a bunch of dependencies right before a major uplift to Jakarta.

The Jakarta works is ~80% done or so, so I don’t think it will be too long of a delay.

rmannibucau:
+1 and maybe jsonp only to limit deps

jeanouii:
@mattrpav I'm fine with it. I'll wait till we have a jakarta ready version for 5.19.x and rebase so I can update the dependencies and the imports.

Or I can rebase on your branch https://github.com/apache/activemq/pull/996, but probably too early

jbonofre:
Closing as non active. Please reopen a new one if needed.

PR #310, opened 2018-10-15
[AMQ-7076] Fix and update features to fully support Karaf 4.2.x

jbonofre:


activemq-karaf/src/main/resources/features-core.xml
```diff
@@ -20,9 +20,12 @@
 
 <features name="activemq-core-${project.version}">
 
+    <repository>mvn:org.apache.karaf.features/spring-legacy/[4,5)/xml/features</repository>
```
graben: Hi @jbonofre,
this line causes my karaf always to resolve featurea from karaf 4.2.2 SNAPSHOT if using 5.15.7 ActiveMQ feature. Might be a bug in the karaf feature resolver?
Regards,
Benjamin

PR #314, opened 2018-11-01
Fix missing trace argument

ecki:
LGTM.com found this mismatch between format and arguments.

Lets also make sure the array is printable.
Source: https://lgtm.com/projects/g/apache/activemq/alerts/?mode=list


ecki:
Looks like the master does not have this problem anymore, 
Code moved to  org.apache.activemq.TransactionContext.commit(Xid, boolean) L564 and rollback have proper log arguments.

PR #320, opened 2018-11-22
[AMQ-7076] Does not define spring-legacy repository by default in Act…

jbonofre:
…iveMQ Karaf features repo, and let the user add it


jbonofre:
This is the fix for `master` branch.

PR #321, opened 2018-11-22
Enable secure processing on all DocumentBuilderFactory.newInstance()

coheigea:
Good security practices say that all DocumentBuilderFactory instances should enable the secure processing feature and disable doctypes to avoid XXE attacks.


jbonofre:
R: @jbonofre

jbonofre:
LGTM, I will merged when Jenkins is happy.

gtully:
it would be nice if there was a corresponding jira to track this improvement into a release.

coheigea:
Hi Gary! Here it is: https://issues.apache.org/jira/browse/AMQ-7110

gtully:
thanks Colm

PR #326, opened 2018-12-02
AMQ-7118: fix issue with the unchanged acknowledgement map.

heathkesler:
If the kahadb ack message file map has not changed, then there is no reason to write it again.  So all the kahadb store tests and activemq unit test suite pass.


jbonofre:
It has been already applied.

PR #327, opened 2018-12-02
AMQ-7118: fix issue where the ackMessageFileMapLocation gets written on every checkpoint

heathkesler:
fix issue where the ackMessageFileMapLocation gets written on every checkpoint, even if it hasnt changed and include the use of the Atomic boolean.


activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
```diff
@@ -1718,7 +1722,10 @@ private void checkpointUpdate(final boolean cleanup) throws IOException {
 
         metadata.state = OPEN_STATE;
         metadata.producerSequenceIdTrackerLocation = checkpointProducerAudit();
-        metadata.ackMessageFileMapLocation = checkpointAckMessageFileMap();
+        if (metadata.ackMessageFileMapDirtyFlag.get() || (metadata.ackMessageFileMapLocation == null)) {
```
alvinlin123: isn't the `ackMessageFileMapDirtyFlag` redundant to the `ackMessageFileMapMod` local variable down below?

PR #334, opened 2019-01-01
AMQ-7125 Stomp subscribe empty destination fix

avi5kdonrh:
Fix for: The Broker does not send error frame back to the client, when an empty destination is specified in a stomp subscribe frame


gtully:
pulled into: https://gitbox.apache.org/repos/asf?p=activemq.git;h=b10458e

PR #336, opened 2019-01-11
AMQ-7132 - ActiveMQ reads lots of index pages upon startup (after a g…

alanprot:
…raceful or ungraceful shutdown)


activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
```diff
@@ -2392,6 +2445,38 @@ public String toString() {
...
+                dataOut.writeLong(object.getMessageSize().getMaxSize());
+                dataOut.writeLong(object.getMessageSize().getMinSize());
+                dataOut.writeLong(object.getMessageSize().getCount());
+            }
+        }
+
+        @Override
+        public MessageStoreStatistics readPayload(final DataInput dataIn) throws IOException {
+
+            if (!dataIn.readBoolean()) {
```
cshannon: Shouldn't we be checking for the KahaDB version here to make sure it's version 7 or greater before trying to read the payload since older versions will never have it?  Any example of that is here:

https://github.com/apache/activemq/blob/activemq-5.15.8/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java#L2809

alanprot: This check ine being done in the StoredDestinationMarshaller, see:

https://github.com/apache/activemq/pull/336/commits/eebcdbfe7b1681e6093439ff5be281a585a307c4#diff-e3b8fff8c2133dfd70999705bbb558b3R2580

In this case, if the Kahadb < 7 I cannot call dataIn.readLong() in the StoredDestinationMarshaller.

cshannon: Ah ok thanks for pointing that out, looks good to me.

PR #339, opened 2019-01-17
AMQ-7138 Hardcoded paths to a karaf etc directory in a file features.xml

jherkel:


jherkel:
replace hardcoded path to a karaf etc directory with ${karaf.etc}

PR #343, opened 2019-01-30
AMQ-7143 - Temporary transaction file (PageFile) being opened and clo…

alanprot:
…sed many times, leading poor performance


activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/disk/page/TransactionTest.java
```diff
@@ -0,0 +1,76 @@
+package org.apache.activemq.store.kahadb.disk.page;
```
cshannon: You are missing the apache license header

alanprot: I just  updated the PR! :D

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/Transaction.java
```diff
@@ -682,8 +681,8 @@ public void commit() throws IOException {
     public void rollback() throws IOException {
         if( writeTransactionId!=-1 ) {
             if (tmpFile != null) {
+                getTempFile().delete();
```
cshannon: Shouldn't this temp file delete come after the close of the RandomAccessFile?

alanprot: Good catch @cshannon  :D Thanks!

I just updated the PR :)

PR #346, opened 2019-02-14
Modified code to not read from file every time but use last modified timestamp and cache filedata.

nomstradamus:
As per the comment given by the earlier developer the code could be optimized by checking only the modified date. Made the changes for the optimization.


lucastetreault:

jbonofre:
Your change basically breaks the client. For instance, `ActiveMQSslConnectionFactoryTest` is stuck, up to `OutOfMemory`.

I'm closing the PR as it's not correct. Feel free to reopen it with correct change: it should pass the Jenkins build (now enabled on PR).

PR #347, opened 2019-02-25
Use secure repository url

oflebbe:
Should be obvious


coheigea:
@jbonofre - This can be closed, already fixed in 'b65797245b'

PR #348, opened 2019-02-27
close DataByteArrayOutputStream buff of WriteBatch

nguyenhoan:


cshannon:
@nguyenhoan - What was the purpose of this change?  Was it to fix a bug you found?

@jbonofre - as I've mentioned before, please do not merge PRs if they don't have Jiras or tests or any validation or even any description associated with the PR.  I had no idea this change even existed because without a Jira it doesn't link to the release notes.

The big problem is now there is a possible regression (still needs to be verified) but no one has any idea now why this change was made or necessary.  If this was supposed to fix a bug we need to have a description of the bug and a test case to validate it.  At minimum all PRs should have a Jira describing the issue/feature, a good commit message describing the commit and unit tests to validate before merging.

There's now discussion that this has caused a regression on this thread: https://issues.apache.org/jira/browse/AMQ-6831 so we might need to rush out 5.15.12 and revert this change if it is indeed causing KahaDB corruption problems as that is a major issue.

nguyenhoan:
> @nguyenhoan - What was the purpose of this change? Was it to fix a bug you found?

@cshannon Yes, it was a fix to a bug I found.

PR #349, opened 2019-03-01
AMQ-7159 - Adding a new attribute on PersistenceAdapterViewMBean to s…

alanprot:
…how information about Storage write latency

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/PersistenceAdapterView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/PersistenceAdapterView.java
@@ -16,16 +16,25 @@
  */
 package org.apache.activemq.broker.jmx;
 
-import java.util.concurrent.Callable;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.apache.activemq.management.TimeStatisticImpl;
 import org.apache.activemq.store.PersistenceAdapter;
+import org.apache.activemq.store.PersistenceAdapterStatistics;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.concurrent.Callable;
 
 public class PersistenceAdapterView implements PersistenceAdapterViewMBean {
 
+    private final static ObjectMapper mapper = new ObjectMapper();
     private final String name;
     private final PersistenceAdapter persistenceAdapter;
 
     private Callable<String> inflightTransactionViewCallable;
     private Callable<String> dataViewCallable;
+    private PersistenceAdapterStatistics persistenceAdapterStatistics;
 
     public PersistenceAdapterView(PersistenceAdapter adapter) {
         this.name = adapter.toString();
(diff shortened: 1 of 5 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/PersistenceAdapterView.java
```diff
@@ -64,11 +88,46 @@ private String invoke(Callable<String> callable) {
         return result;
     }
 
+    private String serializePersistenceAdapterStatistics() {
+        if (persistenceAdapterStatistics != null) {
+            try {
+                final ObjectMapper mapper = new ObjectMapper();
```
cshannon: ObjectMapper is threadsafe so you should just create it one time for the class so we don't need to re-create it over and over.

activemq-broker/src/main/java/org/apache/activemq/store/PersistenceAdapterStatistics.java
```diff
@@ -0,0 +1,32 @@
+package org.apache.activemq.store;
+
+import org.apache.activemq.management.TimeStatisticImpl;
+
+public class PersistenceAdapterStatistics {
```
cshannon: This looks fine except I think the rest of the statistics classes extend StatsImpl so you may want to follow that pattern to be consistent.  (For example look at DestinationStatistics)

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
```diff
@@ -249,6 +250,7 @@ public void writePayload(Metadata object, DataOutput dataOut) throws IOException
     protected PageFile pageFile;
     protected Journal journal;
     protected Metadata metadata = new Metadata();
+    protected PersistenceAdapterStatistics persistenceAdapterStatistics = new PersistenceAdapterStatistics();
```
cshannon: I would probably make this final and then I would also add a getter for this object and also expose it under KahaDBPersistenceAdapter. I routinely use embedded brokers for example and being able to access this through KahaDBPersistenceAdapter without JMX would be nice.  Inside KahaDBPersistenceAdapter I would just return the actual PersistenceAdapterStatistics object so someone could access it.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/PersistenceAdapterView.java
alanprot: Good catch!

activemq-broker/src/main/java/org/apache/activemq/store/PersistenceAdapterStatistics.java
alanprot: Done! :)

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
alanprot: Done.

PR #351, opened 2019-03-27
ActiveMQMessageConsumer - do not log whole message including headers in WARN

maciejmiklas:
Message can contain large number of headers, logging them in warning drastically increases log size.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
@@ -1457,7 +1457,7 @@ public class ActiveMQMessageConsumer implements MessageAvailableConsumer, StatsC
                             session.getConnection().rollbackDuplicate(this, md.getMessage());
                             dispatch(md);
                         } else {
-                            LOG.warn("{} suppressing duplicate delivery on connection, poison acking: {}", getConsumerId(), md);
+                            LOG.warn("{} suppressing duplicate delivery on connection, poison acking: {}", getConsumerId(), md.getMessage());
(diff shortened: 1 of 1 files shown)


lucastetreault:

jbonofre (approved):
It makes sense as it can be very verbose. Thanks !

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
```diff
@@ -1457,7 +1457,7 @@ public void dispatch(MessageDispatch md) {
                             session.getConnection().rollbackDuplicate(this, md.getMessage());
                             dispatch(md);
                         } else {
-                            LOG.warn("{} suppressing duplicate delivery on connection, poison acking: {}", getConsumerId(), md);
+                            LOG.warn("{} suppressing duplicate delivery on connection, poison acking: {}", getConsumerId(), md.getMessage());
```
jbonofre, line 1460: Rethinking about that, I think the whole message can be helpful.
A poison ack happens when the message redelivery exhausted (for instance, the redelivery count is greater than the limit). Having the whole message, including the headers is helpful to have some clue why the message has been redelivered (it could be related to a bad header for a consumer).

maciejmiklas: You can add another line, that logs content of the message in DEBUG, but logging whole message in WARN is a logging-anti-pattern. We have Productive system that is processing millions of messages every day, it runs in INFO and I really like to see in log only important stuff and now it drops messages into log, where each WARN has a size of A4-page.

PR #352, opened 2019-04-03
Applied fix for data writer thread dying an early death

alexndr79:


lucastetreault:

activemq-core/src/main/java/org/apache/activemq/kaha/impl/async/DataFileAppender.java
```diff
@@ -5,9 +5,9 @@
  * The ASF licenses this file to You under the Apache License, Version 2.0
  * (the "License"); you may not use this file except in compliance with
  * the License.  You may obtain a copy of the License at
- *
- *      http://www.apache.org/licenses/LICENSE-2.0
- *
+ * <p>
```
jbonofre, line 8: This kind of change is useless in the PR: it actually creates confusion, as the change part is actually small (exception handling), whereas the formatting changes creates noise.
Can you please rework the PR ? Else, I will extract the change part.

jbonofre (changes requested):

jbonofre:
ActiveMQ 5.6.x is not on the target anymore. Auto closed.

PR #363, opened 2019-05-30
AMQ-7219 - ActiveMQ replays journal file on a clean/unclean shutdown …

alanprot:
…with transacted session + Non persistent Messages


alanprot:
Yeah... @gtully thanks for the heads up! I missed that!

I changed the PR to use the 'processLocation' method -> it locks the index.

PR #365, opened 2019-06-06
AMQ-7149 Remove dependency activemq-http and activemq-stomp

jgallimore:
This PR moves `XStreamSupport` to `activemq-client`. 

Currently this sits in `activemq-stomp`, thus requiring the ActiveMQ HTTP client to require `activemq-stomp` module on its class path, even if the client isn't using stomp.

### DIFF
--- a/activemq-client/pom.xml
+++ b/activemq-client/pom.xml
@@ -71,6 +71,12 @@
       <optional>true</optional>
     </dependency>
 
+    <dependency>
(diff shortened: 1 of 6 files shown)


lucastetreault:

jbonofre (approved):
I think it makes sense. I'm checking the impact in Karaf/OSGi (it should not impact, but I would like to double check). Thanks !

PR #366, opened 2019-06-06
AMQ-7419 - Remove dependency activemq-http and activemq-stomp

jgallimore:
This PR moves `XStreamSupport` to `activemq-client`. 

Currently this sits in `activemq-stomp`, thus requiring the ActiveMQ HTTP client to require `activemq-stomp` module on its class path, even if the client isn't using stomp.

### DIFF
--- a/activemq-client/pom.xml
+++ b/activemq-client/pom.xml
@@ -71,6 +71,12 @@
       <optional>true</optional>
     </dependency>
 
+    <dependency>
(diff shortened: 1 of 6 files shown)


jbonofre:
This is a breaking change in 5.15.x, it should not happen in minor release. I prefer to keep this change only for next 5.16.x.

PR #374, opened 2019-07-15
Remove default "secret" password from the LDAPAuthorizationMap

coheigea:
It's not good security practice to use default passwords.


cshannon:
@jbonofre - As I've said before you need to make sure that there are matching JIRA issues for commits. If we don't have JIRAs then changes are not part of release notes and it is much less visibility. So please stop merging commits without JIRAs.  Can you go back and create issues for everything you have been merging? The release notes need to have all the changes before being voted on.

cshannon:
@coheigea  - And Colm just make sure you create JIRAs for any PRs even if they are small for the same reasons.

PR #376, opened 2019-07-17
Update HTTPClient and replace deprecated calls

coheigea:


cshannon:
@jbonofre and @coheigea  - as I have now asked a few times... these need Jiras so it can be tracked

I'm going to vote -1 on any release that doesn't include Jiras for all of these commits because it's important for visibility...please stop merging random fixes in without issues being created.

coheigea:
@cshannon Understood - I'll create JIRAs going forwards for any new issues. I'm on vacation ATM - will check all issues merged without JIRAs when I get back, if someone doesn't do it in the meantime.

cshannon:
@coheigea - Awesome thanks. You have a lot of fixed merged in so will be good to have visibility when we do the release for the users.

coheigea:
@cshannon I think I've created all relevant JIRAs for 5.15.10 - let me know if there are any problems with it.

PR #388, opened 2019-08-11
[AMQ-7249] Upgrade to Scala 2.13.0

jbonofre:


jbonofre:
Note: scala is only used in leveldb which is now deprecated. I don't consider this one as a blocker as we should remove leveldb from the codebase.

PR #391, opened 2019-08-27
AMQ-7292 - WebConsole problem due to missing ecj jar

coheigea:


jbonofre:
@cshannon hey Chris, the purpose of the Jetty upgrade (and Camel) is to address CVE issues. I think it's important to address those in "minor" releases when possible, That's why I upgraded Jetty there. I think it's worth to keep for 5.15.10 as it's expected by some users.

jbonofre:
@cshannon good idea, I will prepare a release notes "highlights" for the website.

PR #402, opened 2019-10-31
[AMQ-7325] Upgrade to Derby 10.15.2.0

jbonofre:


jbonofre:
`EmbeddedDataSource` needs an update. I'm fixing.

gemmellr:
Derby 10.15.1.3 requires Java 9 to run. Changing the minimum JDK requirements to 9+ seems like it would be a discussion topic.

jbonofre:
@gemmellr that's for master

jbonofre:
But agree to be discussed

gemmellr:
I know its on master but to me the branch isnt that important, still seems like a discuss-before-commit type of change, especially if a release from the branch was already proposed to happen.

jbonofre:
@gemmellr agree, I will start a thread on the mailing list about 5.16.0 minimum Java version. My understanding is to fully support Java 9+, but let's double check.

jbonofre:
@gemmellr just started the discussion on the mailing list.

gemmellr:
I'm not arguing either way, just saying it needs discussed first, given its happening well after a release was proposed. I would say supporting 9+ doesn't necessarily need to mean dropping 8, unless folks want to. Obviously Derby have done that though, so clearly folks are starting to do it.

gemmellr:
Great!

lucastetreault:

PR #409, opened 2019-11-04
Update owasp plugin (and enable it to run correctly)

coheigea:
Currently, the owasp plugin doesn't run correctly due to some system scope failures.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -95,7 +95,7 @@
     <leveldbjni-version>1.8</leveldbjni-version>
     <log4j-version>1.2.17</log4j-version>
     <mockito-version>1.10.19</mockito-version>
-    <owasp-dependency-check-version>4.0.2</owasp-dependency-check-version>
+    <owasp-dependency-check-version>5.2.2</owasp-dependency-check-version>
     <powermock-version>1.6.5</powermock-version>
     <mqtt-client-version>1.15</mqtt-client-version>
     <openjpa-version>1.2.0</openjpa-version>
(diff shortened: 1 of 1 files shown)


chughpiyush:
@coheigea Hi, I am trying to run owasp analysis but it is failing in `leveldb-store` for me, reason being -> `Unable to resolve system scoped dependency: com.sun:tools:jar:1.8.0:system`

- This is happening despite `skipSystemScope` being set as true.
- I am running `aggregate` goal.
- Did you happen to face this issue? or can you please try to reproduce this issue on the latest master branch?

coheigea:
Yes, the owasp plugin configuration (which skips the system scope) is only defined in the "owasp" profile. So to get owasp to work on master, you can do something like: mvn clean install -Powasp -DskipTests

chughpiyush:
> Yes, the owasp plugin configuration (which skips the system scope) is only defined in the "owasp" profile. So to get owasp to work on master, you can do something like: mvn clean install -Powasp -DskipTests

Oh great, That worked! One more question since you're here :)
I am new to this repository and want to contribute some fixes, what will be the best way for me to test locally before creating a PR? (Assuming I want to start with versioning fixes in pom)

Thank you so much!

coheigea:
Generally I'd say do a "mvn clean install" and see if the tests pass (this takes a long time and some of the tests are a bit flaky). If it touches on the distribution, then build a new one, and start it up and run some basic tests with it.

PR #410, opened 2019-11-06
[AMQ-7327] Use maxFrameSize to limit HTTP content length

jbonofre:


activemq-http/src/test/java/org/apache/activemq/transport/http/HttpMaxFrameSizeTest.java
```diff
@@ -0,0 +1,64 @@
...
+        brokerService.start();
+        brokerService.waitUntilStarted();
+    }
+
+    @After
+    public void teardown() throws Exception {
+        brokerService.stop();
+    }
+
+    @Test(expected = JMSException.class)
```
tabish121: Expecting an JMSException on a test that calls many methods that can throw a JMSException doesn't actually test that the change does what you are expecting it to.  The producer send should be the point of test here so that you confirm that it does indeed fail when to large, but also you need some negative validation (send a smaller message) to test that your configuration doesn't get mistakenly applied to smaller messages.

jbonofre: @tabish121 thanks Tim, you are right, let me improve a bit the test coverage ;)

jbonofre: @tabish121 updated.

PR #420, opened 2019-11-20
AMQ-7291 initialize writing to body not needed to set property

me3x:
Unable to write property when readOnlyProperties=false, but readOnlyBody=true


gtully:
I think the change is good, but the pr would  benefit from a little test to show the  broken existing behaviour

PR #421, opened 2019-11-20
AMQ-7349 - Update some javascript libraries

coheigea:
I tested the webconsole and the examples that were changed.


jbonofre:
I tested locally the webconsole without problem.

PR #423, opened 2019-11-23
Stop failed timer task from breaking timers.

mdaley:
Very interested to see whether this PR is considered to be useful...

At my current client, we came across the problem that for some reason (that we don't understand yet) a micro-service with a connection to the broker via the activemq-client would get in a state where every 5 seconds it would report a failure to connect to the broker with the reason being: _IllegalStateException: Timer already cancelled_.  

There are several examples on the internet of this problem occurring but I've not found an actual solution to the problem, or something that explains what the underlying cause is.

After much digging, I've come to think that the basic cause of the problem is in `java.util.Timer`.  Tasks controlled by a timer should never throw exceptions because, if they do, the timer then becomes unusable.  What I think is happening is that for some reason we have a task that is throwing an exception.  However, this exception gets masked by the behaviour of the Timer; and, then, the Timer just continues ad-infinitum with IllegalStateExceptions every five seconds.

The change I've made here is simply to put try/catch into the SchedulerTimerTask so the exception from the failed underlying task can be reported; and so that the timer does not actually break (maybe the next invocation of the underlying task will work fine if it were, for example, to be due to a connection failure or something else intermittent).

In java.util.Timer there is this code:
```
    public void run() {
        try {
            mainLoop();
        } finally {
            // Someone killed this Thread, behave as if Timer cancelled
            synchronized(queue) {
                newTasksMayBeScheduled = false;
                queue.clear();  // Eliminate obsolete references
            }
        }
    }
```

Interesting comment!  Any exception in the mainLoop() kills the Timer; note that the mainLoop() does nothing itself to catch exceptions.

This problem has been noted elsewhere, for example: https://bugs.eclipse.org/bugs/show_bug.cgi?id=394215

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/thread/SchedulerTimerTask.java
+++ b/activemq-client/src/main/java/org/apache/activemq/thread/SchedulerTimerTask.java
@@ -16,6 +16,9 @@
  */
 package org.apache.activemq.thread;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import java.util.TimerTask;
 
 /**
@@ -23,6 +26,8 @@ import java.util.TimerTask;
  *
  */
 public class SchedulerTimerTask extends TimerTask {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SchedulerTimerTask.class);
+
     private final Runnable task;
 
     public SchedulerTimerTask(Runnable task) {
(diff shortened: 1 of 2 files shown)


lucastetreault:

jbonofre (approved):
I tested your change, and it makes sense to me. Thanks !

PR #425, opened 2019-11-28
AMQ-7357 - Update Pax URL

coheigea:


jbonofre:
For the record (and next time ;) ), it would be great to set the target version in the commit comment.

PR #428, opened 2020-01-20
AMQ-7376 - Use correct type for collections retrieval

coheigea:


jbonofre:
It seems that this change is causing issue at least in Spring support. I'm investigating.

PR #432, opened 2020-01-27
AMQ-7390 - Update Jetty to 9.4.26.v20200117

coheigea:


jbonofre:
AMQ-7388 already exists about that.

PR #447, opened 2020-02-09
Make double-checked locking thread-safe by declaring the field volatile.

PascalSchumacher:


jbonofre:
I'm dealing with Jira about that. Thanks.

PascalSchumacher:
@jbonofre Thank you very much!

PR #448, opened 2020-02-10
Fix unused format arguments in log messages.

PascalSchumacher:
I am sure the wording can be improved.

I tried to copy the wording from similar log messages if there were any.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
@@ -2092,7 +2092,7 @@ public class Queue extends BaseDestination implements Task, UsageListener, Index
                                 ConnectionContext connectionContext = createConnectionContext();
                                 dropMessage(ref);
                                 if (gotToTheStore(ref.getMessage())) {
-                                    LOG.debug("Duplicate message {} from cursor, removing from store", this, ref.getMessage());
+                                    LOG.debug("Duplicate message {} from cursor, removing from store", ref.getMessage());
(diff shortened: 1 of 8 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -2092,7 +2092,7 @@ private PendingList doPageInForDispatch(boolean force, boolean processExpired, i
                                 ConnectionContext connectionContext = createConnectionContext();
                                 dropMessage(ref);
                                 if (gotToTheStore(ref.getMessage())) {
-                                    LOG.debug("Duplicate message {} from cursor, removing from store", this, ref.getMessage());
+                                    LOG.debug("{}, duplicate message {} from cursor, removing from store", this, ref.getMessage());
```
coheigea: I think here we should just remove this

activemq-client/src/main/java/org/apache/activemq/util/ClassLoadingAwareObjectInputStream.java
```diff
@@ -129,7 +129,7 @@ private void checkSecurity(Class clazz) throws ClassNotFoundException {
             try {
                 Class<?> answer = Class.forName(className, false, loader);
                 if (LOG.isTraceEnabled()) {
-                    LOG.trace("Loaded class: {} using classloader: {} -> ", new Object[]{className, cl, answer});
+                    LOG.trace("Loaded class: {} using classloader: {} -> {}", new Object[]{className, cl, answer});
```
coheigea: I think here we should remove new Object[] and just use varargs

PR #449, opened 2020-02-10
AMQ-7402: Inconsistent synchronization of getters/setters of some cla…

PascalSchumacher:
…sses

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/management/SizeStatisticImpl.java
+++ b/activemq-client/src/main/java/org/apache/activemq/management/SizeStatisticImpl.java
@@ -19,8 +19,8 @@ package org.apache.activemq.management;
 public class SizeStatisticImpl extends StatisticImpl {
 
     private long count;
-    private long maxSize;
-    private long minSize;
+    private volatile long maxSize;
+    private volatile long minSize;
     private long totalSize;
     private SizeStatisticImpl parent;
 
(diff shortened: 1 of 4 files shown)


chughpiyush:
@jbonofre Is anyone working on failing test cases on Jenkins?

jbonofre:
@chughpiyush yeah, I fixed some of then but not all, working on it (at least disabling some for now).

activemq-client/src/main/java/org/apache/activemq/management/SizeStatisticImpl.java
```diff
@@ -87,7 +87,7 @@ public synchronized void setTotalSize(long size) {
     /**
      * @return the maximum size of any step
      */
-    public long getMaxSize() {
+    public synchronized long getMaxSize() {
```
coheigea: For Max + Min Size, I think synchronized should be removed from the setter methods and just make the variables volatile instead.

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/index/HashIndex.java
```diff
@@ -383,7 +383,7 @@ int indexFor(Key x, int length) {
     // Property Accessors
     // /////////////////////////////////////////////////////////////////
 
-    public Marshaller<Key> getKeyMarshaller() {
+    public synchronized Marshaller<Key> getKeyMarshaller() {
```
coheigea: Instead remove synchronized from the setter, and make the variable volatile

PR #452, opened 2020-02-12
AMQ-7404 add debug logging for selectors loaded from the cache file

jgallimore:
Backport from master for AMQ-7404.


jbonofre:
Please, for the record, open PR only based on `master`. I cherry picked on 5.15.x in second step.

PR #465, opened 2020-02-19
Fix issue where the registry lookup was a hardcoded name and didn't a…

jgallimore:
…ccount for the connector path

The previous commit to ManagementContext to prevent rebinds on the RMI registry used a hardcoded lookup path, as opposed to considering the connectorPath attribute. This commit should fix that.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagementContext.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagementContext.java
@@ -70,6 +70,11 @@ public class ManagementContext implements Service {
      */
     public static final String DEFAULT_DOMAIN = "org.apache.activemq";
 
+    /**
+     * Default registry lookup name
+     */
+    public static final String DEFAULT_LOOKUP_NAME = "jmxrmi";
+
     static {
         String option = Boolean.FALSE.toString();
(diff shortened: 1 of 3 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagementContext.java
```diff
@@ -690,10 +690,17 @@ public JmxRegistry(int port) throws RemoteException {
             super(port);
         }
 
-        @Override
+        private String getLookupName() {
+            if (getConnectorPath() == null || getConnectorPath().length() == 0) {
+                return LOOKUP_NAME;
+            }
 
+            return getConnectorPath().replaceAll("^/+", "").replaceAll("/+$", "");
```
coheigea: As getConnectorPath doesn't change, I guess it would make sense to only perform these regex's once (on setConnectorPath?)

jgallimore: Done! Thanks for the review.

jbonofre:
@jgallimore do you mind to squash the two commits in a single one ? Thanks !

jgallimore:
Sure, see #471 .

PR #477, opened 2020-02-26
[AMQ-7131] Add connectionTimeout to avoid starvation

jbonofre:


michaelandrepearce:
@jbonofre is it worth pulling this library into its own repo so it can have its own lifecycle also will avoid having to dupe it all in artemis

jbonofre:
@michaelandrepearce I'm not sure I follow you. Here, we fixed in `activemq-jms-pool`, you mean to "reuse" artemis connection pool ?

jbonofre:
Ah, I guess you mean that `jms-pool` should be in a shared repository to be used by activemq and artemis, right ?

michaelandrepearce:
Yup exactly

jbonofre:
@michaelandrepearce you are right: if it's used in both ActiveMQ and Artemis, it makes sense to share it. The question is where ? Do we have to create a new `activemq-share` repo ?

PR #505, opened 2020-03-08
Avoid potential overflow in int multiplication before it is converted…

PascalSchumacher:
… to long.

Found by static code analysis: https://lgtm.com/projects/g/apache/activemq?mode=tree&ruleFocus=7900075

At least in some cases (did not check every case) this can not occur because of the range of the variables in use.

Not sure if it is worth merging this non the less. What do you think?


jbonofre:
Let me take a look if it's worth to merge.

jbonofre:
I think the overflow issue is really rare. However, worth to merge it, just to prevent.

PascalSchumacher:
Thanks!

PR #511, opened 2020-03-16
AMQ-7442 - Adding build support for JDK 11

cshannon:
This is the initial work to allow the broker to be compiled and build
with JDK 11 while keeping the source/target Java version at 8 so it
will also run on JDK 8+


cshannon:
closing for now as we need to drop Scala and LevelDB

jbonofre:
I also have a branch with full build on JDK 11+, I will open and will let you know.

lucastetreault:

jbonofre:
Superseded by #647

PR #515, opened 2020-03-23
Removing a deprecated dependency ant:ant:1.6.2

sourabhsparkala:
- Removing `ant:ant:1.6.2` from `org.apache.activemq:activemq-openwire-generator` `pom.xml`

This fixes #AMQ-7448

### DIFF
--- a/activemq-openwire-generator/pom.xml
+++ b/activemq-openwire-generator/pom.xml
@@ -43,11 +43,6 @@
       <groupId>annogen</groupId>
       <artifactId>annogen</artifactId>
     </dependency>
-    <dependency>
-      <groupId>ant</groupId>
-      <artifactId>ant</artifactId>
-      <version>1.6.2</version>
-    </dependency>
   </dependencies>
 
   <profiles>


sourabhsparkala:
> `<ant-version/>` should be removed from the main parent pom.
> 
> And I think this dependency is required as the openwire-generator is used in activemq-client for the `openwire-generate` profile. I think it's require for the provided task. Let me check.

Hello @jbonofre, could you please confirm again if `<ant-version>` should be removed? I just removed an unwanted deprecated dependency. This seems to be required. 

Please let me know.

Thanks
Sourabh

sourabhsparkala:
@jbonofre requesting for an update.

sourabhsparkala:
Hello @jbonofre. I confirmed the profile build of `openwire-generate` as instructed here in https://github.com/apache/activemq/blob/master/activemq-client/pom.xml#L301.

The old Ant dependency does not seem to be associated to any underlying modules as far as I could see.

Hence, it can be safely removed from `activemq-openwire-generator`

sourabhsparkala:
> <ant-version/>

`<ant-version/>` holds the version for the latest Ant groupid. Could you please calrify again?

lucastetreault:

jbonofre:
Superseeded by https://github.com/apache/activemq/pull/630

PR #519, opened 2020-03-26
AMQ-7455 Mitigate high connection bootstrap/teardown frequency...

chammp:
...when running in a EAP6 XA transactional environment.
More information can be found in https://issues.apache.org/jira/browse/AMQ-7455


jbonofre:
I guess it's fixed with AMQ-5790, right ?

jbonofre:
Any update about my comment related to AMQ-5790 ?

chammp:
> Any update about my comment related to AMQ-5790 ?

Ah, sorry for not answering at all.
As a matter of fact we did not yet check the fix because we were successfully using a patched version built in-house and we did not migrate to a newer version of the RA. But we did migrate to EAP7 in the meantime, so the whole setup changed. As far as I know with the current setup and the current EAP7 resource adapter the behavior we witnessed in March don't occur anymore.

lucastetreault:

jbonofre:
@chammp thanks for the update. I think that AMQ-5790 already fix the issue, but I have to double check.

PR #522, opened 2020-04-10
AMQ-7463 Getting concurrent modification exception

anushaCoder:
Raised issue AMQ-7463 Getting concurrent modification exception in FailTransport class due to hashmap. Hence created concurrentHashMap

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/state/ConnectionState.java
+++ b/activemq-client/src/main/java/org/apache/activemq/state/ConnectionState.java
@@ -46,7 +46,7 @@ public class ConnectionState {
     private final List<DestinationInfo> tempDestinations = Collections.synchronizedList(new ArrayList<DestinationInfo>());
     private final AtomicBoolean shutdown = new AtomicBoolean(false);
     private boolean connectionInterruptProcessingComplete = true;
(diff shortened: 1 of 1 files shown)


activemq-client/src/main/java/org/apache/activemq/state/ConnectionState.java
```diff
@@ -46,7 +46,7 @@
     private final List<DestinationInfo> tempDestinations = Collections.synchronizedList(new ArrayList<DestinationInfo>());
     private final AtomicBoolean shutdown = new AtomicBoolean(false);
     private boolean connectionInterruptProcessingComplete = true;
-    private HashMap<ConsumerId, ConsumerInfo> recoveringPullConsumers;
+    private ConcurrentMap<ConsumerId, ConsumerInfo> recoveringPullConsumers;
```
coheigea, line 49: Better just to define this as Map

jbonofre, line 49: As @coheigea just a `Map` here is better IMHO.

jbonofre:
I'm testing the branch locally.

PR #523, opened 2020-04-24
AMQ-7467 - A new XStream serializer is created on every message that flows through the broker causing high CPU utilization under load

thodimi1:


activemq-stomp/src/main/java/org/apache/activemq/transport/stomp/ProtocolConverter.java
```diff
@@ -280,11 +294,7 @@ protected void handleException(Throwable exception, StompFrame command) throws I
         // Let the stomp client know about any protocol errors.
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         PrintWriter stream = new PrintWriter(new OutputStreamWriter(baos, "UTF-8"));
-        if (exception instanceof SecurityException || exception.getCause() instanceof SecurityException) {
-            stream.write(exception.getLocalizedMessage());
-        } else {
-            exception.printStackTrace(stream);
-        }
+        exception.printStackTrace(stream);
```
jbonofre, line 297: Why not keeping the original code here writing SecurityException to the stream ?

activemq-stomp/src/main/java/org/apache/activemq/transport/stomp/ProtocolConverter.java
```diff
@@ -558,10 +568,6 @@ protected void onStompSubscribe(StompFrame command) throws ProtocolException {
             throw new ProtocolException("SUBSCRIBE received without a subscription id!");
         }
 
-        if (destination == null || "".equals(destination)) {
-            throw new ProtocolException("Invalid empty or 'null' Destination header");
-        }
-
         final ActiveMQDestination actualDest = translator.convertDestination(this, destination, true);
```
jbonofre, line 571: Why don't we have the ProtocolException if destination is not defined ?

activemq-stomp/src/main/java/org/apache/activemq/transport/stomp/ProtocolConverter.java
thodimi1: Corrected

activemq-stomp/src/main/java/org/apache/activemq/transport/stomp/ProtocolConverter.java
thodimi1: Corrected

PR #525, opened 2020-05-04
[AMQ-7473] Add stopOnError configuration to stop the broker when locker has an exception

jbonofre:


jbonofre:
I have to improve the stop around HTTP transport connector: the jetty blocking queue is hanging.

jbonofre:
I need to interrupt the threads in the Jetty QueueThreadPool to shutdown the http transport connector as fast as possible.

PR #543, opened 2020-06-22
[AMQ-7500] Update Java examples to use 1.8 source/target

jbonofre:


cshannon:
This was pushed in another commit

PR #552, opened 2020-07-23
AMQ-8009 - Spring upgrade required for Karaf feature

coheigea:


activemq-karaf/src/main/resources/features-core.xml
```diff
@@ -25,7 +25,7 @@
 
     <!-- Bundles needed if only client will be deployed in the container -->
     <feature name="activemq-client" description="ActiveMQ client libraries" version="${project.version}" resolver="(obr)" start-level="50">
-        <feature version="[4,5)">spring</feature>
+        <feature version="[4,6)">spring</feature>
```
jbonofre, line 28: This is not required and actually dangerous as ActiveMQ only supports Spring 4.x for now.
The only requirement in Karaf is to have `spring-legacy` features repo but it's already OK.

coheigea: OK will close.

cshannon: I added a comment to Jira but i'll but the same one here:

As a side note we should definitely support the latest Spring 5 for version 5.17.0. Spring 4.x is long been EOL

jbonofre: @cshannon I agree, but it's not just changing the feature. The "core" dependencies and schema should be updated as well. I've started this task on `master`.

PR #553, opened 2020-07-24
AMQ-8012 Fix ConcurrentModifictionException in SubQueueSelectorCacheBroker

mattrpav:
AMQ-8012

 - Remove non-thread-safe method getSelectors()
 - Update all references to use the thread-safe getSelectorsForDestination() method


cshannon:
The issue here seems to be that the selector Set itself is synchronized but the lock is not used when iterating. 

https://github.com/apache/activemq/blob/activemq-5.16.0/activemq-broker/src/main/java/org/apache/activemq/plugin/SubQueueSelectorCacheBroker.java#L150

The other method called getSelectorsForDestination() appears to copy into a new HashSet but that method also isn't safe either as it doesn't lock while creating the new HashSet (it will iterate and copy into the new set before returning)

cshannon:
Also, all the contains key stuff appears to not be threadsafe either. Multiple methods are checking if key exists and then getting but not checking for null.

cshannon:
I will merge this in and then fix the rest of it.

mattrpav:
@cshannon agreed, good catch! I was eyes on the caller and the dupe method that was returning a ref to the internal collection values.

PR #570, opened 2020-11-06
Change ActiveMQTextMessage to become serializable

tingung:
There is a business use case which need to persist ActiveMQTextMessage (or TextMessage) into database. Required ActiveMQTextMessage to become serializable. 
TextMessage implementation by both IBM Websphere and Weblogic are serializable. 
https://www.ibm.com/support/knowledgecenter/en/SSFKSJ_7.5.0/com.ibm.mq.javadoc.doc/WMQJMSClasses/com/ibm/jms/JMSMessage.html
https://docs.oracle.com/middleware/12213/wls/WLAPI/weblogic/jms/common/TextMessageImpl.html

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQTextMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQTextMessage.java
@@ -43,7 +43,7 @@ import org.apache.activemq.wireformat.WireFormat;
  *
  */
 public class ActiveMQTextMessage extends ActiveMQMessage implements TextMessage, Serializable {
-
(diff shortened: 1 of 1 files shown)


amitgoyal1310:
Can someone from the community please check this request ?

cshannon:
This is not necessary and is not going to work as there are many subtypes in those messages that are not serializable. You can use OpenWireFormat to marshall/unmarshall the data into bytes and persist it to a database.  See https://github.com/apache/activemq/blob/master/activemq-client/src/main/java/org/apache/activemq/openwire/OpenWireFormat.java

Also, in the future a Jira needs to be created and tests need to be added.

PR #571, opened 2020-11-10
Adding an option which allows messages to expire on active durable subscriptions

dhmgit:
When an *active* durable subscription fails to acknowledge messages due to being offline/unreliable/misbehaved, messages which expire according to their TTL are not marked expired on that subscription. This can lead to an unbounded number of messages building up in the message store, eventually exhausting quotas and causing backpressure. This PR introduces a new broker-level setting `enableMessageExpirationOnActiveDurableSubs` which allows messages to expire on *active* durable subscriptions.

We have been running with this setting enabled in quite a few production environments for a couple of years with no observed ill effects. Even so, the new setting is disabled by default to ensure backward compatibility. This is my first attempt to submit a PR to this repo, so please let me know if I need to provide additional information. 

I did not attempt to create a new JIRA issue as there are at least two issues out there already which appear to be related:
AMQ-6636
AMQ-7212

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -203,6 +203,7 @@ public class BrokerService implements Service {
     private final AtomicBoolean preShutdownHooksInvoked = new AtomicBoolean(false);
     private BrokerPlugin[] plugins;
     private boolean keepDurableSubsActive = true;
+    private boolean enableMessageExpirationOnActiveDurableSubs = false;
     private boolean useVirtualTopics = true;
     private boolean useMirroredQueues = false;
     private boolean useTempMirroredQueues = true;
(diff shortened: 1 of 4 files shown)


jbonofre (approved):

ncasaux:
Hello,

Could you please clarify this :
> When an active durable subscription fails to acknowledge messages due to being offline/unreliable/misbehaved

_offline_ seems to be the opposite of _active_, so does this apply when the durable subscriber is offline ? Thanks!

dhmgit:
@ncasaux , apologies for not seeing your comment earlier -- it was buried in a list of several hundred notifications.

This PR is from a while back and was itself based on work that I had done at least a couple years prior. This is to say that I might not be remembering all the details 100% correctly. 

We run .NET applications in on-premises customer environments that we do not control. Those applications use the NMS client library to connect to AMQ, and consume topics using durable subscriptions. In certain customer environments, these client applications can experience sustained unreliable network connectivity. The reasons for the flaky connectivity were varied, but the common result was that some clients would get into a state where they would fail to acknowledge messages. On the other hand, those same clients managed to reconnect (via the failover transport) often enough to keep their durable subscriptions active. Without the setting introduced by this PR, those unACKed messages continue to be referenced by those active subscriptions well after the message TTL expires. This was ultimately leading to quota exhaustion as those expired messages could never be purged.

I hope that helps!

ncasaux:
Thanks for the clarification! In your case, your subscription is actually _active_
So, from what you explain, I would understand your sentence this way:
> When an active durable subscription fails to acknowledge messages due to being ~~offline~~/unreliable/misbehaved

PR #608, opened 2021-01-12
[AMQ-8097] Deal with deserialization with xstream unmarshal poison ack

jbonofre:


jbonofre:
@coheigea mind to take a look ?

activemq-client/src/main/java/org/apache/activemq/util/ClassLoadingAwareObjectInputStream.java
```diff
@@ -40,7 +40,7 @@
     private final ClassLoader inLoader;
 
     static {
-        serializablePackages = System.getProperty("org.apache.activemq.SERIALIZABLE_PACKAGES","org.apache.activemq,org.fusesource.hawtbuf,com.thoughtworks.xstream.mapper").split(",");
+        serializablePackages = System.getProperty("java.lang,org.apache.activemq.SERIALIZABLE_PACKAGES","org.apache.activemq,org.fusesource.hawtbuf,com.thoughtworks.xstream.mapper").split(",");
```
coheigea, line 43: There's a typo here, it should be: "java.lang","

coheigea (changes requested):

activemq-broker/src/main/java/org/apache/activemq/plugin/SubQueueSelectorCacheBroker.java
```diff
@@ -369,7 +369,10 @@ public SubSelectorClassObjectInputStream(InputStream is) throws IOException {
 
         @Override
         protected Class<?> resolveClass(ObjectStreamClass desc) throws IOException, ClassNotFoundException {
-            if (!(desc.getName().equals("java.lang.String") || desc.getName().startsWith("java.util."))) {
+            if (!(desc.getName().startsWith("java.lang.")
+                    || desc.getName().startsWith("com.thoughtworks.xstream")
```
coheigea, line 373: Is there a way to check if xstream is enabled, and if so to include this check, otherwise leave it out?

activemq-client/src/main/java/org/apache/activemq/util/ClassLoadingAwareObjectInputStream.java
jbonofre: Oh yes, my bad ;) I'm fixing.

activemq-broker/src/main/java/org/apache/activemq/plugin/SubQueueSelectorCacheBroker.java
jbonofre: It could be enabled when using HTTP transport connector.

jbonofre: I think it's worth to keep it (just in case for http transport connector). Thoughts ?

jbonofre:
@coheigea I updated the PR, if you mind to do a new pass. Thanks !

coheigea (approved):

PR #612, opened 2021-01-15
[AMQ-8123] Add jaxb-core,istack,activation in lib for RuntimeConfigurationPlugin (with both JDK8 and 11+)

jbonofre:


lucastetreault:

PR #615, opened 2021-03-02
AMQ-6951 Hide jetty version

mattrpav:
Fixes:

* Console
* WS transport
* HTTP transport


lucastetreault:

jbonofre:
Hey @mattrpav . I guess the purpose is for security reason: to "hide" the Jetty server version for clients (and so avoid potential use of Jetty version header for a known attack). Correct ?

jbonofre:
OK I saw the discussion in the Jira, it makes sense. I don't think we need to do it configurable for the transport connector. I will test the PR rebased on `master` as I fixed the tests ;)

PR #617, opened 2021-03-07
[AMQ-8169] Do not unwrap in two threads concurrently

lucastetreault:
According to the [SSLEngine docs](https://docs.oracle.com/javase/8/docs/api/javax/net/ssl/SSLEngine.html) "two threads must not attempt to call the same method (either wrap() or unwrap()) concurrently, because there is no way to guarantee the eventual packet ordering". 

serviceRead(), the loop that calls unwrap(), was called in doInit() as well in initializeStreams() which intermittently leads to errors such as **javax.crypto.AEADBadTagException: Input too short - need tag** as described in the [associated Jira](https://issues.apache.org/jira/browse/AMQ-8169) as the two threads compete to unwrap network traffic.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/transport/nio/NIOSSLTransport.java
+++ b/activemq-client/src/main/java/org/apache/activemq/transport/nio/NIOSSLTransport.java
@@ -200,17 +200,7 @@ public class NIOSSLTransport extends NIOTransport {
     final protected CountDownLatch initialized = new CountDownLatch(1);
 
     protected void doInit() throws Exception {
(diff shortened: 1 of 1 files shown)


mattrpav:
Looks like there is a reason for serviceRead() to be in the doInit(). Looks like it was specifically added for the auto+nio+ssl use case and when used for bridges.  Start/Stop

ref: https://github.com/apache/activemq/commit/ed0e786b6002633411037923fb28a075489e442b

cshannon:
There might be a better way than what I did but if I recall auto+nio+ssl was having issues without this change. It's been a long time since i looked at it though

PR #618, opened 2021-03-07
AMQ-8048 ActiveMQ broker doesn't start with the error : Keystores with multiple certificates are not supported on the base class org.eclipse.jetty.util.ssl.SslContextFactory

charlie-cyf:
Changes:
- broker doesn't start because Jetty 9.4.31.v20200723 deprecated SslContextFactory with SslContextFactory$Server, similar issue: https://issues.apache.org/jira/browse/CALCITE-4095
- replace replace SslContextFactory with SslContextFactory$Server in jetty-demo.xml
- fix test BrokerXmlConfigStartTest for activemq-security.xml


charlie-cyf:
submit to wrong branch.. closing

PR #620, opened 2021-03-08
AMQ-8169: Synchronize on serviceRead inside NIOSSLTransport

cshannon:
This is needed to prevent concurrent access to the SSLEngine during
initialization. This is a regression from when auto+nio+ssl was added.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/transport/nio/AutoInitNioSSLTransport.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/transport/nio/AutoInitNioSSLTransport.java
@@ -158,8 +158,9 @@ public class AutoInitNioSSLTransport extends NIOSSLTransport {
         return readSize;
     }
 
+    //Prevent concurrent access to SSLEngine
     @Override
-    public void serviceRead() {
+    public synchronized void serviceRead() {
         try {
             if (handshakeInProgress) {
                 doHandshake();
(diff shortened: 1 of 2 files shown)


jbonofre:
Thanks, I'm testing on my machine as well.

jbonofre:
I guess the same should be applied to AMQ SSL test, I will tackle this if required.

jbonofre:
FYI, when I do:

```
mvn clean test -Dtest=Stomp11NIOSSLTest
```

I do the same again:

```
[ERROR] Failures: 
[ERROR]   Stomp11NIOSSLTest>Stomp11Test.testDurableSubAndUnSubOnTwoTopics:939
[INFO]
[ERROR] Tests run: 31, Failures: 1, Errors: 0, Skipped: 0
```

It's what I meant when the test is flaky.

jbonofre:
I'm testing on the PR now.

cshannon:
The main thing I was looking at was StompNIOSSLTest#testDisconnectDoesNotDeadlockBroker was breaking in CI and also locally for me and this PR fixed that at least. That other test you reference has always been flaky even before I was working on the auto transport stuff years ago.

jbonofre:
OK, so it's what I thought: it's not the same thing, and the build randomly fail (including on Jenkins). I will do a new test on those tests, eventually excluding them (or isolating in a profile).

cshannon:
Thanks for checking that.  The change seems like it should prevent any more of those errors by preventing concurrent access to the SSL engine. Also generally there won't be any contention after init so there should be very minimal overhead with uncontended locking so I wouldn't expect any noticeable performance hit.

jbonofre (approved):

PR #621, opened 2021-03-11
AMQ-8184 - Re-enable NIO tests

cshannon:
Re-enable the nio transport tests in activemq-unit-tests by default to
catch transport errors. Also fix broken tests


jbonofre:
At least the Stomp NIO SSL test is more stable now. Let me merge this one and I will move forward on other flaky tests.

PR #622, opened 2021-03-15
Add full.test profile to execute all modules and remove unit tests and itests by default

jbonofre:


pom.xml
```diff
@@ -1489,6 +1486,15 @@
 
   <profiles>
 
+    <profile>
+      <id>full.test</id>
+      <modules>
+        <module>activemq-unit-tests</module>
```
ehossack-aws, line 1492: Hi wondering if you would entertain my question here :)

Because I don't have a huge amount of context, this question might be naive, but shouldn't the unit tests be the _core_ thing run on a PR? 

In an ideal world I'd expect all unit tests to run, and a subset of integration tests relevant to the components changed. Obviously the latter part is hard, but is this not removing some of the confidence inspiring tests?

jbonofre: The problem is that unit tests execution takes 3 hours. I don't think it brings value to run a build for 3 or 4 hours on each PR. I agree to execute periodically (I proposed weekly), but be fast on PR.

ehossack-aws: Ah that makes sense. Kindof crazy it takes that long but I think this is the right approach!

ehossack-aws (approved):
Well, I'm just a new contributor here, but I'm in favour of taking incremental approaches to improve the PR process!

PR #630, opened 2021-03-19
Remove openwire generator

jbonofre:


mattrpav:
Hey JB-  curious as to why remove this?  The CMS and NMS sub-projects still active. Is there an alternate module for generating the openwire classes and structures for non-Java

jbonofre:
@mattrpav my bad, this PR was not expected (local test). Nevermind.

PR #633, opened 2021-03-21
Add github action

fpapon:


jbonofre:
I would say -1 for now to replace Jenkins with GitHub Actions. I think we should first cleanup the tests and have a faster build. Else, we will have bunch of GH Actions jobs, and potentially "banned".

fpapon:
@jbonofre ok make sense, let me close this PR and reopen it when needed.

PR #636, opened 2021-03-23
Forward commit commands for local transactions as well as XA transact…

jgallimore:
…ions during ACK compaction.

This PR will forward all commits with transaction info (as opposed to just XA transactions, which would not include commits for local transactions). I have included a unit test, and will open a discussion on the mailing list.

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
@@ -2230,18 +2230,11 @@ public abstract class MessageDatabase extends ServiceSupport implements BrokerSe
     }
 
     private boolean shouldForward(JournalCommand<?> command) {
-        boolean result = false;
-        if (command != null) {
-            if (command instanceof KahaRemoveMessageCommand) {
-                result = true;
-            } else if (command instanceof KahaCommitCommand) {
-                KahaCommitCommand kahaCommitCommand = (KahaCommitCommand) command;
-                if (kahaCommitCommand.hasTransactionInfo() && kahaCommitCommand.getTransactionInfo().hasXaTransactionId()) {
-                    result = true;
-                }
-            }
+        if (command == null) {
+            return false;
         }
-        return result;
+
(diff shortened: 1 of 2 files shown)


activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
```diff
@@ -2236,7 +2236,7 @@ private boolean shouldForward(JournalCommand<?> command) {
                 result = true;
             } else if (command instanceof KahaCommitCommand) {
                 KahaCommitCommand kahaCommitCommand = (KahaCommitCommand) command;
-                if (kahaCommitCommand.hasTransactionInfo() && kahaCommitCommand.getTransactionInfo().hasXaTransactionId()) {
+                if (kahaCommitCommand.hasTransactionInfo())  {
```
gtully, line 2239: we don't need that check for hasTransactionInfo any more, it was redundant before but is more so now. But it is not a biggie.

jgallimore: That does make sense - I guess we wouldn't see a commit command without transaction info. I'll remove that check.

activemq-unit-tests/src/test/java/org/apache/activemq/bugs/AMQ7067Test.java
```diff
@@ -526,6 +526,82 @@ public boolean isSatisified() throws Exception {
         }
     }
 
+    @Test
+    public void testForwardAcksAndCommitsWithLocalTransaction() throws Exception {
```
gtully, line 530: great test :-)

gtully (approved):

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
gtully: thanks. less is always more.

jgallimore:
I've (hopefully) simplified that shouldForward() method, per the comments. Thank you for the review, @gtully !

PR #638, opened 2021-03-23
Forward commit commands for local transactions as well as XA transact…

jgallimore:
…ions during ACK compaction

This is a backport of #636


cshannon:
@gtully - Any thoughts on this Gary? This change makes sense to me at first glance but it looks like you initially added the check for just XA transactions [here](https://github.com/apache/activemq/commit/f7ff4c25e13133132c3103141a5179c93e43b536). I was curious if there was a reason you didn't initially include local Tx as well?

jgallimore:
Closing this as it'll be cherry-picked to the other branches after being merged to master.

PR #642, opened 2021-03-29
fix BrokerServiceTest by replacing powermock

Charlie-chenchrl:
### Changes
- BrokerServiceTest in `activemq-unit-tests` is failing because powermock 1.7.4 doesn't work with newly updated Mockito 3.8.0. [Error Message](https://github.com/Charlie-chenchrl/activemq/runs/2216193106?check_suite_focus=true#step:9:1533)
- BrokerServiceTest used powermock to mock static functions, since Mockito start to support static mocks from 3.4.0 and this is the only place we use powermock, we can replace powermock with Mockito

### Test
- run `mvn clean test -Dtest=BrokerServiceTest`, test result in local CI: https://github.com/Charlie-chenchrl/activemq/runs/2216276748?check_suite_focus=true#step:7:195 (line 195)

### DIFF
--- a/activemq-unit-tests/pom.xml
+++ b/activemq-unit-tests/pom.xml
@@ -302,27 +302,7 @@
     </dependency>
     <dependency>
       <groupId>org.mockito</groupId>
-      <artifactId>mockito-core</artifactId>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
-      <groupId>org.powermock</groupId>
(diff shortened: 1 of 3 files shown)


jbonofre:
I'm testing the fix locally as `full.test` profile is not use on Jenkins by default.

jbonofre (approved):

PR #648, opened 2021-04-15
AMQ-8226 match any amount of spaces before activemq PID in bin/active…

Charlie-chenchrl:
…mq startup script

### Description
- fix for [AMQ-8226](https://issues.apache.org/jira/browse/AMQ-8226). 
- added `\s*` before `$activemq_pid` to match any amount of spaces in `ps -o` command


jbonofre (approved):
Thanks for pointing this !

PR #649, opened 2021-04-16
Revert "AMQ-7298 - rework redelivery message tracking to ensure no du…

jbonofre:
…plicate suppression (and dlq) in error for local transaction batches that failover"

This reverts commit 02548777c2954f5ccad918df8b1280fd601f8fd6.


jbonofre:
@gtully waiting deeper investigation, here's a simple revert of the commit causing redelivery broker plugin issue. Do you have time to investigate or do you want me to do it ? Thanks !

PR #650, opened 2021-04-20
AMQ-8220 add add-opens options for Java 16

jgallimore:


jbonofre:
Please open the PR only on `main`/`master` branch.

PR #656, opened 2021-05-24
AMQ-8277 Copy consumerinfo (removing the additional predicate) before…

jgallimore:
… sending the advisory message to trigger the subscription over the network of brokers.

If a consumer uses an additionalPredicate (as part of a plugin, for example to filter messages to the consumer), the additionalPredicate will cause a ClassCastException and messages will not be forwarded across the network for this consumer.

This PR creates a copy of the ConsumerInfo object (which does *not* copy the additional predicate), and uses the copy of the ConsumerInfo object to send the advisory message to trigger the subscription across the network.

One (potential) drawback here is that the subscription between the brokers is does not filter the messages. 

I did attempt to add this, by making a wrapper that implements both BooleanExpression and DataStructure, and could be marshalled into OpenWire, and then checking it in NetworkBridgeFilter. This did work, but appeared to introduce a number of regressions across the test suite. I'm very happy to take a another swing at this approach if it is preferred.

I have run the full test suite across this PR, and it doesn't appear to break any other tests on my setup.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/advisory/AdvisoryBroker.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/advisory/AdvisoryBroker.java
@@ -148,7 +148,13 @@ public class AdvisoryBroker extends BrokerFilter {
             } finally {
                 consumersLock.writeLock().unlock();
             }
-            fireConsumerAdvisory(context, info.getDestination(), topic, info);
+
+            // Note: info.copy() does not copy the ConsumerInfo.additionalPredicate, which cannot be serialized via
+            // OpenWire marshallers, unless it implements org.apache.activemq.command.DataStructure (which regular
+            // org.apache.activemq.filter.Expression objects do not)
+            final ConsumerInfo clonedConsumerInfo = info.copy();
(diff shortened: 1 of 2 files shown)


cshannon:
The main issue I see here is what you pointed out...while this prevents the error it would result in not filtering between brokers may result in unintended behavior which isn't great either and the only real way to fix it is to make the predicate be included by OpenWire (maybe another version or what you already tried)

For the topic scenario (and using conduit subscriptions) it would work fine as selectors are not used anyways as it is pub/sub and messages are just broadcast so when the messages hit the different brokers they can be filtered there.

When using a queue with multiple consumers I would think this would lead to stuck messages. For example a scenario might be if you had 3 brokers (A, B, C) and you create consumers on A and B with different predicates but those predicates are now filtered out when creating demand with advisories. That means if a message is published to broker C that could would match the predicate on consumer on A but not B it may still be forwarded to B and now stuck.

I'm not really sure how much it matters as I doubt many people are doing this as this issue would have been brought up more often. This fix would probably be ok as long as we update the documentation on the network of brokers page and add something that specifically points out that if that field is set a network of brokers ignores it.

jbonofre:
Auto closed

PR #663, opened 2021-05-27
[AMQ-8284] Update jetty to 9.4.41.v20210516

mattrpav:
- Confirm update w/ deprecation fixes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,7 +69,7 @@
     <jackson-databind-version>2.9.10.8</jackson-databind-version>
     <jasypt-version>1.9.3</jasypt-version>
     <jaxb-bundle-version>2.2.11_1</jaxb-bundle-version>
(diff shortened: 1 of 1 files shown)


jbonofre (changes requested):
Can you please amend your commit with the right Jira number ? It's AMQ-8284.

jbonofre (approved):

jbonofre:
For the record, you can rebase and squash, and then force push to have a single commit instead of two.

mattrpav:
Will do.  Was looking into Jetty 10 and Jettty 11 for a hot minute-- but we'd need a JDK v11 min required

jbonofre:
It's too early to upgrade to jetty 10 or jetty 11 as it will break the third party (Karaf), and also API change. I would target this once Pax Web/Karaf will support it.

PR #666, opened 2021-06-02
AMQ-8287 - fix NIOSSLTransport deadlock with serviceRead lock

cshannon:
This narrows the lock that was added to serviceRead() to secureRead()
which prevents processing commands while locked which should solve the
deadlock issues

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/transport/nio/AutoInitNioSSLTransport.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/transport/nio/AutoInitNioSSLTransport.java
@@ -158,9 +158,8 @@ public class AutoInitNioSSLTransport extends NIOSSLTransport {
         return readSize;
     }
 
-    //Prevent concurrent access to SSLEngine
     @Override
-    public synchronized void serviceRead() {
+    public void serviceRead() {
         try {
             if (handshakeInProgress) {
                 doHandshake();
--- a/activemq-client/src/main/java/org/apache/activemq/transport/nio/NIOSSLTransport.java
+++ b/activemq-client/src/main/java/org/apache/activemq/transport/nio/NIOSSLTransport.java
@@ -243,9 +243,8 @@ public class NIOSSLTransport extends NIOTransport {
         }
     }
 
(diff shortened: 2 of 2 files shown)


cshannon:
@jbonofre and @mattrpav - I think this will do the trick. I did some testing and the StompNIOSSLTest is still fixed with this modified fix. It's pretty straight forward as you can see.

mattrpav (approved):
LGTM. Moving the locking down makes more sense and removes the deadlock. 

During review, I did see a couple areas of improvement. Added JIRAs to address those.

ref: https://issues.apache.org/jira/browse/AMQ-8288
ref: https://issues.apache.org/jira/browse/AMQ-8289

cshannon:
@mattrpav - agreed, there are still thread safety issues to address but narrowing this lock to secureRead() at least fixes the deadlock problem while fixing the StompNIOSSLTest issue still so it's still better than it was. It makes sense to take a look in future Jiras to improve things.

I will go ahead and merge this.

mattrpav:
I think we should consider a v5.15.16 b/c of this. Thoughts?

cshannon:
Is there any particular reason someone using 5.15.15 can't upgrade to 5.16.3?

mattrpav:
Nothing other than feels weird to me to leave a deadlock bug in an otherwise fairly stable release series.

cshannon:
Fair enough, nothing really against doing a release other than we said 5.15.15 was the last one. You have to draw a line somewhere but I would be ok with 5.15.16 with a limited scope (ie just fixing this issue or something) if we wanted to do that just because deadlocks are a major issue. I don't want to pull a bunch of other stuff into 5.15.16 if we did it as it's really time to get people to start upgrading.

jbonofre:
I didn't have time yesterday to do the review, but you have my +1 ;) Thanks !

cshannon:
@jbonofre - thanks, @mattrpav took a look obviously so I went ahead and merged it. There's still work to be done as seen in the new Jiras opened but this fix should at least prevent the deadlock issue.

jbonofre:
@cshannon no problem ! You did well ! I'm rebasing/fixing log4j2 update.

PR #669, opened 2021-06-09
Improved lru-k cache

domson-tech:
improved lru-k cache based on simple lrucache and fifolist, which improved cache hit rate


mattrpav:
@domson-tech interesting patch. Do you have performance numbers you can share comparing this implementation vs the existing?

Creating a JIRA would be helpful, so it shows up on other dev's radar to get eyeballs for review.

Thanks!

jbonofre:
@domson-tech Thanks for the PR. Can you please share kind of abstract about this K-LRU cache on the client side ? What's the purpose/abstract, how to configure it, ... ? Thanks !

jbonofre:
Auto closed

PR #679, opened 2021-07-08
[AMQ-7514] Deprecate configuration that uses non-inclusive language

ehossack-aws:
This is the first in a series of changes that are planned to phase out non-inclusive terminology throughout the config, and codebase.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -153,9 +153,6 @@ public class BrokerService implements Service {
     private boolean useLoggingForShutdownErrors;
     private boolean shutdownOnMasterFailure;
     private boolean shutdownOnSlaveFailure;
-    private boolean waitForSlave;
-    private long waitForSlaveTimeout = DEFAULT_START_TIMEOUT;
-    private boolean passiveSlave;
     private String brokerName = DEFAULT_BROKER_NAME;
     private File dataDirectoryFile;
     private File tmpDataDirectory;
@@ -517,7 +514,6 @@ public class BrokerService implements Service {
         if (startDate == null) {
             return 0;
         }
-
         return new Date().getTime() - startDate.getTime();
     }
 
@@ -771,7 +767,7 @@ public class BrokerService implements Service {
      * delegates to stop, done to prevent backwards incompatible signature change
      */
     @PreDestroy
-    private void preDestroy () {
+    private void preDestroy() {
         try {
             stop();
         } catch (Exception ex) {
@@ -2905,42 +2901,6 @@ public class BrokerService implements Service {
      */
     public void setShutdownOnSlaveFailure(boolean shutdownOnSlaveFailure) {
         this.shutdownOnSlaveFailure = shutdownOnSlaveFailure;
-    }
-
-    public boolean isWaitForSlave() {
-        return waitForSlave;
-    }
-
-    /**
-     * @org.apache.xbean.Property propertyEditor="org.apache.activemq.util.BooleanEditor"
-     */
-    public void setWaitForSlave(boolean waitForSlave) {
-        this.waitForSlave = waitForSlave;
-    }
-
-    public long getWaitForSlaveTimeout() {
-        return this.waitForSlaveTimeout;
-    }
-
-    public void setWaitForSlaveTimeout(long waitForSlaveTimeout) {
-        this.waitForSlaveTimeout = waitForSlaveTimeout;
-    }
-
-    /**
-     * Get the passiveSlave
-     * @return the passiveSlave
-     */
-    public boolean isPassiveSlave() {
-        return this.passiveSlave;
-    }
-
-    /**
-     * Set the passiveSlave
-     * @param passiveSlave the passiveSlave to set
-     * @org.apache.xbean.Property propertyEditor="org.apache.activemq.util.BooleanEditor"
-     */
-    public void setPassiveSlave(boolean passiveSlave) {
-        this.passiveSlave = passiveSlave;
     }
 
     /**
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -151,8 +151,8 @@ public class BrokerService implements Service {
 
     private boolean useShutdownHook = true;
     private boolean useLoggingForShutdownErrors;
-    private boolean shutdownOnMasterFailure;
-    private boolean shutdownOnSlaveFailure;
+    private boolean shutdownOnActiveFailure;
+    private boolean shutdownOnStandbyFailure;
     private String brokerName = DEFAULT_BROKER_NAME;
     private File dataDirectoryFile;
     private File tmpDataDirectory;
@@ -482,7 +482,7 @@ public class BrokerService implements Service {
     }
 
     public void masterFailed() {
-        if (shutdownOnMasterFailure) {
+        if (shutdownOnActiveFailure) {
             LOG.error("The Master has failed ... shutting down");
             try {
                 stop();
@@ -609,7 +609,7 @@ public class BrokerService implements Service {
                 }
             }
 
-            // in jvm master slave, lets not publish over existing broker till we get the lock
+            // in jvm active standby, lets not publish over existing broker till we get the lock
             final BrokerRegistry brokerRegistry = BrokerRegistry.getInstance();
             if (brokerRegistry.lookup(getBrokerName()) == null) {
                 brokerRegistry.bind(getBrokerName(), BrokerService.this);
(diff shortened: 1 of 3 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -152,10 +152,15 @@
     private boolean useShutdownHook = true;
     private boolean useLoggingForShutdownErrors;
     private boolean shutdownOnMasterFailure;
+    private boolean shutdownOnActiveFailure;
```
michaelandrepearce, line 155: Its a private field no need to duplicate just rename the original field. Just need to keep method compatibility

michaelandrepearce: This comment applies generally. No need for me to repeat on every occurance

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -485,15 +490,15 @@ public JmsConnector removeJmsConnector(JmsConnector connector) {
     }
 
     public void masterFailed() {
-        if (shutdownOnMasterFailure) {
-            LOG.error("The Master has failed ... shutting down");
+        if (shutdownOnActiveFailure) {
+            LOG.error("The Active has failed ... shutting down");
```
michaelandrepearce, line 494: Active doesnt sound like a good replacement master
 Typically we would refer to master or slave being the current active node. 

Active = state

Master / slave was what role the node took. 

I think a different word to replace master should be found

michaelandrepearce (changes requested):
Comments inline

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
ehossack-aws: Hmm good point, my brain didn't catch that as my first pass was assuming spring looked at the actual variable names.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
ehossack-aws: The [ticket](https://issues.apache.org/jira/browse/AMQ-7514) mentions active/standby and I'd tend to agree that it's the clearest (it's also what we use at AWS).

Consider:
- of the nodes, only 1 is _active_ serving requests
- the other nodes are not pure replicas, as they have shared storage, rather they are on standby over the same FS

It's possible that in the previous replicated leveldb one could think of replica as a good term here, but I think it's slightly disingenuous. (especially if the storage used isn't actually replicated...)

michaelandrepearce: I disagree with the ticket then
Active is state though. Either master or slave can be active. We will end up with very confusing terminology.

https://www.theserverside.com/opinion/Master-slave-terminology-alternatives-you-can-use-right-now

michaelandrepearce: Quickly reading the ticket actually mentions

New terms:
a. For shared storage: ‘active’ and ’standby’
b. For replication: ‘primary’ and ‘replica'
c. For 'white list' and 'blacklist': 'allow list' and 'deny list'

This is down to replication state so primary replica should be used

jbertram: I agree with @michaelandrepearce. I've added a comment to AMQ-7514 to push along the discussion.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -2896,51 +2922,142 @@ public void setSslContext(SslContext sslContext) {
...
+     */
+    public void setShutdownOnStandbyFailure(boolean shutdownOnStandbyFailure) {
+        this.shutdownOnSlaveFailure = shutdownOnStandbyFailure;
+        this.shutdownOnStandbyFailure = shutdownOnStandbyFailure;
+    }
+
+    /**
+     * @deprecated AMQ-7514 in the move to inclusive nomenclature
+     */
+    @Deprecated
```
mattrpav, line 2961: This method is never called in the code base. My understanding is that it was kept to not break configurations from the long-gone active-active replication. I would contend that since we have agreed to remove leveldb from 5.17.x and there are no planned usages for these metods, that this method (and related ones) could safely be removed instead of deprecated.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -485,15 +490,15 @@ public JmsConnector removeJmsConnector(JmsConnector connector) {
...
     public void masterFailed() {
-        if (shutdownOnMasterFailure) {
-            LOG.error("The Master has failed ... shutting down");
+        if (shutdownOnActiveFailure) {
+            LOG.error("The Active has failed ... shutting down");
             try {
                 stop();
             } catch (Exception e) {
-                LOG.error("Failed to stop for master failure", e);
+                LOG.error("Failed to stop for active failure", e);
```
mattrpav, line 498: Updating log messages is tricky, since it can break monitoring in not-so-easy-to-upgrade ways. I think this should be a sub-task conversation on the parent ticket.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
mattrpav: This group of methods is never read anywhere by the code base. My understanding is that they are legacy hold-over from early replication approach that has been removed long ago. I do not see any reason to deprecate them. Given that we have agreed to make major breaking changes in 5.17.x (remove leveldb), I believe we can safely remove these methods in this pass.

We should not be adding new terms or concepts that do not apply to the code base in this pass.

ehossack-aws: I did a 'find references' search and yeah, you're right @mattrpav. I think this is also relevant in the discussion about active/passive/standby. It seems reasonable in the scope of this change rather than renaming to actually remove those methods/concepts. Then there is no need to consider the 'existing' terminology.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
ehossack-aws: How about in the scope of this PR we add both messages, and then the subtask is to remove the non-inclusive messages? That would support a transition period well.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -2896,51 +2922,142 @@ public void setSslContext(SslContext sslContext) {
...
+     * Configuration `passiveSlave` is deprecated and will be removed in a future release. Use `standbyInstance` instead.
+     * @deprecated AMQ-7514 in the move to inclusive nomenclature
      * @param passiveSlave the passiveSlave to set
      * @org.apache.xbean.Property propertyEditor="org.apache.activemq.util.BooleanEditor"
      */
+    @Deprecated
     public void setPassiveSlave(boolean passiveSlave) {
         this.passiveSlave = passiveSlave;
+        this.standbyInstance = passiveSlave;
+        LOG.warn("Configuration `passiveSlave` is deprecated and will be removed in a future release. Use `standbyInstance` instead.");
```
ehossack-aws, line 3042: @mattrpav per your comment it also appears `isPassiveSlave` is not used. We could also remove this it seems like?

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
mattrpav: 2 messages could yield 2 alerts. I think we break this up into some sub-tasks and add a broker config param to flip b/w the log messages using deprecated terms.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
mattrpav: Yes, I think we break-up the PR's a bit. Removing deprecated methods can be its own thing

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
ehossack-aws: as per @mattrpav 's AMQ-8316, moved the two concretely deprecated messages out of this PR

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
ehossack: This was addressed in latest version of PR

ehossack (commented):
The base commit https://github.com/apache/activemq/pull/700 and this PR can both be merged without any merge conflicts, pending approval of course.

jbonofre:
Auto closed

PR #700, opened 2021-08-20
[AMQ-8316] Remove unused BrokerService methods

ehossack-aws:
This removes passiveSlave and waitForSlave which are no longer used

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -153,9 +153,6 @@ public class BrokerService implements Service {
     private boolean useLoggingForShutdownErrors;
     private boolean shutdownOnMasterFailure;
     private boolean shutdownOnSlaveFailure;
-    private boolean waitForSlave;
-    private long waitForSlaveTimeout = DEFAULT_START_TIMEOUT;
-    private boolean passiveSlave;
     private String brokerName = DEFAULT_BROKER_NAME;
     private File dataDirectoryFile;
     private File tmpDataDirectory;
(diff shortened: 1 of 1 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -153,9 +153,7 @@
     private boolean useLoggingForShutdownErrors;
     private boolean shutdownOnMasterFailure;
     private boolean shutdownOnSlaveFailure;
-    private boolean waitForSlave;
     private long waitForSlaveTimeout = DEFAULT_START_TIMEOUT;
```
mattrpav, line 156: If waitForSlave can be removed, can waitForSlaveTimeout also be removed?

mattrpav (commented):
We probably should create a web page for 5.17.0 changes. This is going to remove config parameters from the XSD and generated xbean properties.

Section:

ehossack-aws:
> We probably should create a web page for 5.17.0 changes. This is going to remove config parameters from the XSD and generated xbean properties.

I'll create a parallel PR for that later this week.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
ehossack-aws: I thiiiink so. I recall going through all the properties, but maybe I missed this one because the name was so similar - I can't find any references to it in the codebase or where an embedded broker might use it.

jbonofre (approved):

ehossack-aws:
@mattrpav , @jbonofre I created a pull on the website to begin collecting changes for 5.17 - https://github.com/apache/activemq-website/pull/63

jbonofre:
@ehossack-aws it makes sense, thanks. I will take a look then.

jbonofre:
It will be superseeded by another PR related to AMQ-9349

PR #702, opened 2021-08-23
[AMQ-8359] Update slf4j to 1.7.32

mattrpav:


jbonofre:
FYI, It would need Karaf 4.3.3 with Pax Logging 2.0.10 to fully work with OSGi. I will also check the import range, it should work already.

mattrpav:
> FYI, It would need Karaf 4.3.3 with Pax Logging 2.0.10 to fully work with OSGi. I will also check the import range, it should work already.

Are you saying, we need to update the activemq-karaf / activemq-karaf-itests to use karaf 4.3.3?  Or is this just a note that minimum version of running 5.17.x will be Karaf 4.3.3?

jbonofre (approved):

PR #710, opened 2021-09-13
Update JaasAuthenticationBroker.java

chris-joys:
Allow user use encrypt password to check entry detail info via web console

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/security/JaasAuthenticationBroker.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/security/JaasAuthenticationBroker.java
@@ -18,6 +18,7 @@ package org.apache.activemq.security;
 
 import java.security.Principal;
 import java.security.cert.X509Certificate;
+import java.util.Properties;
 import java.util.Set;
 
 import javax.security.auth.Subject;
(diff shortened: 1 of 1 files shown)


activemq-broker/src/main/java/org/apache/activemq/security/JaasAuthenticationBroker.java
```diff
@@ -86,6 +88,10 @@ public void addConnection(ConnectionContext context, ConnectionInfo info) throws
     @Override
     public SecurityContext authenticate(String username, String password, X509Certificate[] certificates) throws SecurityException {
         SecurityContext result = null;
+        Properties prop = new Properties();
+    	prop.put(username, password);
```
mattrpav, line 92: Please fix inconsistent spacing

jbonofre:
It would be great to provide some explanation about this change:
1. why using `decrypt` on properties to get the password. My understanding is that you are using encrypted password (instead of plain)
2. If you are using encrypted password, the JAAS plugin should be configured by a property to define if the password is encrypted or not, else I think it will break all users using plain password.

jbonofre (changes requested):
Please answer to the question in my previous comment, and introduce a property enabling encrypted passwords.

PR #711, opened 2021-09-14
AMQ-8275 - Java 16 SSL - reflection warning

jgallimore:
This PR avoids the need to use reflection to explicitly set the hostname on the socket, instead providing the hostname via a SSLParameter. I have tested this using a HAProxy proxy server to route requests from a client to a broker based on the host name.


pax95:
Is there any news any news on this issue since it blocks for update to java 17 ?

jbonofre:
FYI, I'm working on it. I'm fixing assembly build and then I will rebase/merge this PR.

jbonofre (approved):

pax95:
@jgallimore @jbonofre Great job taking care of this. Thanks and merry christmas !!

iPet:
@jbonofre Any ideas on when we can expect a new release including this fix?

jbonofre:
@iPet 5.16.4 vote will start soon (hopefully during the week end or early next week).

fabiojuniork13:
@jbonofre, hello!
I have the same problem, but I am using the openjdk 18. The lasted version of openjdk.
Do you know how to solve this problem, please? I think this is because I am using the openjdk, because in another environment with oracle jdk this problem not happen.

PR #712, opened 2021-09-15
[AMQ-8385] Update to target-version for JDK 11

mattrpav:


jbonofre:
It seems maven metadata are corrupted for activemq-ra. I gonna check and fix.

gemmellr:
Using the compiler plugin standard `maven.compiler.source` and `maven.compiler.target` properties would allow simplifying, removing need for additional compiler plugin config that currently references the 'source-version' and 'target-version' properties.

Additionally using `maven.compiler.release` would have it set the javac -release flag and ensure it actually uses the appropriate target signatures for compilation rather than those of whatever JDK it happens to be running on. That may not be possible or need additional work though if the build is accessing stuff the -release flag doesnt permit (it restricts to only using the supported public APIs of the specified version)

pom.xml
```diff
@@ -115,8 +115,8 @@
     <felix-framework-version>5.6.12</felix-framework-version>
 
     <site-repo-url>scpexe://people.apache.org/www/activemq.apache.org/maven/</site-repo-url>
-    <source-version>1.8</source-version>
-    <target-version>1.8</target-version>
+    <source-version>8</source-version>
```
ehossack-aws: Forgive my ignorance, but why do we set this source-version so low?

gemmellr: I am guessing either just because no 11-level source features are actually being used yet, or alternatively perhaps some things actually start failing to compile if it is bumped, even if they do work when its already targetting 11, because the compiler it starts enforcing some things it didnt while source level remains set to 8. (That happened with the changes I did for Artemis, 8+11 worked, 11+11 needed changes).

mattrpav: @gemmellr yep. Setting to source-version 11 causes the module export errors to show up, since the build hasn't set the module exports yet.

jbonofre (approved):

PR #713, opened 2021-09-21
upgrade jaxb-basics-version to 0.12.0

ehossack-aws:
A dependency bot recommended bumping this version. No breaking changes.
I see no reason why not :)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,7 +107,7 @@
     <xstream-version>1.4.17</xstream-version>
     <xbean-version>4.20</xbean-version>
     <xerces-version>2.12.0</xerces-version>
-    <jaxb-basics-version>0.6.4</jaxb-basics-version>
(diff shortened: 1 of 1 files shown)


mattrpav:
This requires the JDK 9+ target-version.  This dep is why I created #712

mattrpav:
Yep, I might have overstated jaxb2-basics went to 9+ for build. This also requires jaxb-2.3.x api's and impl jars. I believe the fix is more than just update this module. Some of the downstream modules may need 9+. 

What did you test?

jbonofre:
FYI, jaxb-basics is used in the Karaf features and the runtime config. Did you test these parts ? For Karaf, it should not be a problem.

jbonofre (approved):
I tested in Karaf and thanks to the dependency flag, jaxb basics capabilities are already provided by Karaf. I also tested the runtimeConfigurationPlugin (which uses jaxb basics) and it works fine.

PR #714, opened 2021-09-23
AMQ-8317 add config param to toggle inclusive terms in logs

ehossack-aws:
As suggested in #679 , this adds a log toggle config to allow users to opt back in to the non-inclusive terminology previously used, as the messages may be used in monitoring/alerting.


ehossack-aws:
Not super sure the best way to chain commits without a target branch, so linking the new commit in this pr: 
https://github.com/apache/activemq/pull/714/commits/3f7eb0a39ba3e7926e72f91fa3f9f6748e261726

mattrpav:
Please hold on merge, need to review the terminology.

jbonofre:
To be honest, I'm not a big fan of having a configuration for that.
I think we can just go for the change for ActiveMQ 5.17.x and keep as it is on 5.16.x.

cshannon:
Agree with @jbonofre . I think having a configuration makes no sense. Just change it but at this point it should be target for 5.18 as 5.17 should be frozen except for critical stuff

jbonofre:
It will superseded by another PR for 7.0.0 with directly terminology change.

PR #720, opened 2021-10-13
[AMQ-8400] Add synchronization handling around synchronizations to pr…

mattrpav:
…event CME


mattrpav:
Internal tests validation:
500M+ messages sent to 1 queue
400x producers
400x consumers
1-transaction-per-message commit strategy

PR #724, opened 2021-11-16
[AMQ-8397] Add a destination option to sendDuplicateFromStoreToDLQ

mattrpav:
- Default 'true' to match existing behavior
 - Added counter to DestinationView


mattrpav:
I think this should be set to 'false' as default for 5.17.0.

PR #726, opened 2021-11-19
Fix for removed java.xml.bind dependency in Java 11

ehossack:
Without this change, in my local clone of the project (and Intellij) fails to build because there's no shipped version of `DataTypeConverter` that's referenced.

### DIFF
--- a/activemq-web/pom.xml
+++ b/activemq-web/pom.xml
@@ -122,6 +122,11 @@
       <groupId>com.thoughtworks.xstream</groupId>
       <artifactId>xstream</artifactId>
     </dependency>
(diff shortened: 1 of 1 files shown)


mattrpav:
What is the error you get? I build w/ Java 11 without this patch-- so does Apache's Jenkins.

jbonofre:
Java11 build is OK on my boxes. Furthermore, as JDK9+ doesn't provide jaxb anymore, the dependency is actually required.
IMHO, I would -1 this change.

jbonofre:
Tested and I confirm that the dependency is required for JDK11+

PR #730, opened 2021-12-01
[AMQ-8409] Fixed Unexpected \\r instead of \r in header property in i…

fscrec:


mattrpav:
@fscrec care to share the close reason? ie is this a non-issue?

fscrec:
@mattrpav Hi Matt, I just needed to make a change on the code - the fix is only required for STOMP 1.2 and not for other versions. The new pull request has been already added containing the updated code.

PR #731, opened 2021-12-02
[AMQ-8409]  Unexpected \\r instead of \r in header property in incoming messages when using STOMP 1.2 

fscrec:


mattrpav (changes requested):
Would you add a negative test to STOMP 1.0 and 1.1 version tests?

mattrpav:
FYI-- I kicked off new test run on Jenkins

mattrpav:
@fscrec thanks!

mattrpav (changes requested):
@fscrec the code fixes look great. One minor administrative change-- can you make the PR against 'main', so we can cherry pick it to other branches? I think this will allow it to get the unit tests to pass.

fscrec:
@mattrpav I've created a new pull request #736 for the main branch containing only three commits contained here (but from the main branch as the origin as opposed to 5.16.x I used in this PR). Should this PR be closed?

jbonofre:
For the record, you can squash the commits in one and focus only on `main` branch (then we cherry pick on other branches).

mattrpav:
@fscrec yep, I'll close this one. Thanks for the PR!

PR #743, opened 2021-12-28
Fix webconsole producers/consumers page

ygubernatorov:
By quoting queue names.

Issue:
Users are unable to see producers/consumers for that queue in console if queue name contains ":" (and some other symbols). So users get [frustrated](https://stackoverflow.com/questions/44817892/error-while-viewing-queue-consumers) for several years now.

The problem is that activemq/web uses jmx to query for consumers and producers of a queue.
And jmx [does not allow ](https://docs.oracle.com/javase/7/docs/api/javax/management/ObjectName.html) colons and some other symbols in queries. Queue name (and broker name) are _values_ and thus can (and should) be quoted.


jbonofre (approved):
I will merge by hand in order to squash the two commits in a single one.

PR #749, opened 2022-01-25
StatisticsBrokerPlugin: Add feature: request destination firstMessageTimestamp

stolsvik:
_(To main branch)_
Adding a feature (STATS_FIRST_MESSAGE_TIMESTAMP) to the StatisticsBrokerPlugin's destination-statistics for getting the
timestamp of the first message in the destination(s) being requested: If you on the query-message set the property StatisticsBroker.STATS_FIRST_MESSAGE_TIMESTAMP to anything (e.g. boolean true), a long value "firstMessageTimestamp" will be added to the statistics reply message(s). Since the reply message has JMSTimestamp set, which is the broker's now-timestamp, you may also on the query side calculate the age of the first message in milliseconds. The key name was chosen since that is the name of the corresponding feature in Artemis.

This extension of the existing feature is implemented to be as non-intrusive as possible, adding very little runtime cost if not requested. It also seems like the runtime cost for enabling this feature, thus finding and adding the firstMessageTimestamp, is small.

While at it, also slightly improving an existing feature (STATS_DENOTE_END_LIST) where a reply to a destination query can be
"null terminated": After sending the relevant replies, the StatisticsBroker also sends an empty message. This feature is relevant
if the query is a wildcard query, thus returning multiple messages: The empty message denotes the end of the replies. However, to activate this feature, a somewhat complicated query destination had to be constructed. Adopting the solution for the other StatisticsBroker feature where you may reset the broker statistics by adding a property to the query message, this null-termination feature now /also/ checks for the presence of this query modifier STATS_DENOTE_END_LIST as a property. (This property based solution was thus also adopted for the present 'firstMessageTimestamp' solution, as it was found much more intuitive).

Added tests for both the STATS_FIRST_MESSAGE_TIMESTAMP query modifier, and the improved STATS_DENOTE_END_LIST property-based query modifier.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Topic.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Topic.java
@@ -654,7 +654,7 @@ public class Topic extends BaseDestination implements Task {
         return result.toArray(new Message[result.size()]);
     }
 
-    private void doBrowse(final List<Message> browseList, final int max) {
+    public void doBrowse(final List<Message> browseList, final int max) {
         try {
             if (topicStore != null) {
                 final List<Message> toExpire = new ArrayList<Message>();
--- a/activemq-broker/src/main/java/org/apache/activemq/plugin/StatisticsBroker.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/plugin/StatisticsBroker.java
@@ -18,6 +18,8 @@ package org.apache.activemq.plugin;
 
 import java.io.File;
 import java.net.URI;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 
 import javax.jms.JMSException;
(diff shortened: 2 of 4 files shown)


mattrpav:
Great idea! I think we should swizzle the concept to be metrics on the destination itself. This would provide value when the queue is empty as well.

firstMesageTimestamp - the timestamp of the first message to traverse the destination
lastMessageTimestamp - the timestamp of the most recent message to traverse the destination (aka 'message age' when queue is non-zero)

ref: [AMQ-8463] (https://issues.apache.org/jira/browse/AMQ-8463)

stolsvik:
Just for posterity: As discussed on the dev mailing list: The suggestion of moving this to the destination will not give the sought effect. You want the timestamp of the actual message laying in front of the queue (at "head" position, _the next to be dequeued_), and if there are no such message at head, then there is no timestamp of a first message either, since the queue is empty.

The "last message to leave the broker" does not give the same monitoring effect that is wanted.

jbonofre:
I'm reviewing it.

jbonofre (approved):

jbonofre:
I think it's a good first step. We can improve later. I'm merging.

PR #755, opened 2022-02-04
Unit tests for AMQ-8053

lucastetreault:
Add some unit tests for https://issues.apache.org/jira/browse/AMQ-8053.

### DIFF
--- a/activemq-broker/src/test/java/org/apache/activemq/broker/UserIDBrokerTest.java
+++ b/activemq-broker/src/test/java/org/apache/activemq/broker/UserIDBrokerTest.java
@@ -1,0 +1,149 @@
+package org.apache.activemq.broker;
+
+import junit.framework.TestCase;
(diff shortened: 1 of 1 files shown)


jbonofre:
Are you sure about this test ? At first glance, it doesn't seem fully correct to me.

lucastetreault:
Any specific concerns? The first two demonstrate the existing functionality and the 3rd shows the change to not override it when the message goes through a network of brokers. Seems correct to me 🙂

mattrpav:
@lucastetreault appreciate the PR work here =). However, given it is my change, I need to do the tests. 

New PR  [758](https://github.com/apache/activemq/pull/758) re-uses network connector test framework and puts the unit tests in activemq-unit-tests instead of activemq-broker.

lucastetreault:
¯\_(ツ)_/¯

PR #756, opened 2022-02-05
[AMQ-8412] Update client-side maxFrameSize handling to be more symetr…

mattrpav:
…ical with server-side

 - Handle in the OpenWireFormat class
 - Add unit tests to confirm
 - Verify compression is accounted for
 - Verify the ability to disable using wireFormat.verifyMaxFrameSize=false


cshannon:
@mattrpav - Thanks for the quick PR, I should be able to take a look first thing Monday morning

cshannon:
@mattrpav - Overall this is a big improvement over the previous version using getSize(). I verified the sizing and the reported frame sizes when checked on the client side vs server are now quite close (not exact but it's close enough)

I see a few things to change:

1. There needs to be tests for all the transports. You need to also test the normal transports (TCP, SSL, etc) and not just NIO.
2. the nio and nio+ssl transports actually check the frame size in a different location on the server than the OpenWireFormat so the NIOSSLTransport and NIOTransport should be updated to also check the new flag. See: https://github.com/apache/activemq/blob/main/activemq-client/src/main/java/org/apache/activemq/transport/nio/NIOSSLTransport.java#L338 and https://github.com/apache/activemq/blob/main/activemq-client/src/main/java/org/apache/activemq/transport/nio/NIOTransport.java#L142
3. I would change the isAssignableFrom() to an instanceof check as it's slightly fast (see the inline comment)
4. You should write edge case tests. For example, we should test all 3 scenarios: frame size is enabled on both client/server, frame size is only enabled on the server and frame size is only enabled on the client.
5. You should write a test to verify that the framesize flag is not negotiated over open wire to prevent errors in the future or someone changing it. Ie make sure that if either side changes the flag the other side doesn't have it's flag changed.
6. Documentation needs to be added both to the code and to the website that makes it very clear that the flag won't be negotiated and only applies where it is set. https://activemq.apache.org/configuring-wire-formats
7. I think we can probably just rename the flag `verifyMaxFrameSizeEnabled `to just `maxFrameSizeEnabled` . Your comment for the PR also has the wrong name.
8. Lastly I don't think this should be added to 5.16.4 due to it being a change in behavior and a feature. Point releases should really just be about bug fixes and this definitely changes the expected behavior a bit. I think we should only put this into 5.17.0. The flag applies to client/server so we can't put it into 5.16.4 and by default turn it off easily so that's why I think just 5.17.0

activemq-client/src/main/java/org/apache/activemq/util/JMSExceptionSupport.java
```diff
@@ -61,6 +63,9 @@ public static JMSException create(Exception cause) {
         if (cause instanceof JMSException) {
             return (JMSException)cause;
         }
+        if (MaxFrameSizeException.class.isAssignableFrom(cause.getClass())) {
```
cshannon, line 66: I would use instanceof here instead. isAssignableFrom is a dynamic call vs compile time so the performance is slightly worse.

activemq-client/src/main/java/org/apache/activemq/MaxFrameSizeException.java
```diff
@@ -0,0 +1,31 @@
...
+import java.io.IOException;
+/**
+ * An exception thrown when max frame size is exceeded.
+ *
+ * 
+ */
+public class MaxFrameSizeException extends IOException {
+    private static final long serialVersionUID = -7681404582227153308L;
+
+    public MaxFrameSizeException(String message) {
```
cshannon, line 28: I would make this a little more descriptive. Maybe MaxFrameSizeExceededException instead.

mattrpav: done.

activemq-client/src/main/java/org/apache/activemq/util/JMSExceptionSupport.java
mattrpav: done.

activemq-client/src/main/java/org/apache/activemq/util/JMSExceptionSupport.java
```diff
@@ -61,6 +63,9 @@ public static JMSException create(Exception cause) {
         if (cause instanceof JMSException) {
             return (JMSException)cause;
         }
+        if (cause instanceof MaxFrameSizeExceededException) {
+            return new JMSException(cause.getMessage(), "41300");
```
cshannon, line 67: I'd include the max frame size exception as the root cause so the root cause exists and is a type of IOException whether server or client

activemq-client/src/main/java/org/apache/activemq/transport/nio/NIOTransport.java
```diff
@@ -139,8 +139,10 @@ protected void serviceRead() {
                     nextFrameSize = inputBuffer.getInt() + 4;
 
                     if (wireFormat instanceof OpenWireFormat) {
-                        long maxFrameSize = ((OpenWireFormat)wireFormat).getMaxFrameSize();
-                        if (nextFrameSize > maxFrameSize) {
+                        OpenWireFormat openWireFormat = (OpenWireFormat)wireFormat;
+                        long maxFrameSize = openWireFormat.getMaxFrameSize();
+
+                        if (openWireFormat.isMaxFrameSizeEnabled() && nextFrameSize > maxFrameSize) {
                             throw new IOException("Frame size of " + (nextFrameSize / (1024 * 1024)) + " MB larger than max allowed " + (maxFrameSize / (1024 * 1024)) + " MB");
```
cshannon, line 146: I'd use the createFrameSizeException() here

mattrpav: Let's do that part in 5.17.x since it could (low change) break server-side log parsers

ref: https://issues.apache.org/jira/browse/AMQ-8476

activemq-client/src/main/java/org/apache/activemq/util/JMSExceptionSupport.java
mattrpav: done

cshannon:
@mattrpav - In progress changes looking good so far, thanks!

activemq-client/src/main/java/org/apache/activemq/util/JMSExceptionSupport.java
```diff
@@ -61,6 +63,11 @@ public static JMSException create(Exception cause) {
         if (cause instanceof JMSException) {
             return (JMSException)cause;
         }
+        if (cause instanceof MaxFrameSizeExceededException) {
+            JMSException jmsException = new JMSException(cause.getMessage(), "41300");
+            jmsException.setLinkedException(cause);
```
cshannon, line 68: Instead of using linked I would use initCause() as thats what the server side does.  A server side error thrown will have a TransportDisposedIOExceptionset when calling e.getCause() so it would be consistent

cshannon:
@mattrpav - I reworked your test to add in the other transports and cases (testing out a mixture of client/server enabled and disabled). I pushed it here to a branch https://github.com/cshannon/activemq/tree/AMQ-8142-updated

It now generates a total of 48 tests , see what you think

activemq-client/src/main/java/org/apache/activemq/util/JMSExceptionSupport.java
mattrpav: Should we do both? A number of the JMSExceptionSupport hanlding is linking both in.  only costs a pointer

cshannon: My commit does both now

cshannon:
I have a negotiation test that I will push separate so I will merge this now.

PR #760, opened 2022-02-08
Some issues from static code analysis

winne42:
Hi guys,

I noticed some issues on lgtm.com and tried to fix them. Hope I didn't break anything.
Main changes:

- use try-with-resources, avoid leaking streams
- use for-each
- use diamond operator
- replace anonymous class with lambda/method reference in some instances
- catch Exception instead of Throwable
- declare methods synchronized that inherit from synchronized methods
- remove redundant null checks (OK, we can discuss this)
- avoid some NPEs
- remove unnecessary boxing/unboxing
- remove unnecessary toString()
- make private fields final if possible
- fix invalid JavaDoc
- remove unnecessary modifiers (e.g., private final, public on interface methods)
- make constructors in abstract classes protected, not public
- use recommended order of modifiers
- use Collections.emptyList() instead of Collections.EMPTY_LIST
- simplify some boolean expressions
- combine nested ifs
- use built-in string formatting of loggers instead of string concatenation
- use Exception constructor with cause parameter instead of initCause()
- use Java[] array syntax instead of C style
- use StandardCharsets.UTF8 instead of "UTF8"
- use .toArray() with empty instead of pre-sized array (recommended since late Java 6)

Enjoy!

### DIFF
--- a/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpNioSslTransport.java
+++ b/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpNioSslTransport.java
@@ -77,7 +77,7 @@ public class AmqpNioSslTransport extends NIOSSLTransport {
     }
 
     @Override
-    protected int secureRead(ByteBuffer plain) throws Exception {
+    protected synchronized int secureRead(ByteBuffer plain) throws Exception {
         if (initBuffer != null) {
             initBuffer.buffer.flip();
             if (initBuffer.buffer.hasRemaining()) {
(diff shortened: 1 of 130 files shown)


mattrpav:
This PR is not commit-able. Static code analysis tools work well for end-developer, but don't handle the context of infrastructure software such as messaging brokers and databases.

1. Too many changes in on PR
2. Missing context-- catching 'Throwable' is by design in many places
3. This PR adds 'boxed' primitives back in
4. other stuff.

winne42:
@mattrpav OK, you are right and thanks for the context. I will split this in several PRs and come back to you.

PR #764, opened 2022-02-09
use more try-with-resources (some resources leaked); boyscouting

winne42:


winne42:
@jbonofre what do you think, could we merge this?

mattrpav (commented):
This PR mixes type-o's, code style and try-with-resources. Given the serious classes it touches, I think it might be good to break it up into the minor type-o, style fixes and the try-with-resources changes as separate PRs. Specifically, I think we *really* want the unit tests to pass with these changes.

jbonofre:
Auto closed

PR #766, opened 2022-02-11
[AMQ-8472] Fix reload4j scope in activemq-partition module

jbonofre:


jbonofre:
@cshannon here's a proposed fix

cshannon:
Looking at my build the slf4j stuff is the main culprit of the error:

`Caused by: java.lang.IllegalStateException: Detected both log4j-over-slf4j.jar AND bound slf4j-reload4j.jar on the class path`

gemmellr:
Yes I was only talking about reload4j itself, in relation to log4j, that it wasnt a change in behaviour to pass reload4j on as log4j was already being passed on via the 'linkedin'/zookeeper bits before.

As the new JIRA didnt mention it, I didnt realise the point actually centred more around the slf4j bit. I'd agree that as the slf4j-log4j12 equivalent wasnt being passed on before the reload4j equivalent shouldnt be now.

activemq-partition/pom.xml
```diff
@@ -92,6 +84,16 @@
       <type>test-jar</type>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>ch.qos.reload4j</groupId>
+      <artifactId>reload4j</artifactId>
+      <scope>test</scope>
```
gemmellr, line 90: Personally I would be fine with test scope if this still works when used (by dependants, not within the build tests), but just to note: this is effectively different than what it did in 5.16.3 before (and different than you said you were going to do on the JIRA) since the module wont pass on reload4j like this anymore, wheras log4j was being passed on before due to the zookeeper related bits that brought it in before (but are now excluded via dependencyManagement).

cshannon:
> Yes I was only talking about reload4j itself, in relation to log4j, that it wasnt a change in behaviour to pass reload4j on as log4j was already being passed on via the 'linkedin'/zookeeper bits before.
> 
> As the new JIRA didnt mention it, I didnt realise the point actually centred more around the slf4j bit. I'd agree that as the slf4j-log4j12 equivalent wasnt being passed on before the slf4j-reload4j equivalent shouldnt be now.

Ok so maybe we just drop the slf4j part and keep reload4j and let people exclude if they want.

jbonofre:
@cshannon yes, it makes sense to me. Let me update the PR.

jbonofre:
@cshannon @gemmellr I've updated the PR to:
1. remove reload4j-slf4j dependency
2. still use reload4j-core in compile scope (allowing people to exclude it if needed)
Thoughts ?

cshannon:
@jbonofre - i think this is fine so i think you can go ahead with this and re-cut the release

gemmellr:
Seems like it would have been simplest to put it at test scope the way it was before, but if the tests pass then I guess its no biggie, just means anyone investigating them possibly needing to add it. Though I guess theres probably little chance of anyone bothering in this case, so...meh :)

gemmellr:
(The PR title and commit message are wrong though, since the scope of reload4j wasnt changed at all)

PR #767, opened 2022-02-15
Adding SECURITY.md for ActiveMQ

coheigea:
It's good security practice to add a SECURITY.md to a github repo, containing details of supported versions, where to find CVEs and how to report new issues.

Feel free to modify as required.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,14 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.16.x   | :white_check_mark: |
+| 5.15.x   | :white_check_mark: |
+| < 5.15.x | :x:                |
+
+## Reporting a Vulnerability
+
+For information on how to report a new security problem please see [here](https://www.apache.org/security/).
+Our existing security advisories are published [here](https://activemq.apache.org/security-advisories).


jbonofre (approved):

SECURITY.md
```diff
@@ -0,0 +1,14 @@
+# Security Policy
+
+## Supported Versions
+
+| Version | Supported          |
+| ------- | ------------------ |
+| 5.16.x   | :white_check_mark: |
+| 5.15.x   | :white_check_mark: |
+| < 5.15.x | :x:                |
```
gemmellr, line 9: This seems to be in disagreement with what has been discussed and agreed long-ago, and so feels potentially misleading at this point. Certainly worthy of discussion before merging a 'policy' PR after 4 minutes at a time most folks arent around and likely just before announcing a new release that seems to disagree with it.

5.15.15 was discussed, agreed, and finally announced to be the last 5.15.x release over 9 months ago. 5.16.3 occurred 3 months later and contained various dep updates with security fixes, 5.16.4 is about to occur over 9 months later doing the same again, with no equivalent 5.15.x update release being made or ever mentioned in a discussion.

jbonofre: @gemmellr I will change 5.15.x. I just missed it. Nothing changed to what have been discussed on the mailing list already.

gemmellr: Ok great. I think it is also past time it was removed from the website too, so perhaps while adding 5.16.4?

jbonofre: Good idea, I will do that on website as part of the 5.16.4 update.

PR #771, opened 2022-02-16
[AMQ-5137] make networkConnector decreaseNetworkConsumerPriority="true" the default

mattrpav:
- [x] Create PR for code change
- [ ] Update website network of brokers page to document the change


cshannon:
This change itself seems fine but I'm not sure if any existing network bridge tests will break. So you need to check all the network tests and run them to see if anything doesn't work now that the default changed.

Also, yes the change is trivial but again, all changes should have tests. You don't need to write anything crazy but you could at least write a test to verify the new default is true (could just check the network connector/bridge or even use the JMX NetworkConnectorView after start up) to prevent getting broken in the future.

mattrpav:
@cshannon yep. Plan to. I kicked in these PR’s to get the tests running in parallel to see fallout. I’ll be going through the unit test failures and adding the default value tests next

mattrpav:
Note-- first test run failed on a StompTransport test (appears unrelated). Kicking off fresh run of tests

cshannon (approved):

PR #772, opened 2022-02-16
[AMQ-6883] Update windows wrapper confs to be consistent with windows…

mattrpav:
… bin/*.bat file for memory and dedicatedTaskRunner


jbonofre (approved):
It makes sense to me.

PR #778, opened 2022-02-22
Fix TempUsage memory leak

artpdr:
If you have an application that creates lots of queues it will eventually
fail with OOM because TempUsage is started on Queue#start but never stopped.

The `systemUsage.getTempUsage().start()` used on the Queue#start
adds elements on a List from TempUsage parent and these elements
are never removed.

To reproduce this issue you need to leave an application
running for a long time creating different queues.

The only way to avoid the leak right now is to stop the BrokerService,
which isn't a solution.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
@@ -1058,12 +1058,19 @@ public class Queue extends BaseDestination implements Task, UsageListener, Index
             pagedInMessages.clear();
 
             systemUsage.getMemoryUsage().removeUsageListener(this);
+
             if (memoryUsage != null) {
                 memoryUsage.stop();
             }
+
             if (systemUsage.getStoreUsage() != null) {
                 systemUsage.getStoreUsage().stop();
             }
+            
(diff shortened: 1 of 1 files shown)


artpdr:
Guys can you take a look at this. This issue requires a server restart from time to time to avoid an application OOM.

AndreRog (approved):

cshannon:
Thanks for the PR, this looks like a good catch. A couple things: Pull requests should have a Jira associated with it describing the issue. https://issues.apache.org/jira/projects/AMQ . Also, ideally it would be good to include a unit test that can demonstrate the issue and show that it's fixed.

In this case I'll create the Jira as I am going to re-work the fix slightly and I will a test real quick because we are about to do a new release but for the future you can keep that in mind.

cshannon:
@artpdr - I amended/merged your commit and updated some tests. https://issues.apache.org/jira/browse/AMQ-8509

https://github.com/apache/activemq/commit/d1e9ba68b3cbea3e9294cf4993d005c3c8ea86f2

Thanks again for catching that, this will go into 5.17.0 and 5.16.5.

artpdr:
Thanks a lot @cshannon. Good point on the testing part... I didn't include the test because I didn't find any test for that class.

PR #782, opened 2022-02-24
[AMQ-8513] Upgrade to ant 1.10.12

jbonofre:


gemmellr:
The latest Ant is 1.10.12, this seems silly.

jbonofre:
1.6.5 includes latest CVE. But you are right, let me try ant 1.10.12, it should work out of the box.

gemmellr:
1.8.1 is used elsewhere in the build. Probably makes sense to manage the version.

gemmellr:
There are loads of CVE fixes in the later streams if thats why you are doing it: https://ant.apache.org/security.html

jbonofre:
@gemmellr good point, let me update Jira and this PR.

gemmellr:
Head before the change, note that the older one this changed is also in a different groupId which should also be updated:
```
$ mvn dependency:tree | grep "ant:ant"
[INFO] \- ant:ant:jar:1.6.2:compile
[INFO] |     \- org.apache.ant:ant:jar:1.8.1:provided
[INFO] |        \- org.apache.ant:ant-launcher:jar:1.8.1:provided
[INFO] |     \- org.apache.ant:ant:jar:1.8.1:compile
[INFO] |        \- org.apache.ant:ant-launcher:jar:1.8.1:compile
[INFO] |  |     \- org.apache.ant:ant:jar:1.8.1:compile
[INFO] |  |        \- org.apache.ant:ant-launcher:jar:1.8.1:compile
[INFO] |     \- org.apache.ant:ant:jar:1.8.1:compile
[INFO] |        \- org.apache.ant:ant-launcher:jar:1.8.1:compile
```

activemq-openwire-generator/pom.xml
```diff
@@ -44,9 +44,9 @@
       <artifactId>annogen</artifactId>
     </dependency>
     <dependency>
-      <groupId>ant</groupId>
+      <groupId>org.apache.ant</groupId>
       <artifactId>ant</artifactId>
-      <version>1.6.2</version>
+      <version>1.10.12</version>
```
gemmellr, line 49: There are various other occurrences of Ant in the build per the previous discussion, which this change doesnt affect. Use dependencyManagement to align the dep versions from one place at the root.

jbonofre:
Yes thanks I will do that.

activemq-openwire-generator/pom.xml
jbonofre: I don't see other direct other reference (outside of openwire generator). The only ant dependency I can see with `mvn dependency:tree` are transitive dependency. So I think it's OK like this.

gemmellr: Yes they are transitive. No that doesnt make it OK. If the point is to upgrade to get CVEs fixes for example then they all need to be upgraded to actually achieve that. You wouldnt leave an old log4j 2.13 dep in place for example just because its transitive. The build should generally align on 1 version unless there is good reason not to.

jbonofre: Hmm, ok. I just hope that the upgrade will work for the transitive. I'm trying and I will update the PR accordingly. Thanks.

gemmellr:
Most of the new exclusions dont seem to make sense. You excluded "_ant:ant_" via both dependencyManagement *and* the actual dependency declarations for "_org.linkedin.zookeeper-impl"_, whereas from the previous dependency:tree output that already appears to be using the newer "org.apache.ant:ant" bits already, seemingly making all of those exclusions completely superfluous. The only one that seems like it should have any effect is the generator one.

jbonofre:
@gemmellr I cleaned up the exclusion. Can you please take a new look ?

pom.xml
```diff
@@ -639,6 +646,10 @@
         <artifactId>org.linkedin.zookeeper-impl</artifactId>
         <version>${linkedin-zookeeper-version}</version>
         <exclusions>
+          <exclusion>
+            <groupId>ant</groupId>
+            <artifactId>ant</artifactId>
+          </exclusion>
```
gemmellr, line 652: As prior comment, this exclusion doesnt seem like it would be doing anything, as org.linkedin.zookeeper-impl wasnt showing a dependency on "ant:ant" previously, only "org.apache.ant:ant". Seems like it can be removed.

activemq-spring/pom.xml
```diff
@@ -87,6 +87,11 @@
       <artifactId>org.linkedin.zookeeper-impl</artifactId>
       <scope>provided</scope>
     </dependency>
+    <dependency>
+      <groupId>org.apache.ant</groupId>
+      <artifactId>ant</artifactId>
+      <scope>provided</scope>
+    </dependency>
```
gemmellr, line 94: Is this actually needed? It was only a transitive dependency before, sio unless the module has its own Ant usage it seems like it should still be transitive. 

The dependencyManagement will still control the version without adding this, and the scope be based on the scope of whatever pulls it in transitively.

activemq-partition/pom.xml
```diff
@@ -49,6 +49,10 @@
       <groupId>org.linkedin</groupId>
       <artifactId>org.linkedin.zookeeper-impl</artifactId>
     </dependency>
+    <dependency>
+      <groupId>org.apache.ant</groupId>
+      <artifactId>ant</artifactId>
+    </dependency>
```
gemmellr, line 55: As other comment, is this actually needed?

activemq-spring/pom.xml
jbonofre: It was what my mean in my previous comment, but I think I was not clear.

gemmellr:
LGTM now.

PR #783, opened 2022-02-24
[AMQ-8514] Upgrade to groovy 4.0.1

jbonofre:


jbonofre:
The purpose is just to include CVE fixed minot version. I can jump to "bigger" version.

gemmellr:
That would have made perfect sense at the time (these also look around 17 years old), but entirely weird now given all the issues it would still contain that were fixed in later releases. 04 isnt even the latest version of that stream, it seems 06 is, though that one doesnt look to have a pom. Seems like the 1.x stream proper also came later, nevermind the other minors and majors after that.

cshannon:
I would just wait on stuff like this (ant, groovy,etc) and freeze the release. We can target upgrades on these dependencies to the latest versions (like groovy 3 or 4) for 2.17.1 or 2.18.

gemmellr:
Yep it certainly does seem like it could wait until after 5.17.0, having done for so very long already.

jbonofre:
Oh yeah it's not necessary for 5.17.0. It's just that I found these dependencies very old ;)

PR #784, opened 2022-02-25
Use Jetty atomic jar files instead of jetty-all (deprecated) and update to Jetty 9.4.45.v20220203

jbonofre:


activemq-amqp/pom.xml
```diff
@@ -77,10 +77,13 @@
...
     </dependency>
     <dependency>
-      <!-- using it for Jetty's JNDI context to work /w Joram tests. -->
-      <groupId>org.eclipse.jetty.aggregate</groupId>
-      <artifactId>jetty-all</artifactId>
-      <classifier>uber</classifier>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-server</artifactId>
+      <scope>test</scope>
+    </dependency>
```
gemmellr, line 83: Is this actually needed? If the comment is accurate (may not be), seems as likely only the lower jetty-jndi dep is actually needed.

pom.xml
```diff
@@ -963,28 +963,87 @@
...
+        <version>${jetty-version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-util</artifactId>
+        <version>${jetty-version}</version>
+      </dependency>
+      <dependency>
+        <groupId>org.eclipse.jetty</groupId>
+        <artifactId>jetty-http</artifactId>
```
gemmellr, line 998: Using jetty-bom might be simpler than introducing all of these (and the others).

gemmellr (commented):
It would probably have been simpler separate the version bump, do that for 5.17.0 and proceed with the release already, and then work on the dep changes for a followup, but if this works and folks think its ok it would be good to change as using the -all is ugly (just raised a JIRA against Artemis for the same thing recently after spotting it in some earlier changes)

activemq-amqp/pom.xml
gemmellr: The answer is its not strictly needed, as its a transitive dependency of jettys websocket-server dep that already is used in the module, but is used and so isnt really doing any harm adding it.

jbonofre: Let me give you some background:
- jetty 9.4.45 contains new CVE fixes
- just updating to jetty 9.4.45 using jetty-all uber jar won't work as jetty broke the uber jar
So, that's why I did both changes in the same PR.

pom.xml
jbonofre: Good point about the bom, I can update the PR to use it.

activemq-amqp/pom.xml
gemmellr: I wasnt aware since it wasnt said anywhere that the uber jar was broken. Makes sense then.

PR #785, opened 2022-02-25
[AMQ-8515] FailoverTransport should handle MaxFrameSizeExceededException

mattrpav:


mattrpav:
@cshannon please give this a look over, it'd be good to get it into 5.17.0. This fix addresses the use case that originally sparked the improvement-- failover with a message that server-side maxFrameSize check failing results in infinite retries.

cshannon:
@mattrpav - Nice catch, this is definitely an issue with the exception type and this seems like the right fix. Thanks for updating the tests, they look good. We definitely don't want to be attempting to reconnect on this failure as it would never work so it needs to be propagated back up.

cshannon (approved):

cshannon:
As long as the CI tests pass I think this is good to merge

mattrpav:
First round of tests passed clean. I pushed an update of change that included comments in the test w/ link to the jira about how client check=false + failover + server check=true is an infinite retry.

I'll merge once the next round of tests goes green.

Thanks!

PR #787, opened 2022-03-01
AMQ-8520: Re-enable all modules to compile and build by default

cshannon:
Updates the modules part of the previous full.test profile to always
build but skip running tests by default.

Also fixed miss dependency updates including log4j and jetty

### DIFF
--- a/activemq-itests-spring31/pom.xml
+++ b/activemq-itests-spring31/pom.xml
@@ -83,4 +83,25 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+
+  <profiles>
+    <profile>
+      <!-- Default - no tests -->
+      <id>default</id>
+      <activation>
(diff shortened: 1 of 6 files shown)


cshannon:
This is what I have so far but is still broken as there is a lot of Log4j specific stuff that won't compile so need to decide how to fix the tests

cshannon:
There's a ton of places where log4j is used directly and is breaking everything. @mattrpav is taking a look now with a follow on PR to fix it. We can merge this plus the follow on PR and sort out the tree when ready.

cshannon:
We should also update CI to run this project as well. For PRs we could activate the smoke tests or quick tests only etc. We could also set up a daily job and run a couple other profiles such as the ctivemq.exclude-broken-tests which should run a lot more tests but take longer.

jbonofre:
Testing the PR now.

PR #791, opened 2022-03-04
add the capability to override memory options

ilyesAj:
since the last release , the xmx and xms are hardcoded on `env` , the goal is to use ACTIVEMQ_OPTS_MEMORY environement variable if needed to override memory options

### DIFF
--- a/assembly/src/release/bin/env
+++ b/assembly/src/release/bin/env
@@ -30,8 +30,9 @@
 # ACTIVEMQ_TMP="$ACTIVEMQ_BASE/tmp"
 
 # Set jvm memory configuration (minimal/maximum amount of memory)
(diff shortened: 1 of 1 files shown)


cshannon:
Thanks for the PR, this seems reasonable to me, would you be able to create a Jira for this and include the Jira number in the commit to track the change? https://issues.apache.org/jira/projects/AMQ

cshannon:
Looks like this was changed here: https://github.com/apache/activemq/commit/a767bda603119387af8e8ce06af268470fe33c3a

The Xms change made sense (going lower) but obviously dropping that If statement is an issue. So been broken for a while but we should change it back to allow overriding.

jbonofre (approved):

PR #792, opened 2022-03-04
NO-JIRA: Activate quick-tests by default

cshannon:


cshannon:
@mattrpav - We can let this run and see how long it takes

mattrpav (approved):
I think we should merge this so we can split up the next round of fixes for exclusions and tests to fix

cshannon:
Sounds good, you can go ahead and merge it if you want

PR #795, opened 2022-03-08
AMQ-8528: Fix test failures in integration module

cshannon:
**Broken Tests Fixed by this PR:**
- [x] org.apache.activemq.network.DurableFiveBrokerNetworkBridgeTest
- [x] org.apache.activemq.spring.SpringTest
- [x] org.apache.activemq.broker.BrokerRedeliveryTest
- [x] org.apache.activemq.bugs.AMQ3537Test
- [x] org.apache.activemq.bugs.AMQ4952Test
- [x] org.apache.activemq.usecases.OfflineDurableSubscriberTimeoutTest
- [x] org.apache.activemq.store.jdbc.JDBCConcurrentDLQTest

**Broken Tests already fixed in main: (JMS 2.0, only applies to 5.18.x so don't need to backport)**
- [x] org.apache.activemq.bugs.AMQ6122Test (this works fine in 5.17.x)
- [x] org.apache.activemq.bugs.AMQ5212Test

**Broken Tests already fixed in another PR:**
- [x] org.apache.activemq.broker.virtual.MessageDestinationVirtualTopicTest - Fixed here: https://github.com/apache/activemq/pull/796 - This can be backported to 5.17.1

The issue is Spring 5.3 seems to require an updated annotation spec. geronimo-annotation_1.0_spec is too old. However, it's broken due to @Resource annotation usage which we only used in a couple unit tests. The jar is marked as optional and it's never used anywhere in the real codebase so the broker will run out of the box just fine. I think we can just fix/update this in 5.17.1 and 5.18.0 because if someone is using that annotation in their own code and Spring 5.3 they can just make sure to include the right annotation jar (spring already pulls it in) so they really just need to add an exclusion on the gernomio one so not a huge deal as it's marked as optional anyways. 

**Flaky or non broken Tests: (These tests intermittently fail, or binding errors, etc. Can fix later.)**
- org.apache.activemq.usecases.StartAndConcurrentStopBrokerTest - fails intermittently for me
- org.apache.activemq.proxy.ProxyConnectorTest - due to Address already in use bind error
- org.apache.activemq.store.jdbc.JDBCIOExceptionHandlerTest - passed for me

These last 2 are already part of the exclude broken test profile so should also be excluded from quick tests and other profiles
- org.apache.activemq.network.SSHTunnelNetworkReconnectTest - works for me as long as ssh is properly configured, so probably just an environment thing in CI so should exclude
- org.apache.activemq.network.NetworkFailoverTest.testRequestReply - Has been broken for a long time

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/broker/BrokerRedeliveryTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/broker/BrokerRedeliveryTest.java
@@ -146,6 +146,7 @@ public class BrokerRedeliveryTest extends org.apache.activemq.TestSupport {
 
         PolicyEntry policyEntry = new PolicyEntry();
         policyEntry.setUseCache(false); // disable the cache such that duplicates are not suppressed on send
+        policyEntry.setSendDuplicateFromStoreToDLQ(true);
 
         PolicyMap policyMap = new PolicyMap();
         policyMap.setDefaultEntry(policyEntry);
--- a/activemq-unit-tests/pom.xml
+++ b/activemq-unit-tests/pom.xml
@@ -1074,6 +1074,11 @@
                 <exclude>org/apache/activemq/store/kahadb/disk/util/DataByteArrayInputStreamTest.*</exclude>
                 <exclude>org/apache/activemq/bugs/AMQ5266StarvedConsumerTest.*</exclude>
                 <exclude>org/apache/activemq/bugs/AMQ5266Test.*</exclude>
+
+                <!--Broken tests -->
+                <exclude>**/NetworkFailoverTest.*/**</exclude>
+                <!-- This test only works on machines which have ssh properly configured -->
+                <exclude>**/SSHTunnelNetworkReconnectTest.*</exclude>
               </excludes>
             </configuration>
           </plugin>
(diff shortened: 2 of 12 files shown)


cshannon:
@jbonofre  and @mattrpav - This is a work in progress, I'm going to make more changes and push them up. I'll let you guys know when I'm done.

cshannon:
@jbonofre and @mattrpav - Ok I'm all finished, we are good to go to keep going with 5.17.0. There's nothing here that justifies cancelling the vote. There are a couple issues to fix later (notably the JMS 2.0 broken tests in main) and some other flaky tests but nothing that's a real regression.

cshannon:
I figure we can merge this into main after CI runs and then backport to 5.17.x for 5.17.1 plus continue on with some of my comments to fix the flaky tests, jms 2.0 issues and genonimo annotation jar.

activemq-unit-tests/src/test/java/org/apache/activemq/broker/BrokerRedeliveryTest.java
```diff
@@ -146,6 +146,7 @@ public void testNoScheduledRedeliveryOfDuplicates() throws Exception {
 
         PolicyEntry policyEntry = new PolicyEntry();
         policyEntry.setUseCache(false); // disable the cache such that duplicates are not suppressed on send
+        policyEntry.setSendDuplicateFromStoreToDLQ(true);
```
mattrpav, line 149: Nice catch. This is a good example of how dupe from store can occur as well.

mattrpav (approved):
LGTM

cshannon:
I'm going to go ahead and merge this into main and into 5.17.x as I think it should be fine

PR #805, opened 2022-03-31
[AMQ-8556] Upgrade Jackson 2.13.2/2.13.2.2

mattrpav:


gemmellr:
databind needs its own version to get newer version

mattrpav:
> databind needs its own version to get newer version

yeah, saw that they got updated separately. I'll break out the dep management as those deps tend to drift for patches.

PR #809, opened 2022-03-31
[AMQ-8561] Add missing jetty-webapp dependency in activemq-web

jbonofre:


mattrpav:
@jbonofre can you squash + merge?

PR #812, opened 2022-04-05
AMQ-8550 - Check for null keystore/truststore passwords

cshannon:
Inside ActiveMQSslConnectionFactory the passwords should be checked for
null so a NPE isn't thrown. Null will be passed to the factories instead
and the keystore/truststore factories will try and load the keystores
using null for the password which may or may not work depending on the
implementation and if password is set.


mattrpav (approved):
LGTM. Good pick up to add the same check for keystore

PR #827, opened 2022-04-17
Fix unit tests which fail related to log4j 2.17.2 update

jbonofre:


pom.xml
```diff
@@ -224,7 +224,7 @@
     <module>activemq-run</module>
     <module>activemq-shiro</module>
     <module>activemq-spring</module>
-    <module>activemq-runtime-config</module>
+    <!-- <module>activemq-runtime-config</module> -->
```
gemmellr, line 227: Presumably a build break waiting to happen, only still working now as it can download a [always-stale] snapshot, but once you try to release it wont be able to. Essentially the same as the issues that cropped up during the 5.17.0 release.

(That or its entirely unused and should be properly removed rather than just commented out)

activemq-unit-tests/src/test/java/org/apache/activemq/store/SharedFileLockerLoggingTest.java
```diff
@@ -0,0 +1,171 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.slf4j.LoggerFactory;
+
+
+import static org.junit.Assert.assertTrue;
+
+public class SharedFileLockerLoggingTest {
+    private static final org.slf4j.Logger LOG = LoggerFactory.getLogger(SharedFileLockerLoopTest.class);
```
gemmellr, line 51: Log name seems wrong given the new class.

jbonofre:
Runtime config works. It was just a local change. I will fix that and the log names.

PR #839, opened 2022-05-10
Add FirstMessageAge and FirstMessageTimestamp JMX metrics for queues.

NikitaShupletsov:
This change is meant to match the piece of functionality from Artemis: https://github.com/apache/activemq-artemis/commit/00837c120d131cd6db706aed5f549db06cd7b1e7

Add two new JMX attributes for queues: FirstMessageAge and FirstMessageTimestamp.
First one shows the diff between current time and the message timestamp of the first message in the queue.
Second one shows the message timestamp of the first message in the queue.


mattrpav:
ref: https://issues.apache.org/jira/browse/AMQ-8463

I'd like to review the naming here, given the plans to add additional metrics. We should have it all align.

Metrics:
```
nextMessageEnqueuedTimestamp (from the message’s brokerInTime)
nextMessageID

(‘last’ aka the most recent message to be ack’d — processed in-flight)
lastEnqueuedMessageTimestamp (from brokerInTime on the message)
lastDequeuedTimestamp (clock timestamp of when the message was dequeued)
lastDequeuedMessageTimestamp (from brokerInTime on the message)
lastDequeuedMessageID

(‘First’ aka first all time in the life of the destination since boot — processed in-flight)
firstEnqueuedMessageTimestamp (from brokerInTime)
firstDequeuedTimestamp (wall clock time when the message was dequeued)
firstDequeuedMessageTimestamp (from brokerInTime)
```

NikitaShupletsov:
Initially the idea was to call it oldestMessageAge or ageOfOldestMessage, but then I came across the artemis metric and decided to use the same name for the sake of consistency. 
So, what do you think? does `oldest` instead of `first` sound better?

jbonofre:
The change looks good to me. I think using `firstMessageTimestamp` is good. However, I would also add `lastMessageTimestamp`.

I will resume my work on this one (spinning another one).

mattrpav:
I think this approach is incorrect, and we should revisit the approach in 6.1.0

jbonofre:
We will see later but I will probably close this PR with a new one for 6.1.0 with a different impl.

PR #840, opened 2022-05-11
[AMQ-8601] Refine the logging messages

zisding:
Hello,

While viewing the [MAPREDUCE-4262](https://issues.apache.org/jira/browse/MAPREDUCE-4262), I found that the logging statements might give inaccurate messages. 

I also found that the in the line 93 of the file [UpdateVirtualDestinationsTask](https://github.com/apache/activemq/blob/b3c2c49f96bebee14b74c6456feb233e0312724b/activemq-runtime-config/src/main/java/org/apache/activemq/plugin/UpdateVirtualDestinationsTask.java#L93), the log messages says "Removing virtual destination  ". However, the removing action should already completed in previous code (line 92).
```
plugin.virtualDestinationRemoved(connectionContext, removedVirtualDest);                            
LOG.info("Removing virtual destination: {}", removedVirtualDest); 
```
Would it be better if we change the verb "Removing" to "Removed" to indicate the action is completed, which is similar to the previous [logging statement?](https://github.com/apache/activemq/blob/b3c2c49f96bebee14b74c6456feb233e0312724b/activemq-runtime-config/src/main/java/org/apache/activemq/plugin/UpdateVirtualDestinationsTask.java#L82)
```
virtualDestinationInterceptor                        
 .setVirtualDestinations(getVirtualDestinations());                
plugin.info("applied updates to: "                        
+ virtualDestinationInterceptor); 
```
Or can we move the logging statement to the line before 92? Since when there was an exception in previous lines, the logging message would not be printed, which may be not good for debugging. 

A similar issue is also found in the file [StatisticsBrokerPlugin](https://github.com/apache/activemq/blob/9b1eb96b838957cd60541ca5e057567be3f11990/activemq-broker/src/main/java/org/apache/activemq/plugin/StatisticsBrokerPlugin.java#L47) where it seems that the logging statement it is describing the whole method. Would it be better to move this logging statement to the beginning of the method? 

Especially considering the following logging practices:

- https://github.com/apache/activemq/blob/9b1eb96b838957cd60541ca5e057567be3f11990/activemq-broker/src/main/java/org/apache/activemq/plugin/DiscardingDLQBrokerPlugin.java#L56
- https://github.com/apache/activemq/blob/9b1eb96b838957cd60541ca5e057567be3f11990/activemq-runtime-config/src/main/java/org/apache/activemq/plugin/RuntimeConfigurationPlugin.java#L37 

Thanks.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/plugin/StatisticsBrokerPlugin.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/plugin/StatisticsBrokerPlugin.java
@@ -43,8 +43,8 @@ public class StatisticsBrokerPlugin implements BrokerPlugin {
      * @see org.apache.activemq.broker.BrokerPlugin#installPlugin(org.apache.activemq.broker.Broker)
      */
     public Broker installPlugin(Broker broker) throws Exception {
-        StatisticsBroker answer = new StatisticsBroker(broker);
         LOG.info("Installing StaticsBroker");
+        StatisticsBroker answer = new StatisticsBroker(broker);
         return answer;
     }
 }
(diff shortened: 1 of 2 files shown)


jbonofre (approved):
LGTM, that's really minor minor minor things ;)

PR #846, opened 2022-06-06
[AMQ=8625] Upgrade maven-bundle-plugin 5.1.6

mattrpav:


mattrpav:
@jbonofre I successfully tested this update with Karaf v4.3.7. Any reason to hold off merging due to bundle plugin or karaf changes?

jbonofre:
@mattrpav it should be fine (karaf version now exports java.* packages). Anyway, I have a bunch of other dependency updates to do, so I will base on that one.

PR #861, opened 2022-07-23
[AMQ-8971] activemq-client karaf feature supporting both JMS 1.1 and JMS 2.0

artnaseef:


artnaseef:
Note there is a test application at the following github repo that I used to reproduce the problem, and test the solution:

  https://github.com/artnaseef/AMQ8971

It only tests with an application using AMQ out-of-the-box (i.e. JMS 1.1).  JMS 2.0 (camel etc) should work, but isn't tested with that repo.  In other words, this update makes the `activemq-client` feature work for both JMS 1.1 and JMS 2.0 applications.

activemq-karaf/src/main/resources/features-core.xml
```diff
@@ -63,4 +63,12 @@
       <bundle dependency="true">mvn:com.fasterxml.jackson.core/jackson-databind/${jackson-databind-version}</bundle>
       <bundle dependency="true">mvn:com.fasterxml.jackson.core/jackson-annotations/${jackson-version}</bundle>
     </feature>
+
+    <!-- JMS1.1 and JMS2.0 support via the same Geronimo JMS 2 bundle  -->
+    <!-- Ideal since all JMS code shares the same class definitions    -->
+    <!-- Fortunately, JMS2.0 is fully backward compatible with JMS 1.1 -->
+    <feature name="activemq-combined-jms-spec">
```
jbonofre, line 70: I think we could have done this directly in the activemq-client feature, no need to separate.

activemq-karaf/src/main/resources/features-core.xml
```diff
@@ -63,4 +63,12 @@
...
+        <feature prerequisite="true">wrap</feature>
+        <bundle dependency="true">wrap:mvn:org.apache.geronimo.specs/geronimo-jms_2.0_spec/1.0-alpha-2$overwrite=merge&amp;Export-Package=javax.jms;version=1.1,javax.jms;version=2.0</bundle>
```
jbonofre, line 72: Instead of rewrapping, why not simply installing the two spec versions (1.1 and 2.0) ?

PR #870, opened 2022-07-31
[AMQ-9005] remove xalan dependency due to it being end of life

lucastetreault:
https://issues.apache.org/jira/browse/AMQ-9005

There's no use of Xalan in code and all tests are passing without the dependency. Seems safe to remove but I assume we'll need a note about it in the release notes.

### DIFF
--- a/activemq-broker/pom.xml
+++ b/activemq-broker/pom.xml
@@ -67,13 +67,6 @@
       <optional>true</optional>
     </dependency>
 
-    <!-- to support XPath based Selectors -->
-    <dependency>
-      <groupId>xalan</groupId>
-      <artifactId>xalan</artifactId>
-      <optional>true</optional>
-    </dependency>
-
     <!-- =============================== -->
     <!-- Testing Dependencies -->
     <!-- =============================== -->
--- a/activemq-kahadb-store/pom.xml
+++ b/activemq-kahadb-store/pom.xml
@@ -106,11 +106,6 @@
       <artifactId>spring-context</artifactId>
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>xalan</groupId>
-      <artifactId>xalan</artifactId>
-      <optional>true</optional>
-    </dependency>
 
     <!-- =============================== -->
     <!-- Testing Dependencies            -->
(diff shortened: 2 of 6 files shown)


jbonofre (approved):

mattrpav:
xalan reference missed:

https://github.com/apache/activemq/blob/main/assembly/src/release/examples/conf/resin-web.xml

pjfanning:
> xalan reference missed:
> 
> https://github.com/apache/activemq/blob/main/assembly/src/release/examples/conf/resin-web.xml

I'm guessing but that TransformerFactory setting can probably just be removed. That would likely mean that the JDK built-in transformer is used (unless another JAXP compliant transformer was on the classpath, eg Saxon).

mattrpav:
Other xalan references need to be cleaned up and XPath unit tests validated

./activemq-client/src/main/java/org/apache/activemq/filter/XPathExpression.java
./activemq-broker/src/main/java/org/apache/activemq/filter/XalanXPathEvaluator.java

jbonofre:
Good catch. I will do a new pass. Thanks !

lucastetreault:
> Other xalan references need to be cleaned up and XPath unit tests validated
> 
> ./activemq-client/src/main/java/org/apache/activemq/filter/XPathExpression.java ./activemq-broker/src/main/java/org/apache/activemq/filter/XalanXPathEvaluator.java

So I'm not entirely sure what to do with these...

activemq-broker/src/main/java/org/apache/activemq/filter/XalanXPathEvaluator is actually identical to activemq-broker/src/main/java/org/apache/activemq/filter/JAXPXPathEvaluator and neither of them use Xalan. I guess we can delete XalanXPathEvaluator and change the reference to it in XPathExpression to use JAXPXPathEvaluator. 

The SelectorParserTest is the only test I can find using XPathExpression and it passes with these changes. 

> > xalan reference missed:
> > https://github.com/apache/activemq/blob/main/assembly/src/release/examples/conf/resin-web.xml
> 
> I'm guessing but that TransformerFactory setting can probably just be removed. That would likely mean that the JDK built-in transformer is used (unless another JAXP compliant transformer was on the classpath, eg Saxon).

I don't know anything about resin. How can I test this?

lucastetreault:
I'll open another PR but I'll need input on how to test the resin-web.xml change :)

PR #906, opened 2022-09-07
AMQ-9074: javadoc output simplifications

gemmellr:
use -noindex to remove various .js files, disable html source listings for javadoc

https://issues.apache.org/jira/browse/AMQ-9074


gemmellr:
I prepared a site update with a backported version of this, see PR referenced above.

cshannon (approved):

PR #907, opened 2022-09-29
Add Jenkinsfile for s390x

srajmane:
Add Jenkinsfile for s390x

https://issues.apache.org/jira/browse/AMQ-8133

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,11 +21,7 @@
 
 pipeline {
 
-    agent {
-        node {
-            label 'ubuntu'
-        }
-    }
+    agent none
 
     tools {
         // ... tell Jenkins what java version, maven version or other tools are required ...
@@ -43,6 +39,7 @@ pipeline {
 
     stages {
         stage('Initialization') {
+            agent { label 'ubuntu && s390x' }
             steps {
                 echo 'Building branch ' + env.BRANCH_NAME
                 echo 'Using PATH ' + env.PATH
@@ -50,6 +47,7 @@ pipeline {
         }
 
         stage('Cleanup') {
+            agent { label 'ubuntu && s390x' }
             steps {
                 echo 'Cleaning up the workspace'
                 deleteDir()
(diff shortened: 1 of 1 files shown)


srajmane:
@mattrpav @jbonofre 
Could you please setup/create a job for s390x ? I have created jenkinsfile for s390x in this PR.

mattrpav:
@srajmane do you have any stats of build times on s390x? on the x86 stuff the build can run for several hours.

I'm thinking step 1 re s390x may be to have weekly schedule vs build on-every-PR

srajmane:
@mattrpav Build time is 6-6.5 Hrs on s390x. For now, it is okay to have weekly schedule.

srajmane:
@mattrpav Could you please have a look?  Thanks !!

jbonofre:
I guess it won't work on Apache Jenkins right ? Is it for a fork you have on your side ?

mattrpav:
@jbonofre the Jenkins error is on the buggy test that fails on other CI runs as well.

@srajmane is there a way to merge the change so we have one JenkinsFile? That way we would inherit pipeline changes without having to keep two files in sync (things like JDK LTS runs, mvn verify, etc)

srajmane:
@mattrpav Thank you for the update. Sure, I will check and will try to keep one Jenkinsfile.

jbonofre:
My point is that we don't have all executors available on Apache Jenkins.

mattrpav:
> My point is that we don't have all executors available on Apache Jenkins.

Apache Infra added a couple for s390

jbonofre:
> > My point is that we don't have all executors available on Apache Jenkins.
> 
> Apache Infra added a couple for s390

OK, last time I asked it was in discussion because it's hard to manage. Fair enough then ;)

srajmane:
@jbonofre @mattrpav Added s390x node in Jenkinsfile. Please have a look. 
Thanks !!

srajmane:
@jbonofre @mattrpav 
 Could you please have a look? Thanks !!

mattrpav:
@srajmane I opened #956 which uses parameters (manual execution) as a first step. Need to understand how long the build runs on Apache's s390x. If we get quick turn-around times, we can add it to all PRs moving forward.

PR #909, opened 2022-10-11
[AMQ-9111] Upgrade to Jackson 2.13.4

mattrpav:


jbonofre:
I think it would make more sense to directly upgrade to 2.14.0.rc1 as it fixes other issues. I actually already have the PR ready ;)

PR #912, opened 2022-10-11
[AMQ-9111] Upgrade to Jackson 2.13.4

mattrpav:


jbonofre:
Didn't you see my comment ? I think it's better to directly upgrade to 2.14.0.rc1.

gemmellr:
RC2 is already out, though the maintainer doesnt recommend using either in production and 2.14.0 isnt due for 3-4 weeks.

2.13.4.1 also seems likely to be coming with a backport of the newest issue.

jbonofre:
My point is that current 2.13.x version doesn't fix latest issues/security issues. So for me it doesn't make sense to upgrade, better to wait 2.14 or 2.13 including latest security fixes.

gemmellr:
Not sure it makes a big difference, fixes 1 of the 2 so its better than not doing anything, and can always do another upgrade when a suitable release is actually available. Seems as useful as updating to a candidate the maintainer pleads that you shouldn't ever use for a final release.

jbonofre:
@gemmellr it sounds reasonable. Agree.

gemmellr:
databind 2.13.4.2 was released with backport of the other issue fix.

PR #920, opened 2022-10-15
[AMQ-9107] Add a test for offline durable subscriptions

lucastetreault:
As per Matt's request [here](https://github.com/apache/activemq/pull/908#discussion_r994672165), adding a test to make sure offline durable subscribers can be added/removed as usual.


lucastetreault:
Pushed an empty commit to trigger the tests again. All the tests passed on my own Jenkins for this commit 🤷‍♂️

mattrpav (changes requested):
@lucastetreault Would you please update the test to check for does-not-exist then exists of the JMX beans? JMX bean registration can fail in its own code path.

lucastetreault:
@mattrpav, I added a test case to the MBeanTest. Hopefully more in line with what you were thinking.

jbonofre (approved):

jbonofre:
LGTM, thanks. I will manually merge to squash the two commits in one.

jbonofre (approved):

PR #927, opened 2022-11-01
AMQ-9153: Fix slow consumer advisory for queue subscriptions

cshannon:
Due to changes with Queues to check if consumers are full before adding more messages to the subscription, the Queue dispatch logic needed to be updated to mark subscriptions as slow and send advisories if configured instead of relying on the subscription itself to do it.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
@@ -2210,6 +2210,16 @@ public class Queue extends BaseDestination implements Task, UsageListener, Index
                         // no further dispatch of list to a full consumer to
                         // avoid out of order message receipt
                         fullConsumers.add(s);
+
+                        //For full consumers we need to mark that they are slow and
+                        // then call the broker.slowConsumer() hook if implemented
+                        if (s instanceof PrefetchSubscription) {
+                            final PrefetchSubscription sub = (PrefetchSubscription) s;
+                            if (!sub.isSlowConsumer()) {
+                                sub.setSlowConsumer(true);
+                                broker.slowConsumer(sub.getContext(), this, sub);
+                            }
+                        }
                         LOG.trace("Subscription full {}", s);
                     }
                 }
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/advisory/AdvisoryTests.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/advisory/AdvisoryTests.java
@@ -28,6 +28,7 @@ import java.util.HashSet;
 import javax.jms.BytesMessage;
 import javax.jms.Connection;
 import javax.jms.ConnectionFactory;
+import javax.jms.Destination;
 import javax.jms.Message;
 import javax.jms.MessageConsumer;
 import javax.jms.MessageListener;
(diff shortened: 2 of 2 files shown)


cshannon:
I will probably go ahead and merge this shortly as I have another advisory I am working on and want this test fixed. We can always make changes here if something isn't right.

jbonofre (approved):
LGTM from a code standpoint. I'm doing a quick test.

cshannon:
> LGTM from a code standpoint. I'm doing a quick test.

The advisory is off by default anyways and I doubt a lot of people use it. It's pretty spammy if you turn it on as it fires whenever prefetch is full (durable subs is the same way). The fact that no one complained in 10 years (or at least I don't recall but maybe someone created a Jira or posted on the mailing list) tells me there's probably not a lot of people that care.

jbonofre:
Yeah, agree, I don't think it's much used.

mattrpav:
@cshannon we've noticed the spammy behavior when advisoryForSlowConsumer is enabled and it is not a reliable source for monitoring. 

I think long term, moving to more metrics is better than Advisories for monitoring related things. Advisory back pressure is a problem-- esp in networks, so I think we are best serve with treating Advisories as 'internal communication channel for brokers'. 

For monitoring and observability, having a way to identify slow consumers is *super* important. I'll create a JIRA about turning slow consumer to a metric on destination and I think we should remove advisory for slow consumer.

Similar to the dupe-from-store-send-to-dlq is now a counter vs creating extra IO of a copy of the message to the DLQ.

cshannon:
> @cshannon we've noticed the spammy behavior when advisoryForSlowConsumer is enabled and it is not a reliable source for monitoring.
> 
> I think long term, moving to more metrics is better than Advisories for monitoring related things. Advisory back pressure is a problem-- esp in networks, so I think we are best serve with treating Advisories as 'internal communication channel for brokers'.
> 
> For monitoring and observability, having a way to identify slow consumers is _super_ important. I'll create a JIRA about turning slow consumer to a metric on destination and I think we should remove advisory for slow consumer.
> 
> Similar to the dupe-from-store-send-to-dlq is now a counter vs creating extra IO of a copy of the message to the DLQ.

I'm not against having a different/better approach but I don't think we should remove any advisories. I find several of them useful and rely on them for a lot of things. Even though this advisory is spammy someone could be using it (it works for topics) so I would be against removing it as it's off by default. If you don't need it just don't turn it on.

I'm working on a PR for a new advisory for dispatched messages to consumers which I need for a use case I have but it will be similar to this and off by default.

For now I think we should merge this as it fixes the broken behavior and again, it's off by default anyways but we advertise it as a feature.

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -2210,6 +2210,16 @@ private PendingList doActualDispatch(PendingList list) throws Exception {
                         // no further dispatch of list to a full consumer to
                         // avoid out of order message receipt
                         fullConsumers.add(s);
+
+                        //For full consumers we need to mark that they are slow and
+                        // then call the broker.slowConsumer() hook if implemented
+                        if (s instanceof PrefetchSubscription) {
```
mattrpav, line 2216: Should we honor the isAdvisoryForSlowConsumer config flag? Here or maybe better in the AdvisoryBroker.

mattrpav:
> I'm not against having a different/better approach but I don't think we should remove any advisories. 

Sure thing. I created [AMQ-9154](https://issues.apache.org/jira/browse/AMQ-9154) to track the improvement to add a SlowConsumerCount metric

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
cshannon: You can't really add that check here because that the callback could be used for many different things. Advisories is just one use case and implementation.

mattrpav (approved):
LGTM

PR #928, opened 2022-11-01
AMQ-9107 - rework performance improvement for consumer closing in managed region broker

cshannon:
This new approach just looks matching Subscriptions from the region for the destination which prevents having to store another map and falls back to the old approach if something went wrong.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagedRegionBroker.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagedRegionBroker.java
@@ -23,9 +23,9 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CopyOnWriteArraySet;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.ThreadPoolExecutor;
-import java.util.function.Consumer;
 
 import javax.jms.IllegalStateException;
+import javax.jms.JMSException;
 import javax.management.InstanceNotFoundException;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
@@ -42,6 +42,7 @@ import org.apache.activemq.broker.BrokerService;
 import org.apache.activemq.broker.ConnectionContext;
 import org.apache.activemq.broker.ProducerBrokerExchange;
 import org.apache.activemq.broker.jmx.OpenTypeSupport.OpenTypeFactory;
+import org.apache.activemq.broker.region.AbstractRegion;
 import org.apache.activemq.broker.region.Destination;
 import org.apache.activemq.broker.region.DestinationFactory;
 import org.apache.activemq.broker.region.DestinationInterceptor;
@@ -61,6 +62,7 @@ import org.apache.activemq.command.ActiveMQDestination;
 import org.apache.activemq.command.ActiveMQMessage;
 import org.apache.activemq.command.ActiveMQTopic;
 import org.apache.activemq.command.ConnectionInfo;
+import org.apache.activemq.command.ConsumerId;
 import org.apache.activemq.command.ConsumerInfo;
 import org.apache.activemq.command.Message;
 import org.apache.activemq.command.MessageAck;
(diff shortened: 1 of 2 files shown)


cshannon:
still a work in progress, need to double check for multiple matching subscriptions

mattrpav:
@cshannon should we have a test for wildcard consumers?

cshannon:
@mattrpav , @jbonofre . @lucastetreault - This is ready for review, see what you think. Part of this commit reverts the previous change and updates with a new test.

Because technically ConsumerInfos can be associated with multiple Subscriptions and because ConsumerInfo objects can change if a durable subscription goes offline and back online I thought it was generally a bad idea to store a map associating them as you'd have to keep it up to date on changes with durables (memory leak in the original PR because of that) and doesn't really work for one to many association of composite destinations even if rarely used.

The new approach here is to just use the existing Regions that already handle all of that and track in a concurrent map consumer ids and subscriptions so all we have to do is look through the region or regions (if composite) to grab the subs and then we are good. We shouldn't need it but I added a fail safe where it falls back on Exception or if it couldn't find a sub using the new method but that shouldn't happen unless something went very wrong.

I would expect performance to still be much better than before and similar to the previous commit as we are still just looking up the subscriptions in concurrent maps vs iterating.

mattrpav (approved):
LGTM -- Thanks for the unit tests to cover durable topic, queue, composite and wild card consumers.

jbonofre (approved):
LGTM, thanks.

lucastetreault:
Hey @cshannon, I had a chance to look at this today. I ran the example app against a build including this change and the performance seems good 👍  

I was trying to understand what I missed in my original changes. I ran the tests you added without your changes to ManagedRegionBroker (e.g.: https://github.com/apache/activemq/compare/main...lucastetreault:activemq:AMQ-9107-test) expecting failures but all the tests are passing even with the addition of checking that `consumerSubscriptionMap` is empty. Can you help me understand what I'm missing? 

As for the memory leak, it seems I missed one spot where I should have been removing from consumerSubscriptionMap in `public void unregisterSubscription(Subscription sub)` around [line 300](https://github.com/apache/activemq/compare/main...lucastetreault:activemq:AMQ-9107-test#diff-17798bd5f1fa8141e349c916704b7f1d7ff660f02bd3582ef696344a5bb3c723R300). I added a quick and dirty test [here](https://github.com/apache/activemq/compare/main...lucastetreault:activemq:AMQ-9107-test#diff-aa152194070293f1fb21ac1c45654d8041aa45cc3a6770aae3b3fd85961db08fR175) that passes now.

cshannon:
@lucastetreault  - The main issue here was the memory leak with durable subscriptions closing and coming back online with new consumers. This could have been fixed of course and your original fix probably would have been ok to keep. But besides the issue with the memory leak, I went with a different approach for 1 main reason which I thought was important and a secondary reason.

1. The primary reason I decided to go with the new approach is my implementation now avoids having to track duplicate metadata. Since the information is already tracked in the `ManagedRegionBroker` (since it extends `RegionBroker`) in the different `Regions` objects and well tested for a long time it made more sense to me to just use the existing maps so we don't have to worry about the extra logic of trying to track new maps/metadata and verifying things will work and are kept in sync since it's already there and tested. I saw zero benefit of attempting to track something again that we already track and can easily look up and only drawbacks in terms of now something else to keep in sync and update and things like memory leaks which happened.
2. A more minor secondary thing was you made the assumption that there is always a 1 to 1 mapping of ConsumerInfo to subscriptions. I think this is minor because this is usually the case (and I'd have to check but I think is always be the case now) even with composite destinations which is why the composite destination test works since it shares a subscription. However, in theory, this doesn't have to be which is why the original code was looking through all subscriptions for a matching ConsumerInfo and my approach preserves that by finding all subscriptions that match in the different Regions. The Regions implementation uses a 1 to 1 mapping of consumer id to subs but it also checks each region and if for some reason the code in the different regions ever associated more than one Consumer to subscription the new approach would still work.

In regards to the tests passing, it's because the main issue was the memory leak and my test class of course doesn't try and check the map that had the memory leak (`consumerSubscriptionMap`) since it no longer exists in my PR. If I run my test against your original fix and update the `testDurableConsumerRemoval` test to check that `consumerSubscriptionMap ` is size zero at the end after all consumers close (which it should be), it's not and is still 1000 and the test fails. I assume of course it would be fine if you fixed that memory leak.

lucastetreault:
Thanks for the explanation @cshannon!

PR #929, opened 2022-11-03
AMQ-9157 - Add a new optional advisory for dispatched messages

cshannon:


mattrpav (changes requested):
- [x] Confirmed classes that extend Subscription do not overload any of the updated methods
- [x] Reviewed Broker and Destination filter method changes
- [x] Reviewed unit tests 
- [x] Question: Should we add a test for message browse kicking an advisory on dispatch? 
- [ ] Question: Should we add a test for RuntimeConfigPlugin?

activemq-broker/src/main/java/org/apache/activemq/broker/util/LoggingBrokerPlugin.java
```diff
@@ -526,10 +526,7 @@ public void isFull(ConnectionContext context, Destination destination, Usage<?>
     @Override
     public void messageConsumed(ConnectionContext context, MessageReference messageReference) {
         if (isLogAll() || isLogConsumerEvents() || isLogInternalEvents()) {
-            String msg = "Unable to display message.";
```
mattrpav, line 529: Good clean-up =)

activemq-broker/src/main/java/org/apache/activemq/broker/util/LoggingBrokerPlugin.java
```diff
@@ -538,22 +535,25 @@ public void messageConsumed(ConnectionContext context, MessageReference messageR
     @Override
     public void messageDelivered(ConnectionContext context, MessageReference messageReference) {
         if (isLogAll() || isLogConsumerEvents() || isLogInternalEvents()) {
-            String msg = "Unable to display message.";
-
-            msg = messageReference.getMessage().toString();
-
+            String msg = messageReference.getMessage().toString();
```
mattrpav, line 538: Good clean-up =)

cshannon:
@mattrpav - Added a Queue browser test and also updated the runtime config tests for the new advisory property

mattrpav (approved):
LGTM

PR #931, opened 2022-11-11
[AMQ-9161] Relocate activeio classes used by JDBC Journal impl  and deprecate JDBC Journal

mattrpav:
- [x] Relocate the subset of activeio classes used by jdbc journal to jdbc-store module
- [x] Remove activeio-core (no other module uses it)
- [x] Rename System properties activeio -> activemq.store.journal
- [x] Update javadoc @see: references
- [x] Deprecate JDBC Journal for removal

Note: The activeio classes are used by the JDBC Journal from 4.x era, not the more modern JDBCPersistenceAdapter.

### DIFF
--- a/activemq-console/pom.xml
+++ b/activemq-console/pom.xml
@@ -49,11 +49,6 @@
       <groupId>${project.groupId}</groupId>
       <artifactId>activemq-kahadb-store</artifactId>
     </dependency>
-    <dependency>
-      <groupId>${project.groupId}</groupId>
-      <artifactId>activeio-core</artifactId>
-    </dependency>
-
     <!-- geronimo -->
     <dependency>
       <groupId>jakarta.jms</groupId>
--- a/activemq-jdbc-store/pom.xml
+++ b/activemq-jdbc-store/pom.xml
@@ -53,12 +53,6 @@
       <artifactId>derbytools</artifactId>
       <optional>true</optional>
     </dependency>
-    <dependency>
-      <groupId>${project.groupId}</groupId>
-      <artifactId>activeio-core</artifactId>
-      <optional>true</optional>
-    </dependency>
-
     <!-- =============================== -->
     <!-- Testing Dependencies -->
     <!-- =============================== -->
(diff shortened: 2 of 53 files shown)


cshannon:
This seems fine to me since only JDBC uses it and the module hasn't been updated in over 10 years but I want to get some other PMC members to chime in also and see what they think about moving the entire module.

tabish121:
> This seems fine to me since only JDBC uses it and the module hasn't been updated in over 10 years but I want to get some other PMC members to chime in also and see what they think about moving the entire module.

I guess it is okay although I'm not sure how moving will make it any less unmaintained and probably a good target for deprecation and removal altogether.  Main thing that occurred to me was, why now?

mattrpav:
> I guess it is okay although I'm not sure how moving will make it any less unmaintained and probably a good target for deprecation and removal altogether. Main thing that occurred to me was, why now?

@tabish121 agreed. The intent of this changes is to take a step forward towards eventual deprecation. This approach is lieu of an activeio upgrade + release.

tabish121:
> > I guess it is okay although I'm not sure how moving will make it any less unmaintained and probably a good target for deprecation and removal altogether. Main thing that occurred to me was, why now?
> 
> @tabish121 agreed. The intent of this changes is to take a step forward towards eventual deprecation. This approach is lieu of an activeio upgrade + release.

If that's the intention then I'd spell it out more clearly in the JIRA so folks can see.

mattrpav:
> If that's the intention then I'd spell it out more clearly in the JIRA so folks can see.

JIRA updated with those notes, and I went ahead and added a task to @Deprecate the JDBC Journal now, so it'll be in 5.18.0 as deprecated.

gemmellr:
If you are deprecating one type of JDBC support but keeping another type of support, the JIRA really doesnt make that clear...it just reads like [eventual] total removal of JDBC support.

Its not clear to the classes need relocated in order to deprecate the journal, particularly when the main @Deprecated(forRemoval = true) additions are only in the classes already in the tree? (Might be nice if the deprecations noted to use the other stuff instead, if that is the intent).

Changing system properties seems like a breaking change of sorts, which seems a bit weird do while only keeping this unmaintained stuff around in the first place to support any existing users, and then only to remove it later. So its still there if people want it, but doesnt necessarily behave like it did when used previously, which may or may not be immediately obvious to them? Seems a bit off.

I'd probably go a different route and either just remove it now, or leave it essentially as-is but deprecated for removal and maybe logging a warning or error message if used to note its going away and when...then later either again simply remove it, or first critically disable it with an exception indicating its done for but noting e.g a sys prop to turn it back on for one final chance to move off it. A deprecation + logging style change could also be backported to 5.17.x to get the ball rolling.

mattrpav:
Replaced by #989

PR #937, opened 2022-11-18
Added destination field to Job

shikhar97gupta:
JIRA issue: https://issues.apache.org/jira/projects/AMQ/issues/AMQ-9166?filter=allopenissues&orderby=created+DESC%2C+priority+DESC%2C+updated+DESC


mattrpav:
Hi @shikhar97gupta -- I'll start reviewing the PR this weekend.

Please review: https://community.apache.org/newcomers/index.html

shikhar97gupta:
I am unable to replicate error on running the tests with the same command as in the checks tab (mvn -B -e -fae test).

The activemq code clone on my system gives a different error (fails a broker test) on running the same command compared to the error mentioned in the checks tab.

activemq-broker/src/main/java/org/apache/activemq/broker/scheduler/Job.java
```diff
@@ -78,4 +80,6 @@ public interface Job {
      */
     public int getExecutionCount();
 
+    public abstract ActiveMQDestination getDest();
```
mattrpav, line 83: Let's change this to just be a 'String' type that we can specify as the destination.getPhysicalName()

mattrpav (changes requested):
I'm thinking we can change the approach to be simpler:

activemq-broker/src/main/java/org/apache/activemq/broker/scheduler/Job.java
shikhar97gupta: Something like this? 

getAllJobs JMX operation:

```diff

public class JobSchedulerView implements JobSchedulerViewMBean {

   public TabularData getAllJobs(String startTime, String finishTime) throws Exception {
         OpenTypeFactory factory = OpenTypeSupport.getFactory(Job.class);
         CompositeType ct = factory.getCompositeType();
         TabularType tt = new TabularType("Scheduled Jobs", "Scheduled Jobs", ct, new String[] { "jobId" });
         TabularDataSupport rc = new TabularDataSupport(tt);
         long start = JobSupport.getDataTime(startTime);
         long start = JobSupport.getDataTime(startTime);
         long finish = JobSupport.getDataTime(finishTime);
         List<Job> jobs = this.jobScheduler.getAllJobs(start, finish);
+        OpenWireFormat wireFormat = new OpenWireFormat();
         for (Job job : jobs) {
+            Message msg = (Message) wireFormat.unmarshal(new ByteSequence(job.getPayload()));
+            ActiveMQDestination destination = (ActiveMQDestination) msg.getJMSDestination();
+            job.setDestinationName(destination.getPhysicalName());
             rc.put(new CompositeDataSupport(ct, factory.getFields(job)));
         }
         return rc;

```

mattrpav: Yep!

shikhar97gupta:
@mattrpav Any other changes required?

mattrpav:
@shikhar97gupta these changes are looking good! One final request -- move the 'with destination name' to a separate method. This will ensure that users that use the existing method do not have any performance impact when viewing scheduled jobs.

shikhar97gupta:
> @shikhar97gupta these changes are looking good! One final request -- move the 'with destination name' to a separate method. This will ensure that users that use the existing method do not have any performance impact when viewing scheduled jobs.

@mattrpav Applied this change for all 3 methods separately -- getAllJobs(), getAllJobs(String, String) and getNextScheduleJobs()

mattrpav:
@shikhar97gupta looks good! If you can get the last housekeeping tasks done, we'll include this change in the next set of 5.17.x, 5.18.x and 6.1.x releases.

- [ ] Add a unit test to validate the destination name field is working correctly in this test: 
```
activemq-unit-tests/src/test/java/org/apache/activemq/broker/scheduler/JobSchedulerJmxManagementTests.java
```
- [ ] Rebase your PR off of 'main'

jbonofre (approved):
It looks good to me. Thanks !

PR #942, opened 2022-11-21
[AMQ-9102] Add http.nonProxyHost support in HTTP transport

jbonofre:


mattrpav (commented):
Is there a way to create a unit test for the scenario?

jbonofre:
@mattrpav not sure we have already tests on HttpConnector proxy support. Let me check.

jbonofre:
@mattrpav I checked and we don't have any test around proxy (getProxyPort() is not actually HTTP proxy, it's just port found locally for the test). I tested manually and it looks good.

PR #947, opened 2023-01-03
Failing unit test for AMQ-9187 and a fix

devlinm:
Hi,
Added a test case to demonstrate the bug described in AMQ-9187 and a fix.

The fix is very simple - swapping the order of the AdvisoryBroker and SchedulerBroker interceptors. I don't think there will be any weird side effects of doing this because the SchedulerBroker only intercepts the 'send' method whereas the AdvisoryBroker doesn't intercept that 'send' method. Let me know what you think.

I would like to backport this to the 5.17 and 5.16 branches once we get this accepted into main. Thanks!

regards,
- Martin

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -2426,6 +2426,10 @@ public class BrokerService implements Service {
      * @throws IOException
      */
     protected Broker addInterceptors(Broker broker) throws Exception {
+        if (isAdvisorySupport()) {
+            // AMQ-9187 - the AdvisoryBroker must be after the SchedulerBroker
+            broker = new AdvisoryBroker(broker);
+        }
         if (isSchedulerSupport()) {
             SchedulerBroker sb = new SchedulerBroker(this, broker, getJobSchedulerStore());
             sb.setMaxRepeatAllowed(maxSchedulerRepeatAllowed);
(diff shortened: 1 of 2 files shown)


jbonofre:
@devlinm Thanks ! do you mind to squash the commits in one ?

devlinm:
@jbonofre no problem thanks.

devlinm:
I see some unit tests broke... I'll have a look at that today.

devlinm:
@jbonofre This test is failing: org.apache.activemq.advisory.AdvisoryTests#testMessageExpiredAdvisoryDurableServer due to a timeout. It only fails when the parameter persistent=true. This is suspicious as its an Advisory Test and I did switch the order of the AdivisoryBroker.

So I reverted my change and retested - it turns out that its still failing with the old code! So doesn't look like its actually related to my change AFAIK. Does this test fail in your environment? I'm using corretto openjdk version "11.0.17"

I'll look at this again later in the week when/if time permits.

jbonofre:
Let me try.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -2426,6 +2426,9 @@ protected DestinationInterceptor[] createDefaultDestinationInterceptor() {
      * @throws IOException
      */
     protected Broker addInterceptors(Broker broker) throws Exception {
+        if (isAdvisorySupport()) {
```
mattrpav, line 2429: Please add a comment here that the order matters and include JIRA number

mattrpav (changes requested):
@devlinm thanks for the contribution.

Please see requested change(s).

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
devlinm: done 👍

devlinm:
BTW I squashed the commits. 👍

jbonofre:
I'm adapting this PR to use `jakarta` namespace instead of `javax`. I will merge manually.
I'm testing now ;)

PR #949, opened 2023-01-05
Track scheduler activity

maxfortun:
[AMQ-9188](https://issues.apache.org/jira/browse/AMQ-9188)
Augment scheduler to forward scheduled and dispatched messages to a destination, if such is configured.
e.g.
<broker ... schedulerSupport="true" schedulerActivityDestination="topic://ActiveMQ.Scheduler.Activity" >
If schedulerActivityDestination is not set, current behavior is unaffected.
Otherwise...
When message is successfully scheduled scheduler should set property AMQ_SCHEDULER_ACTIVITY to SCHEDULED and forward scheduled message to the schedulerActivityDestination
When message is successfully dispatched scheduler should set property AMQ_SCHEDULER_ACTIVITY to DISPATCHED and forward scheduled message to the schedulerActivityDestination

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -231,6 +231,7 @@ public class BrokerService implements Service {
     private boolean forceStart = false;
     private IOExceptionHandler ioExceptionHandler;
     private boolean schedulerSupport = false;
+    private String schedulerActivityDestination = null;
     private int maxSchedulerRepeatAllowed = MAX_SCHEDULER_REPEAT_ALLOWED;
     private File schedulerDirectoryFile;
     private Scheduler scheduler;
@@ -1693,11 +1694,11 @@ public class BrokerService implements Service {
     }
     
     public boolean isEnableMessageExpirationOnActiveDurableSubs() {
-    	return enableMessageExpirationOnActiveDurableSubs;
+        return enableMessageExpirationOnActiveDurableSubs;
     }
     
     public void setEnableMessageExpirationOnActiveDurableSubs(boolean enableMessageExpirationOnActiveDurableSubs) {
-    	this.enableMessageExpirationOnActiveDurableSubs = enableMessageExpirationOnActiveDurableSubs;
+        this.enableMessageExpirationOnActiveDurableSubs = enableMessageExpirationOnActiveDurableSubs;
     }
 
     public boolean isUseVirtualTopics() {
@@ -2976,6 +2977,20 @@ public class BrokerService implements Service {
      */
     public void setSchedulerSupport(boolean schedulerSupport) {
         this.schedulerSupport = schedulerSupport;
+    }
+
+    /**
+     * @param schedulerActivityDestination the schedulerActivityDestination to set
+     */
+    public void setSchedulerActivityDestination(String schedulerActivityDestination) {
+        this.schedulerActivityDestination = schedulerActivityDestination;
+    }
+
+    /**
+     * @return the schedulerActivityDestination
+     */
+    public String getSchedulerActivityDestination() {
+        return schedulerActivityDestination;
     }
 
     /**
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -79,6 +79,7 @@ import org.apache.activemq.broker.region.virtual.VirtualDestination;
 import org.apache.activemq.broker.region.virtual.VirtualDestinationInterceptor;
 import org.apache.activemq.broker.region.virtual.VirtualTopic;
 import org.apache.activemq.broker.scheduler.JobSchedulerStore;
+import org.apache.activemq.broker.scheduler.JobSchedulerActivityListener;
 import org.apache.activemq.broker.scheduler.SchedulerBroker;
 import org.apache.activemq.broker.scheduler.memory.InMemoryJobSchedulerStore;
 import org.apache.activemq.command.ActiveMQDestination;
@@ -231,7 +232,6 @@ public class BrokerService implements Service {
     private boolean forceStart = false;
     private IOExceptionHandler ioExceptionHandler;
     private boolean schedulerSupport = false;
-    private String schedulerActivityDestination = null;
     private int maxSchedulerRepeatAllowed = MAX_SCHEDULER_REPEAT_ALLOWED;
     private File schedulerDirectoryFile;
     private Scheduler scheduler;
@@ -245,6 +245,7 @@ public class BrokerService implements Service {
     private boolean networkConnectorStartAsync = false;
     private boolean allowTempAutoCreationOnSend;
     private JobSchedulerStore jobSchedulerStore;
+    private JobSchedulerActivityListener jobSchedulerActivityListener;
     private final AtomicLong totalConnections = new AtomicLong();
     private final AtomicInteger currentConnections = new AtomicInteger();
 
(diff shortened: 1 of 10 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -231,6 +231,7 @@ public class BrokerService implements Service {
     private boolean forceStart = false;
     private IOExceptionHandler ioExceptionHandler;
     private boolean schedulerSupport = false;
+    private String schedulerActivityDestination = null;
```
mattrpav, line 234: Please add a default value from a constant

mattrpav (changes requested):

mattrpav:
@maxfortun are you able to share any more details about the use case? I'm pretty sure a standard plugin could support this without needing to modify broker internals or the current scheduler code.

activemq-broker/src/main/java/org/apache/activemq/broker/scheduler/SchedulerBroker.java
```diff
@@ -353,6 +355,12 @@ private void doSchedule(Message messageSend, Object cronValue, Object periodValu
 
         getInternalScheduler().schedule(jobId.toString(),
                 new ByteSequence(packet.data, packet.offset, packet.length), cronEntry, delay, period, repeat);
+
```
maxfortun, line 358: @mattrpav this block is the main reason for a PR. As there is no way to get to the job id at the time of the scheduling.

activemq-broker/src/main/java/org/apache/activemq/broker/scheduler/SchedulerBroker.java
```diff
@@ -418,6 +426,12 @@ public void scheduledJob(String id, ByteSequence job) {
             producerExchange.setMutable(true);
             producerExchange.setProducerState(new ProducerState(new ProducerInfo()));
             super.send(producerExchange, messageSend);
+
```
maxfortun, line 429: @mattrpav Once message is dispatched it has scheduledJobId property, so I can use that, but I figured it is nice to have a single source to monitor scheduler activity.

activemq-broker/src/main/java/org/apache/activemq/broker/scheduler/SchedulerBroker.java
```diff
@@ -353,6 +355,12 @@ private void doSchedule(Message messageSend, Object cronValue, Object periodValu
 
         getInternalScheduler().schedule(jobId.toString(),
                 new ByteSequence(packet.data, packet.offset, packet.length), cronEntry, delay, period, repeat);
+
+        if(jobSchedulerActivityListener != null) {
+            msg.setProperty(ScheduledMessage.AMQ_SCHEDULED_ID, jobId.toString());
+            msg.setProperty(ScheduledMessage.AMQ_SCHEDULER_ACTIVITY, ScheduledMessage.AMQ_SCHEDULER_ACTIVITY_SCHEDULED);
+            jobSchedulerActivityListener.scheduled(this, msg);
+        }
```
maxfortun, line 363: @mattrpav this block is the main reason for a PR. As there is no way to get to the job id at the time of the scheduling.

activemq-broker/src/main/java/org/apache/activemq/broker/scheduler/SchedulerBroker.java
```diff
@@ -418,6 +426,12 @@ public void scheduledJob(String id, ByteSequence job) {
             producerExchange.setMutable(true);
             producerExchange.setProducerState(new ProducerState(new ProducerInfo()));
             super.send(producerExchange, messageSend);
+
+            if(jobSchedulerActivityListener != null) {
+                Message msg = messageSend.copy();
+                msg.setProperty(ScheduledMessage.AMQ_SCHEDULER_ACTIVITY, ScheduledMessage.AMQ_SCHEDULER_ACTIVITY_DISPATCHED);
+                jobSchedulerActivityListener.dispatched(this, msg);
+            }
```
maxfortun, line 434: @mattrpav Once message is dispatched it has scheduledJobId property, so I can use that, but I figured it is nice to have a single source to monitor scheduler activity.

mattrpav:
@maxfortun sounds like the need is for additional extensions for registerJob (and for completeness, we should add unregisterJob). 

There is already a JobListener interface that has an operation to catch when scheduled messages are fired. I'm wondering if we could extend that by adding the 2 add'l methods.

Then from there, confirm wiring in a custom JobListener (aka a mini-plugin just for Job events).

Thoughts?

maxfortun:
> @maxfortun sounds like the need is for additional extensions for registerJob (and for completeness, we should add unregisterJob).
> 
> There is already a JobListener interface that has an operation to catch when scheduled messages are fired. I'm wondering if we could extend that by adding the 2 add'l methods.
> 
> Then from there, confirm wiring in a custom JobListener (aka a mini-plugin just for Job events).
> 
> Thoughts?

I also initially thought that the JobListener is the correct place for this feature, but since I am not familiar enough with the broker internals and maintainers philosophy I did not want to make any changes that are not backward compatible. If we change the JobListener interface then the upgrades for people who have any custom scheduler modules for the broker will no longer be a drop in easy-peasy process and would require quite a bit of work. Which may be ok, if that's how activemq is developed. Also, currently there is only 1 listener, and it is the `SchedulerBroker`, which is added and removed when broker starts and stops. I presume this is for the failover support. Using job listeners may make this feature more of a rearchitecture than a feature. I am not certain that this deserves so much effort. Thoughts?

mattrpav:
> I also initially thought that the JobListener is the correct place for this feature, but since I am not familiar enough with the broker internals and maintainers philosophy I did not want to make any changes that are not backward compatible. If we change the JobListener interface then the upgrades for people who have any custom scheduler modules for the broker will no longer be a drop in easy-peasy process and would require quite a bit of work. Which may be ok, if that's how activemq is developed. Also, currently there is only 1 listener, and it is the `SchedulerBroker`, which is added and removed when broker starts and stops. I presume this is for the failover support. Using job listeners may make this feature more of a rearchitecture than a feature. I am not certain that this deserves so much effort. Thoughts?

We can add a default no-op method to fix the upgrade compatibility issue.

I think overall, adding a couple methods to existing internal interface is preferred over adding a new interface, config, and other internal changes. Also, the approach of only having one handler is limiting.

Note-- the existing SchedulerBroker implements the JobListener only to have a handle to internal broker for publishing the message. It is an o-o hack of sorts to reduce the need to inject a BrokerService to the JobListener to start deliverying messages.

Once the interface is updated, adding an AdvisoryJobListener would be trivial, and you could still add add'l listeners to do pub-sub stuff for your internal job track sync tool.

maxfortun:
There is one other issue with this approach. In its current design the JobListener is blind to the message and its attributes, it treats it as a binary payload. It may be quite a lift to wire it for what I need to happen. Which is to forward a meaningful activity message. As far as multiple activity trackers are concerned, it is rather straightforward to implement a  class that wraps many trackers into one.

maxfortun:
Something like this:
```xml
        <jobSchedulerActivityListener>
            <bean xmlns="http://www.springframework.org/schema/beans" id="jobSchedulerMultipleActivityListeners"
                class="org.apache.activemq.broker.scheduler.JobSchedulerMultipleActivityListeners">
                <property name="listeners">
                    <list>
                        <bean xmlns="http://www.springframework.org/schema/beans"
                            id="jobSchedulerForwardingActivityListener"
                            class="org.apache.activemq.broker.scheduler.JobSchedulerForwardingActivityListener" />
                    </list>
                </property>
            </bean>
        </jobSchedulerActivityListener>
```

mattrpav:
Yeah, I just can't get there with the current approach. I think this could all be handled within the JobScheduler/JobSchedulerFacade and not have to bubble it all the way up to BrokerService.

I'd vote -1 for this change as-is.

Pinging @cshannon to see if he has any add'l thoughts.

My suggested approaches:
1. Extend current JobListener
2. Extend advisory support to send an advisory message w/ the job id, message id and target destination (this would be add add'l broker filter methods, and then support in the AdvisoryBroker so it can be turned off when advisory support is disabled)

maxfortun:
Closing this PR as a leaner approach was suggested. Submitting #950 as a proposal in continuation of this discussion.

PR #951, opened 2023-01-10
AMQ-9192 - Fix flaky AdvisoryTests

cshannon:
Properly shutdown broker for each test and speed up tests by sending less messages


cshannon:
I created a PR here mostly to let the tests run to see if it fixes things and then it can be merged and I will back port it to 5.17.x as well (along with any tweaks)

cshannon:
There might still be something to tweak but it looks like it fixed the vast majority of the Advisory test issues so I'm going to merge this and let it run against main and then fix whatever else.

PR #953, opened 2023-01-10
[AMQ-8354] Add helper methods needed for replica plugin.

NikitaShupletsov:
AMQ-8354

extracted internal changes from https://github.com/apache/activemq/pull/848 to make it easier to review.

Added some new methods that will be used only in replica plugin.
Updated `processDispatchNotification` and related methods to work correctly(it's not been used since pure master-slave feature was deleted). it will also be used in the plugin

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
@@ -45,6 +45,7 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
+import java.util.stream.Collectors;
 
 import javax.jms.InvalidSelectorException;
 import javax.jms.JMSException;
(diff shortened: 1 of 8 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -1296,9 +1297,36 @@ public QueueMessageReference getMessage(String id) {
         return null;
     }
 
+    public List<MessageId> getAllMessageIds() throws Exception {
```
NikitaShupletsov, line 1300: here out of memory error is possible: https://github.com/apache/activemq/pull/848#discussion_r1062161446
will address it

activemq-broker/src/main/java/org/apache/activemq/broker/scheduler/SchedulerBroker.java
```diff
@@ -54,7 +54,7 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-public class SchedulerBroker extends BrokerFilter implements JobListener {
+public class SchedulerBroker extends MutableBrokerFilter implements JobListener {
```
michaelandrepearce, line 57: Wont this be overhead now for use cases not needing this? due to the sync in mutable thats needed.

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -2387,8 +2495,8 @@ private QueueMessageReference getMatchingMessage(MessageDispatchNotification mes
                     messages.reset();
                     while (messages.hasNext()) {
                         MessageReference node = messages.next();
-                        messages.remove();
```
michaelandrepearce, line 2390: Why move this?

activemq-broker/src/main/java/org/apache/activemq/broker/scheduler/SchedulerBroker.java
NikitaShupletsov: I wasn't able to find any information about possible `AtomicReference` overhead. I believe if there is some, it should be almost not noticeable

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
NikitaShupletsov: The idea of this change is to delete only the message that we are looking for. Because deleting all messages up to the one we need prevents reading them in future. Basically if we delete them here, we will not be able to get them anymore.

michaelandrepearce: And? That's now unneeded change in behaviour

NikitaShupletsov: I am not sure why.  this method is used to "deliver" a specific message by id to a specific consumer. we still want to have the access to all other messages. in case we want to "deliver" them too, or in case there is a real consumer

PR #954, opened 2023-01-10
NO-JIRA - Set tests to rerun failing tests 2 times

cshannon:
This should help reduce failures from flaky tests and make the build more reliable

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -99,7 +99,7 @@ pipeline {
                 echo 'Running tests'
                 // all tests is very very long (10 hours on Apache Jenkins)
                 // sh 'mvn -B -e test -pl activemq-unit-tests -Dactivemq.tests=all'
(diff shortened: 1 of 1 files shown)


cshannon:
@jbonofre and @mattrpav - What do you guys think about this change? I've been using this a lot recently (I first saw it used on Apache Accumulo) and it's been really nice to reduce intermittent failures. I think we should use it for 5.x to reduce some of the flaky test issues. We can try 2 but could even go higher (I wouldn't go more than 3-5). The downside is it may hide some issues but a real issue should likely fail even if running twice and I think the benefit of reducing the constant build failures will be good.

mattrpav:
@cshannon I think it’s a good approach and agree with you on all points.

mattrpav (approved):
LGTM

cshannon:
Cool I'll merge it in

PR #955, opened 2023-01-11
[#9194] Add JDK 19 to Jenkins build

mattrpav:


mattrpav:
jdk-19 compile works without any code changes

PR #961, opened 2023-02-01
[AMQ-9204] Upgrade to jetty 9.4.50.v20221201

mattrpav:


jbonofre:
I did other changes with this update as Jetty change its layers (especially for websockets).

PR #964, opened 2023-02-01
[AMQ-9207] Upgrade various dependencies

mattrpav:
Dependency updates:

com.thoughtworks.xstream:xstream .................... 1.4.19 -> 1.4.20 
commons-daemon:commons-daemon ......................... 1.3.2 -> 1.3.3 
org.apache.ant:ant ................................ 1.10.12 -> 1.10.13 
org.apache.shiro:shiro-core ......................... 1.10.1 -> 1.11.0 
org.apache.shiro:shiro-spring ....................... 1.10.1 -> 1.11.0 
org.codehaus.jettison:jettison ........................ 1.5.2 -> 1.5.3 
regexp:regexp ............................................. 1.3 -> 1.5 
org.apache.httpcomponents:httpclient ................ 4.5.13 -> 4.5.14 
org.apache.httpcomponents:httpcore .................. 4.4.15 -> 4.4.16

### DIFF
--- a/activemq-rar/pom.xml
+++ b/activemq-rar/pom.xml
@@ -110,8 +110,8 @@
           <artifactId>junit</artifactId>
         </exclusion>
         <exclusion>
-          <groupId>regexp</groupId>
-          <artifactId>regexp</artifactId>
+          <groupId>jakarta-regexp</groupId>
+          <artifactId>jakarta-regexp</artifactId>
         </exclusion>
         <exclusion>
           <groupId>javax.jmdns</groupId>
--- a/activemq-unit-tests/pom.xml
+++ b/activemq-unit-tests/pom.xml
@@ -240,8 +240,8 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>regexp</groupId>
-      <artifactId>regexp</artifactId>
+      <groupId>jakarta-regexp</groupId>
+      <artifactId>jakarta-regexp</artifactId>
       <scope>test</scope>
     </dependency>
     <dependency>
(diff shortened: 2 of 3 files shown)


jbonofre:
Please ping me before such kind of updates, I already had all atomic dep updates ready.

jbonofre:
I agree for the PR, but the Jira should be atomic as it's better for release notes IMHO.
We can group several dep updates in one PR/commit as soon as we have one Jira per update.

I will rebase my changes (I have others on the way).

mattrpav:
> I agree for the PR, but the Jira should be atomic as it's better for release notes IMHO. We can group several dep updates in one PR/commit as soon as we have one Jira per update.

That sounds good, I'll go ahead and update JIRA for individual JIRAs for the release notes. Thx

cshannon:
> > I agree for the PR, but the Jira should be atomic as it's better for release notes IMHO. We can group several dep updates in one PR/commit as soon as we have one Jira per update.
> 
> That sounds good, I'll go ahead and update JIRA for individual JIRAs for the release notes. Thx

Right, for minor stuff there's no reason to have like 10 commits when they can be combined into one commit like this but the major stuff like Spring for sure should be by itself.

For Jira I think we should be able to create one top level Jira issue for the minor dependency updates and then have subtasks for each individual update. Then the PR that does the multiple changes at once could just reference the main Jira but the release notes would still list all the individual dependencies because of the sub tasks (I think but need to test)

jbonofre:
@cshannon yeah it's what I meant. The most important for our users is to have details in the release notes.

PR #965, opened 2023-02-02
[AMQ-9217] Fix per-destination message audits in IndividualDeadLetterStrategy

mattrpav:


cshannon (changes requested):
I would move most of the changes into the IndividualDeadLetterStrategy and not put it into AbstractDeadLetterStrategy. 

Since a shared DLQ is the default and only uses a single destination it makes no sense to have a flag to enable audit per destination and to add in all of the extra logic/sync to do so when you can only have 1 message audit anyways and the flag does nothing. It's also misleading to the user.

I think the abstract class should basically not change at all except for maybe adding the new method lookupActiveMQMessageAudit() and by default it just uses the default messageAudit that exists today. That message audit can be made protected and then in the individual DLQ strategy you can add all the logic to handle multiple DLQs and override that method to use the per destination audits if enabled.

cshannon:
Also, I am wondering if this change should actually have a flag at all..seems like it's just a bug and should be fixed for IndividualDeadLetterStrategy. I think the only thing that really makes sense is to have an audit per DLQ destination.

If a user has enabled a DLQ per destination and are using something like Virtual Topics it is pretty weird to only DLQ one of the messages to one of the destinations if there are several consumer queues and more than one consumer caused the message to go to the DLQ. I guess someone reading the DLQ could be using a wildcard consumer so they would end up with duplicates which is a behavior change but if doing that then it would make more sense to use the shared strategy.

Maybe for a future release (like 5.18.0 or 5.19.0) for 5.18.0 we change the default behavior so that the individual strategy defaults to using an audit per destination as that seems like the more "correct" thing to do while keeping a flag to disable it i guess.

jbonofre:
I don't think a flag is what we should do. For me, it sounds more like a bug, so just changing the default behavior is OK.

cshannon:
> I don't think a flag is what we should do. For me, it sounds more like a bug, so just changing the default behavior is OK.

Yeah this probably makes the most sense, it would be a behavior change but the current behavior just seems broken and a bug so might as well change it for the individual strategy. Anytime we add flags 99% of the time no one is going to change the default so we should just fix it I think unless someone can think of a good reason not to change it.

mattrpav:
Changes:

mattrpav:
Changes:

- [ ] Move logic for per-destination audits to IndividualDeadLetterStrategy

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/AbstractDeadLetterStrategy.java
```diff
@@ -108,20 +111,25 @@ public void setExpiration(long expiration) {
...
     }
 
     public void setMaxProducersToAudit(int maxProducersToAudit) {
-        messageAudit.setMaximumNumberOfProducersToTrack(maxProducersToAudit);
+        this.maxProducersToAudit = maxProducersToAudit;
     }
 
     public void setMaxAuditDepth(int maxAuditDepth) {
-        messageAudit.setAuditDepth(maxAuditDepth);
+        this.maxAuditDepth = maxAuditDepth;
```
cshannon, line 125: This sets the variable but no longer sets the shared audit so that will break things.

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/AbstractDeadLetterStrategy.java
```diff
@@ -108,20 +111,25 @@ public void setExpiration(long expiration) {
         this.expiration = expiration;
     }
 
+
```
cshannon, line 114: I would get rid of the extra random white space changes in the file like here.

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/AbstractDeadLetterStrategy.java
```diff
@@ -31,13 +31,15 @@ public abstract class AbstractDeadLetterStrategy implements DeadLetterStrategy {
     private boolean processNonPersistent = false;
     private boolean processExpired = true;
     private boolean enableAudit = true;
-    private final ActiveMQMessageAudit messageAudit = new ActiveMQMessageAudit();
```
cshannon, line 34: There's no real reason to rename this variable or change this anymore.

cshannon: Also if we decide to not have a flag for Individual and only use a message audit per destination (which I think makes more sense) then we can actually just move this message audit variable down into the Shared strategy. The abstract wouldn't need to have anything defined except the abstract method for looking up the message audit as leaving this here wouldn't make sense if it won't be used by Individual.

cshannon:
@mattrpav - I reworked this a bit and have a commit here as a follow on to yours that I think makes a bit more sense: https://github.com/cshannon/activemq/commit/bf30a92809d49c7857843af7fdcbc95a114640cd

It simplifies things and drops the flag as I think it's just a bug that should be fixed. The more I think about it the more it just seems completely wrong to not have an audit per DLQ destination. I didn't do much with the tests so that would need to be looked at. I also wonder if 10k is too high and maybe we should lower the default because of memory usage. 

The last thing is I don't really like the LRU cache we have, it's super outdated. It would be much better to use something like Caffeine where we don't need to synchronize. That being said, that's probably a bigger task as we'd need to pull in that dependency and I would think we'd just want to replace it everywhere, not just here. Also, (in theory) the DLQ stuff shouldn't be hit all the time so hopefully performance wouldn't be a big deal as I wouldn't normally expect a ton of attempts to acquire the lock at the same time but I guess it's possible if things are going really bad.

mattrpav:
@cshannon sounds good. I merged your changes (I prefer that abstract get/setter approach you did too). 

Add'l fixes -- updated the 'disabled' test to now confirm the sharedeadletterstrategy dropping messages during audit, and removed a dead store field.

Ready for final review and merge.

cshannon (approved):

cshannon:
I should add a quick note that I merged this but this doesn't solve all issues (but makes things much better). It uses the original destination and not the DLQ destination the message is going to which means that in some cases (like if destinationPerDurableSubscriber flag is true) a copy of the message for each destination still wouldn't be DLQ'd as more than one destination still shares a message audit. So while this greatly improves things it isn't perfect. A follow in issue/PR can be created to improve it further if desired to handle the more complex cases.

PR #968, opened 2023-02-07
WIP: [AMQ-9218] Jakarta transition module for activemq-client

mattrpav:


mattrpav:
#983 should merge first, and we can re-base this without j2ee spec

activemq-client-jakarta/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+                <exclusion>
+                    <groupId>jakarta.jms</groupId>
+                    <artifactId>jakarta.jms-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>jakarta.jms</groupId>
+            <artifactId>jakarta.jms-api</artifactId>
+            <version>3.0.0</version>
```
cshannon, line 44: Any versions in this file, like here, should really be managed in the root pom under dependency management. It makes tracking versions by the project much easier if it's in one spot.

activemq-client-jakarta/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.thoughtworks.xstream</groupId>
+            <artifactId>xstream</artifactId>
+            <version>${xstream-version}</version>
```
cshannon, line 49: Version not needed, see comment for jakarta.jms-api version

activemq-client-jakarta/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+    </parent>
+    <artifactId>activemq-client-jakarta</artifactId>
+    <packaging>bundle</packaging>
+    <name>ActiveMQ :: Client Jakarta</name>
+    <description>Jakarta transition module for ActiveMQ Client implementation</description>
+    <dependencies>
+        <dependency>
+            <groupId>org.apache.activemq</groupId>
+            <artifactId>activemq-client</artifactId>
+            <version>${project.version}</version>
```
cshannon, line 33: Version not needed, see comment for jakarta.jms-api version

activemq-client-jakarta/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+                        <phase>initialize</phase>
+                        <goals>
+                            <goal>unpack</goal>
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>org.apache.activemq</groupId>
+                                    <artifactId>activemq-client</artifactId>
+                                    <version>${activemq.version}</version>
```
cshannon, line 69: Version not needed, see comment for jakarta.jms-api version

activemq-client-jakarta/pom.xml
```diff
@@ -0,0 +1,166 @@
...
+                                </artifactItem>
+                            </artifactItems>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>com.google.code.maven-replacer-plugin</groupId>
+                <artifactId>replacer</artifactId>
+                <version>1.5.3</version>
```
cshannon, line 84: like dependencies, plugins should have their version managed in the root pom in pluginManagement

cshannon (changes requested):
Besides the version comments (move to root pom to manage) the main thing I noticed was on the first build just running a "mvn clean install -DskipTests" it blew up because it couldn't find the sources jar for activemq-client.

`[ERROR] Failed to execute goal org.apache.maven.plugins:maven-dependency-plugin:3.3.0:unpack (unpack-source) on project activemq-client-jakarta: Unable to find/resolve artifact.: Could not find artifact org.apache.activemq:activemq-client:jar:sources:5.18.0-SNAPSHOT`

I think you will need to update activemq-client to always build the sources jar by default.

mattrpav:
Cleaned up the dependency:tree for application users to have a cleaner transitive dependency tree.

 - [x] Set activemq-client as 'provided' 
 - [x] Directly depend on slf4j
 - [x] Directly depend on hawtbuf

```
[INFO] org.apache.activemq:activemq-client-jakarta:bundle:5.18.0-SNAPSHOT
[INFO] +- org.apache.activemq:activemq-client:jar:5.18.0-SNAPSHOT:provided
[INFO] +- jakarta.jms:jakarta.jms-api:jar:3.1.0:compile
[INFO] +- org.fusesource.hawtbuf:hawtbuf:jar:1.11:compile
[INFO] +- org.slf4j:slf4j-api:jar:1.7.36:compile
[INFO] +- com.thoughtworks.xstream:xstream:jar:1.4.20:provided
[INFO] |  \- io.github.x-stream:mxparser:jar:1.2.2:provided
[INFO] |     \- xmlpull:xmlpull:jar:1.1.3.1:provided
[INFO] \- junit:junit:jar:4.13.2:test
[INFO]    \- org.hamcrest:hamcrest-core:jar:1.3:test
```

cshannon:
@mattrpav - going forward instead of always force pushing it's probably better to just push new commits. The issue is any comments/history get blown away every time you force push so I can't see what you changed from the last time. This is ok for small stuff but is annoying for larger commits.

The best workflow I have found is what is used on Apache Accumulo:

1. Push new commits as you make changes, **don't force push**
2. If you want to get the latest changes from the original branch (ie main) don't rebase off main as then you'd have to force push. Instead just merge main into this branch periodically and push up the commit. ie "**git merge main**"
3. When it comes time to do the final merge, pick the "**Squash and merge**" option. This will squash everything into 1 commit and add it to the end so we essentially get the same effect as if you had force pushed and re-based but the history doesn't always get blown away. When you pick this option it allows you to modify the commit message so you can change it before the final merge.

cshannon (approved):
LGTM

PR #969, opened 2023-02-08
[AMQ-9219] 5.17.x transitional module for activemq-client-jakarta

mattrpav:


cshannon:
So looking at this and thinking about it more I am seeing several problems including the fact that I realized this doesn't actually solve the problem for Spring boot compatibility.

1. This doesn't really make sense to introduce a new module like this into 5.17.x for a point/bug fix release as 5.17.x has already been out a while. Furthermore this only copies JMS 1.1 classes so is completely broken for most of the methods. This should be targeted for a future version.

2. In main/5.18.x there has been more JMS 2.0 work done so we should be using that as the baseline for this because it handles things properly like working between vendors. But there's still more work to be done because not everything is implemented. Any of the API methods that can be implemented now probably should be. I think most things besides shared subs can probably be done.

3. Anyone that uses the Jakarta API could reasonably assume that calling any method should work. Since there is no previous Jakarta messaging version before 3.0 any code that is Jakarta compatible could assume the spec should work so we would need to finish the implementation or at least be very clear in the documentation what won't work.

4. This doesn't actually fix the issue for Spring Boot 3 Spring Boot won't be able to add back in support because Spring boot is going to pull in the server itself to allow running the 5.x broker and the broker doesn't use the new Jakarta API. A new Jakarta client would only work if using just the client in the JVM and not the broker itself. For Spring 6.x I'm not sure, it may be good enough if it's just client support but if Spring 6 is calling off to any new methods that are not implemented in their messaging module (which they reasonably can do as I pointed out in number 3) then this breaks and won't work.

activemq-client-jakarta/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
```diff
@@ -0,0 +1,808 @@
...
+    }
+
+    @Override
+    public long getJMSDeliveryTime() throws JMSException {
+        throw new UnsupportedOperationException("getJMSDeliveryTime() is unsupported in transitional client");
+    }
+
+    @Override
+    public void setJMSDeliveryTime(long deliveryTime) throws JMSException {
+        throw new UnsupportedOperationException("setJMSDeliveryTime(long) is unsupported in transitional client");
```
cshannon, line 796: This has been implemented in Main, this PR should be updated to include the latest changes from Main since this will be targeted for 5.18.0 now

cshannon (changes requested):
I think this should go into 5.18.0 and should be updated to include all the latest JMS 2.0 changes that were done that are in main. The changes are here: https://github.com/apache/activemq/pull/682

cshannon:
I just noticed https://github.com/apache/activemq/pull/968/files so maybe this PR just gets closed and we go with that one instead if not using this for 5.17.x anymore

PR #979, opened 2023-02-15
AMQ-7309 - Implement Message#isBodyAssignableTo and Message#getBody methods

cshannon:
This might still need some tweaks but there's some initial tests pushed and wanted to get eyes on it and any feedback.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
@@ -956,4 +956,22 @@ public class ActiveMQBytesMessage extends ActiveMQMessage implements BytesMessag
             }
         }
     }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean isBodyAssignableTo(Class c) {
+        return getContent() == null || c.isAssignableFrom(byte[].class);
+    }
+
+    @SuppressWarnings("unchecked")
+    protected <T> T doGetBody(Class<T> asType) {
+        //Make sure the bytes are stored before trying to copy and return
+        if (dataOut != null && getContent() == null) {
+            storeContent();
+        }
+
+        final ByteSequence content = getContent();
+        return content != null ? (T) new ByteSequence(content.getData(), content.getOffset(),
+            content.getLength()).getData() : null;
+    }
 }
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
@@ -971,6 +971,6 @@ public class ActiveMQBytesMessage extends ActiveMQMessage implements BytesMessag
         }
 
         final ByteSequence content = getContent();
-        return content != null ? (T) Arrays.copyOf(content.getData(), content.getLength()): null;
+        return content != null ? (T) content.toArray() : null;
     }
 }
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
@@ -971,7 +971,6 @@ public class ActiveMQBytesMessage extends ActiveMQMessage implements BytesMessag
         }
 
         final ByteSequence content = getContent();
-        return content != null ? (T) new ByteSequence(content.getData(), content.getOffset(),
-            content.getLength()).getData() : null;
+        return content != null ? (T) Arrays.copyOf(content.getData(), content.getLength()): null;
     }
 }
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMapMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMapMessage.java
@@ -856,9 +856,13 @@ public class ActiveMQMapMessage extends ActiveMQMessage implements MapMessage {
         final ByteSequence content = getContent();
         final Map<String, Object> map = content != null ? deserialize(content) : null;
 
-        if (map != null) {
+        //This implementation treats an empty map as not having a body so if empty
+        //we should return null as well
+        if (map != null && !map.isEmpty()) {
             map.replaceAll((k, v) -> v instanceof UTF8Buffer ? v.toString() : v);
+            return (T) map;
+        } else {
+            return null;
         }
-        return (T) map;
     }
 }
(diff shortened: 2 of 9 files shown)


cshannon:
I think this is ready for review now, I updated the getBody() method to copy into a new object. Can let the CI tests run to make sure I didn't break anything as I had to do a little refactoring to make copying the body simpler.

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
```diff
@@ -956,4 +956,22 @@ protected void doCompress() throws IOException {
...
+    @SuppressWarnings("unchecked")
+    protected <T> T doGetBody(Class<T> asType) {
+        //Make sure the bytes are stored before trying to copy and return
+        if (dataOut != null && getContent() == null) {
+            storeContent();
+        }
+
+        final ByteSequence content = getContent();
+        return content != null ? (T) new ByteSequence(content.getData(), content.getOffset(),
+            content.getLength()).getData() : null;
```
tabish121, line 975: This isn't a deep copy from what I see in the ByteSequence code, just a shallow wrapping of the array, plus it creates a new ByteSequence that would never actually be used so needless GC.  I'd just do an Arrays.copyOfRange on the data in the original since you want t deep copy here to avoid the caller modifying the bytes in the message content as you want to get the same payload on a rollback / recover etc.

cshannon: Thanks, good catch. I just copied this from the parent Message class in its copy() method:

https://github.com/apache/activemq/blob/ae4e305f850dd35d92f8e49758309f21609d3206/activemq-client/src/main/java/org/apache/activemq/command/Message.java#L184-L189

This makes me wonder if this needs to be fixed there at some point as well. I've sometimes seen weird behavior with the VM transport that appeared to be from shared messages between server/client even though copying should be happen so maybe this is a culprit.

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQTextMessage.java
```diff
@@ -210,4 +210,22 @@ public String toString() {
...
+         */
+        if (getText() == null) {
+            return true;
+        }
+        return c.isAssignableFrom(java.lang.String.class);
+    }
+
+    @SuppressWarnings("unchecked")
+    protected <T> T doGetBody(Class<T> asType) throws JMSException {
+        return getText() != null ? (T) getText() : null;
```
gemmellr, line 229: Is the simpler "return (T) getText();" not equivalent?

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
gemmellr: (also if there is somehow any offset/length diff then the returned bytes would actually be invalid as they dont account for those)

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
```diff
@@ -795,14 +795,22 @@ public void setJMSDeliveryTime(long deliveryTime) throws JMSException {
...
+            return doGetBody(asType);
+        }
+
+        throw new MessageFormatException("Message body cannot be read as type: " + asType);
+    }
 
 	@Override
 	public boolean isBodyAssignableTo(Class c) throws JMSException {
-        throw new UnsupportedOperationException("isBodyAssignableTo(Class) is not supported");
+        return true;
```
gemmellr, line 809: indentation is off (on the method itself...and some others above)

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
tabish121: It's going to depend on how that is really used, a shallow copy may be sufficient for the JMS 1.1 APIs where you didn't get direct access as you are getting in this simplified API method where you can accidentally hand out the byte[] payload reference.

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMapMessage.java
```diff
@@ -827,4 +839,26 @@ public Map<String, Object> getContentMap() throws JMSException {
         initializeReading();
         return map;
     }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean isBodyAssignableTo(Class c) throws JMSException {
+        final Map<String, Object> map = getContentMap();
+        if (map == null || map.isEmpty()) {
```
gemmellr, line 847: doGetBody doesnt seem to special case the empty map as this does...should it? (since it special casing saying true while being given any class here)

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQTextMessage.java
```diff
@@ -210,4 +210,22 @@ public String toString() {
...
         return super.toString();
     }
+
+    @SuppressWarnings("unchecked")
+    public boolean isBodyAssignableTo(Class c) throws JMSException {
+        /*
+         *If the message is a TextMessage, ObjectMessage, MapMessage or BytesMessage
+         * and the message has no body, then the above does not apply and this method
+         *  will return true irrespective of the value of this parameter.
+         */
```
gemmellr, line 220: Could probably make the comment make more sense by not using the full spec sentence :)

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQBytesMessage.java
mattrpav: Is using the readBytes() an option here?

```
    @Override
    public <T> T getBody(Class<T> c) throws JMSException {
        if(c == null || getContent() == null) { return null; }

        byte[] tmpBytes = new byte[(int)getBodyLength()];
        readBytes(tmpBytes);
        return (T)tmpBytes;
    }
```

cshannon: I'm going to add a copy method to ByteSequence, that seems simplest.

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMapMessage.java
cshannon: Good catch, it likely should because the map message impl treats an empty map as basically null (no body) so if it's empty we should return null I think and not an empty map.

PR #981, opened 2023-02-19
[AMQ-9223] Upgrade to slf4j 2.0.6

jbonofre:


cshannon (changes requested):
@jbonofre - We should wait on this because log4j2 has a transitive dependency bug that was just recently resolved: https://issues.apache.org/jira/browse/LOG4J2-3601 but not until version 2.20.0 so I don't think this should go into 5.18.0

jbonofre:
@cshannon yes, it's just to prepare. We can include in 5.19.0.

gemmellr:
> We should wait on this because log4j2 has a transitive dependency bug that was just recently resolved: https://issues.apache.org/jira/browse/LOG4J2-3601 but not until version 2.20.0 so I don't think this should go into 5.18.0

Indeed it was https://lists.apache.org/thread/t8of737y6lvf8gfwyb1who3xlcf01j6v :)

Given that issue, this PR is likely disabling logging in various places currently due to still using 2.19.0. and not adjusting the dependencies as a workaround.  That bug and OSGI stuff is what stopped me trying to use SLF4J2 on Artemis back when we switched it to using SLF4J+Log4J2. I know there have been OSGI related changes to SLF4J2 since then, though I do still see recent issues+PRs open around that suggesting there may still be things to address in that space.

That said, I also didnt see much huge rush to update back then, since applications can just depend on slfj-api 2 themselves if desired, and it will then be chosen during dependency resolution over whatever version activemq used anyway, and work so long as the client isnt using the new fluent APIs.

I see 2.20.0 has just hit Maven Central in the last few hours though so a complete PR could now actually be prepared here.

jbonofre:
log4j 2.20.0 is now available and we have a PR about that: https://github.com/apache/activemq/pull/985

I proceeed and merge the 2 PRs.

PR #982, opened 2023-02-20
WIP: [AMQ-8320] Add JMS 2.0 JMSProducer deliveryDelay support

mattrpav:
1. DeliveryDelay producer value set to ActiveMQ header for scheduler
2. DeliveryTime value stored as message header on produce
3. DeliveryTime header removed on consume

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
@@ -629,6 +629,9 @@ public class ActiveMQMessageConsumer implements MessageAvailableConsumer, StatsC
                 }
             });
         }
+        if (m.propertyExists(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY)) {
+            m.setJMSDeliveryTime(m.getLongProperty(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY));
+        }
         return m;
     }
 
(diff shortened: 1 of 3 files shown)


cshannon:
I assume this is targeted for 5.19.0 at this point? It would be good to cut off 5.18.0 so we can release it soon.

mattrpav:
@cshannon No rush to get it into 5.18.0. It was a pretty small change, so I thought I'd put it out here for discussion at least.

gemmellr:
Since message objects can be reused/resent, and might already have an existing value for JMSDeliveryTime, it should always be setting the JMSDeliveryTime value on send even if there isnt a delay, to ensure the JMSDeliveryTime value on the message after send reflects behaviour of that send and not some previous send or receive.

That doesnt mean it has to transit a delivery time on the wire...ideally it wouldnt if there is no delay being requested for this send...just that after send the local message object should report the delivery time for that send. Elsewhere I added an internal message method to set the value and note whether it is for transmission.

It doesnt feel like it should be setting a message property at all, only a header. I don't expect the TCK tests which check additional properties arent being added to the messages are using delivery-delay at the time..but they could. I know the older 'message property extension' scheduling stuff uses one, but its not clear to me this needs to. They could just be treated as separate mechanisms, which they kind of are already as you have to choose to respect the new value on the producer or the old one on the message. Theres no reason the broker couldnt do the same, and this new mechanism essentially ignore that old prop.

jbonofre:
I added milestone on GH to indicate the target version for PR. This one is targeted for 6.1.0 (updated the Jira too).

PR #983, opened 2023-02-21
[AMQ-9225] Remove j2ee management dependency as it is no longer part …

mattrpav:
…of JEE spec

Approach:

Implement self-hosted interfaces to replace the J2EE management statistics interfaces


cshannon:
This generally seems fine to me to remove as long as tests pass and things like OSGi don't break, etc. There's no real reason we need to have it for the broker internally and if user still needs the jar they can easily add it themselves.

cshannon (commented):
This generally seems fine to me to remove as long as tests pass and things like OSGi don't break, etc. There's no real reason we need to have it for the broker internally and if user still needs the jar they can easily add it themselves.

PR #989, opened 2023-03-14
[AMQ-9161] Mark JDBC JournalPersistenceAdapter as deprecated for removal

mattrpav:
This replaces #931


cshannon (commented):
Marking as deprecated seems fine but it might be a good idea to add some comments as to why it's being removed, replacement in the future etc in case someone updates and their build flags it.

mattrpav:
> Marking as deprecated seems fine but it might be a good idea to add some comments as to why it's being removed, replacement in the future etc in case someone updates and their build flags it.

@cshannon comments added.

jbonofre (approved):

cshannon (approved):

cshannon:
This commit broke the ActiveMQ schema parsing with the extra javadoc comment blocks, fixed in #990

PR #995, opened 2023-03-30
[AMQ-9237] Add META-INF/services to activemq-client-jakarta

mattrpav:


cshannon (changes requested):
This shows that we really need tests. We likely need to use Maven here to and do something where we fork another JVM. We should really run the client against a broker to verify it can connect and work so I think that needs to be included as part of this. We need to verify all the transports work in a real test that isn't ignored.

cshannon:
Alright, can go ahead and get this merged so we can do 5.18.1 as need to get this fixed but for 5.19.0, especially with more planned changes we definitely can't release without automated testing in some way.

cshannon (approved):

PR #997, opened 2023-04-11
AMQ-9242 - remove compile time dependency log4j-slf4j2-impl dependency from activemq-partition

cshannon:


cshannon:
I meant to get this into 5.18.1 but forgot but it's fine to wait for 5.18.2 as it's been broken a while, not a new thing and this module isn't really used anymore anyways.

mattrpav (approved):
LGTM (pending CI go-green). Agree, this is more of a POC module than a widely used component.

cshannon:
Looked like the run on CI was hung for some reason but seems unrelated to the changes here so I'm going to merge it. I can investigate if it hangs again but seems very unlikely it has anything to do with just changing this compile time dependency.

PR #998, opened 2023-04-12
AMQ-9243 - Remove deprecated jetty-continuation dependency

cshannon:
This commit removes the dependency on jetty-continuation and updates the MessageServlet and AjaxListener servlets to use the Servlet Async api directly for async requests through a new Async holder object that is used to track/manage the request.

As I mentioned in the Jira, this is also fine to backport to 5.18.x as there's no public facing API changes, it's just an internal detail. Jetty was already using the Async api and we are doing the same, just more directly now.

The tests in the activemq-web-demo can be run to test out and demonstrate the new changes. Running the [AjaxTest](https://github.com/cshannon/activemq/blob/6a979685c6a7d5cb6a091cb85e42bd03093adfa4/activemq-web-demo/src/test/java/org/apache/activemq/web/AjaxTest.java) in that project shows the AjaxListener still works with async requests and I also added a couple more tests to [RestTest](https://github.com/cshannon/activemq/blob/6a979685c6a7d5cb6a091cb85e42bd03093adfa4/activemq-web-demo/src/test/java/org/apache/activemq/web/RestTest.java) in this PR to demonstrate that the Async request/resume is working with the changes.

### DIFF
--- a/activemq-web-console/pom.xml
+++ b/activemq-web-console/pom.xml
@@ -211,10 +211,6 @@
               <groupId>org.eclipse.jetty.websocket</groupId>
               <artifactId>websocket-server</artifactId>
           </exclusion>
-          <exclusion>
-              <groupId>org.eclipse.jetty</groupId>
-              <artifactId>jetty-continuation</artifactId>
-          </exclusion>
       </exclusions>
     </dependency>
     <dependency>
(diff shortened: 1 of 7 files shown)


mattrpav (approved):
LGTM

PR #1001, opened 2023-04-18
Adding CodeQL workflow on PRs and on a schedule

coheigea:
Using OpenSSF scorecard, ActiveMQ currently gets 0 for SAST:

| 0 / 10  | SAST                   | SAST tool is not run on all    | https://github.com/ossf/scorecard/blob/27cfe92ed356fdb5a398c919ad480817ea907808/docs/checks.md#sast                   |
|         |                        | commits -- score normalized to |                                                                                                                       |
|         |                        | 0                              |                                                                                                                       |

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,94 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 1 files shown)


coheigea:
FAO @jbonofre

coheigea:
@mattrpav Not that I'm aware of, I don't see anything here at any rate https://cwiki.apache.org/confluence/display/INFRA/Git+-+.asf.yaml+features#Git.asf.yamlfeatures-GitHubsettings
I have generally migrated over to github actions as I find it's simpler and we can control everything via yaml in our codebases rather than relying on ASF infra.

jbonofre:
I propose to enable this when fully moving to GitHub Action.

PR #996, opened 2023-03-30
[AMQ-9239] Jakarta JMS broker

mattrpav:
Reviewer Notes:

 - [x] JDK 17 required for Spring 6
 - [x] Spring dependency update needed to pull in spring-jms jakarta.jms namespace
 - [x] qpid dependency update needed to pull in jakarta.jms namespace
 - [x] Rebase with to pickup jetty-continuation to servlet async refactor
 - [x] Convert to jakarta.annotation
 - [x] Convert to jakarta.jms
 - [x] Convert to jakarta.resource
 - [x] Convert to jakarta.servlet v5 (Jakarta EE 9)
 - [x] Convert to jakarta.transaction
 - [x] Jetty 11+ requires sub-protocol validation for web sockets [Jetty #3709](https://github.com/eclipse/jetty.project/issues/3709)
 - [x] Rebase & convert joram-jms-tests to ActiveMQ hosted
 - [x] Jolokia is disabled, as it does not currently support jakarta
 - [x] activemq-partition and zookeeper test dep removed (AMQ-9259)
 - [x] Update web console to use jakarta jsp template urn
 - [x] Address FIXMEs in *-web projects
 - [x] Upgrade stompjms-client (v3.1.0 supports jakarta.jms.*)
 - [x] Update assembly examples
 - [x] Temporarily disable Camel (no Jakarta / Spring 6 support until v4)
 - [x] Fix activemq-http 60s stopTimeout temporary work-around
 - [x] Fix activemq-ra/ MessageEndpointProxyTest.java and ServerSessionImplTest.java (rollback jmock upgrade)
 - [x] Fix activemq-spring/AMQDeadlockTest3 (failing)
 - [x] Fix activemq-unit-tests/BrokerServiceTest.testLargeFileSystem (byte-buddy opens)
 - [x] Fix network/CustomBridgeFactoryTest (mockito / byte-buddy opens) 
 - [x] Fix store.jdbc/LeaseDatabaseLockerTest, JDBCIOExceptionHandlerMockeryTest
 - [x] Refactor activemq-spring/ParallelXATransactionTest (jencks lib is EOL and javax.jms) 
 - [x] Web Console - queue delete, pause resume throws exception
 - [x] Run simulated -Prelease,deploy to exercise maven plugins (esp javadoc)
 - [x] Updated activemq-client-jakarta with Maven relocation info back to activemq-client 

TODO:
 - [ ] Review jetty.xml pathSpec change

Tests passing:
[...]

PR #1002, opened 2023-04-24
[AMQ-9239] Host joram-jms-unit tests ahead of converting to jakarta.jms

mattrpav:
Currently this module is unmaintained and we need to convert it for jakarta broker support.

Modules that use this module:
activemq-amqp
activemq-unit-tests

ref: https://github.com/fusesource/joram-jms-tests

### DIFF
--- a/activemq-tooling/activemq-joram-jms-tests/src/main/java/org/objectweb/jtests/jms/conform/connection/TopicConnectionTest.java
+++ b/activemq-tooling/activemq-joram-jms-tests/src/main/java/org/objectweb/jtests/jms/conform/connection/TopicConnectionTest.java
@@ -1,0 +1,171 @@
+/*
+ * Copyright 2009 Red Hat, Inc.
+ * Red Hat licenses this file to you under the Apache License, version
+ * 2.0 (the "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+ * implied.  See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+package org.objectweb.jtests.jms.conform.connection;
+
+import javax.jms.JMSException;
+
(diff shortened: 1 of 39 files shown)


gemmellr:
Personally I dont see enough value in these tests to copy+convert them, a lot of it is just testing basic stuff the rest of the tests should be covering themselves. I would simply drop their usage (and did exactly that elsewhere). They dont cover any of the newer stuff, not having been updated in over a decade. I cant remember a single time theyve actually shown a failure. The actual official spec tests are easily available now unlike back then if needed.

mattrpav:
@gemmellr I think removing it would be a good end goal. However, right now its deeply integrated into amqp tests. I'll make a JIRA for removing joram and refactoring the amqp tests post-jakarta work.

gemmellr:
Knowing how the Joram tests work, 'deeply integrated' seems a bit of an exageration. Theres one package of classes that could just be outright deleted, and it looks like there are a couple of other tests misusing some related code to enable broker protocol tracing that arguably shouldnt be enabled anyway and can be got other ways. Seems trivial to remove.

mattrpav:
@gemmellr Yep, I've got AMQ-9249 tee'd up for after getting the Jakarta broker work in order. 

If you've got spare cycles now, go for it! I'm happy to rebase from there, if you can get it merged in before the Jakarta work is done.

Thanks

gemmellr:
Just giving some feedback on a simpler approach, its not something I will be working on.

mattrpav:
@gemmellr no worries, appreciated!

activemq-tooling/activemq-joram-jms-tests/src/main/java/org/objectweb/jtests/jms/conform/connection/TopicConnectionTest.java
```diff
@@ -0,0 +1,171 @@
+/*
+ * Copyright 2009 Red Hat, Inc.
```
cshannon, line 2: I noticed there is a Red Hat reference in some files like here, I assume we need to get rid of this?

mattrpav: Already corrected in the jakarta PR commit:

https://github.com/apache/activemq/pull/996/commits/17d7f024873765ba05ac735f53dc321117dc8881

PR #1004, opened 2023-05-02
[#9254] DataFile readRecord fallback to OS file.length in rare edge case

mattrpav:


mattrpav:
related: https://github.com/apache/activemq/commit/8c3ef6cadb46d9694c68aa649a7952eb1612279f

cshannon (commented):
This change looks good to me, I think checking the real file length here only if that check fails makes sense as it should be a rare edge case so we avoid the file system read.

The main thing of course is testing. I know you were working on creating an integration type test with a small KahaDB that could be checked in but a better thing might just be to do some real unit testing.

You really just need to verify that the logic works correctly so we an just add a unit test for DataFileAccessor, we don't need a real broker. I would just use mockito and mock out what you need. I'll work on a quick test and can create a PR to your branch.

cshannon:
@mattrpav - let me know what you think of my commit I just pushed. We should be able to just do a normal merge commit when we merge the PR and keep both commits separate in this case instead of squashing.

PR #1009, opened 2023-05-22
[AMQ-9259] Remove activemq-partition and zookeeper test dependency

mattrpav:


cshannon (approved):
LTM, very much long overdue to remove this stuff so thanks for doing it.

PR #1013, opened 2023-06-06
AMQ-9692 - Fix network subscriptions for composite consumers

cshannon:
This fixes network subscriptions that are generated on demand when a consumer uses composite destinations. Before this fix conduit subscriptions didn't work correctly. This fix now splits up the composite dest and generates correct demand for each of the individual destinations.


cshannon:
closing, wrong jira number, will reopen

PR #1014, opened 2023-06-06
AMQ-9262 - Fix network subscriptions for composite consumers

cshannon:
This fixes network subscriptions that are generated on demand when a consumer uses composite destinations. Before this fix conduit subscriptions didn't work correctly. This fix now splits up the composite dest and generates correct demand for each of the individual destinations.


cshannon:
I will let CI run but plan to merge this in tomorrow if everything looks good.

cshannon:
@mattrpav - feel free to take a look (or not), it's not a very common issue many people will run into it. I have tested things pretty extensively already before pushing this patch.

PR #1035, opened 2023-06-21
[AMQ-9244] Add JWT authentication plugin

jbonofre:


activemq-broker/pom.xml
```diff
@@ -66,6 +66,18 @@
       <artifactId>activemq-jaas</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>com.nimbusds</groupId>
+      <artifactId>nimbus-jose-jwt</artifactId>
+      <version>9.22</version>
```
jbertram, line 72: The latest version is `9.31`. How come an older version is being used instead of the latest?

activemq-broker/pom.xml
```diff
@@ -66,6 +66,18 @@
...
+      <optional>true</optional>
+    </dependency>
+    <dependency>
+      <groupId>org.bitbucket.b_c</groupId>
+      <artifactId>jose4j</artifactId>
+      <version>0.7.9</version>
```
jbertram, line 78: The latest version is `0.9.3`. How come an older version is being used instead of the latest?

activemq-broker/src/main/java/org/apache/activemq/security/jwt/Claims.java
```diff
@@ -0,0 +1,104 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.security.jwt;
+
+import com.fasterxml.jackson.annotation.JsonValue;
+
+import java.util.Set;
+
+public enum Claims {
```
jbertram, line 23: What exactly are all the values, and why are so many unused values needed?

activemq-broker/src/main/java/org/apache/activemq/security/jwt/JwtAuthenticationPlugin.java
```diff
@@ -0,0 +1,60 @@
...
+ * @org.apache.xbean.XBean element="jwtAuthenticationPlugin"
+ *                         description="Provides a JWT authentication plugin"
+ *
+ *
+ * This plugin is rather simple and is only meant to be used as a first step. In real world applications, we would need
+ * to being able to specify different key format (RSA, DSA, EC, etc), the issuer, the claim to use for groups and maybe
+ * some mapping functionalities.
+ * The header name is also hard coded for simplicity.
+ */
+public class JwtAuthenticationPlugin implements BrokerPlugin {
```
jbertram, line 35: Perhaps I'm wrong, but this doesn't appear usable as is. How can users configure these values? Is the idea here that they will need implement and deploy their own version of this class in order to configure auth as needed? If so, are there plans to make this configurable in the future?

activemq-broker/src/main/java/org/apache/activemq/security/jwt/TokenUtils.java
```diff
@@ -0,0 +1,157 @@
...
+import java.security.PublicKey;
+import java.security.spec.PKCS8EncodedKeySpec;
+import java.security.spec.X509EncodedKeySpec;
+import java.util.Base64;
+import java.util.List;
+
+/**
+ * Utilities for generating a JWT for testing
+ */
+public class TokenUtils {
```
jbertram, line 41: Nothing from this class is used anywhere. The JavaDoc says it's used for testing, but there are no tests. This class seems unnecessary at this point.

activemq-broker/src/main/java/org/apache/activemq/security/jwt/JwtAuthenticationBroker.java
```diff
@@ -0,0 +1,179 @@
...
+                        .setVerificationKey(parsePCKS8(jwtValidatingPublicKey))
+                        .setJwsAlgorithmConstraints(
+                                new AlgorithmConstraints(AlgorithmConstraints.ConstraintType.WHITELIST,
+                                        AlgorithmIdentifiers.RSA_USING_SHA256,
+                                        AlgorithmIdentifiers.RSA_USING_SHA384,
+                                        AlgorithmIdentifiers.RSA_USING_SHA512)
+                        );
+
+                final JwtConsumer jwtConsumer = builder.build();
+                final JwtContext jwtContext = jwtConsumer.process(username);
```
jbertram, line 109: On [the Jira](https://issues.apache.org/jira/browse/AMQ-9244?focusedCommentId=17723012&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17723012) it was indicated that the _password_ field should contain the token. If `username` is used instead will this fit the reported use-case?

jbertram:
Generally speaking this looks more like a proof-of-concept (i.e. based on [this blog post](https://www.tomitribe.com/blog/jwt-authentication-and-authorization-with-apache-activemq/)) rather than a feature which is ready to use. I'm not sure it makes sense to merge it at this point especially with no tests to validate the functionality and to mitigate future regressions.

jbonofre:
That's the starting point. The intention is not to merge right now. I wanted to share here as few users want to try. I will work on this PR after the releases plan.

jbertram:
@jbonofre, understood. That wasn't clear from [your comment](https://issues.apache.org/jira/browse/AMQ-9244?focusedCommentId=17723067&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-17723067) on the Jira where you said,

> It doesn't break anything, just add a new plugin, so I don't see problem to merge.

Thanks for the clarification.

cshannon:
I went ahead and converted this to a draft PR. Generally speaking we should be using draft PRs if something is not ready to merge. This way it makes it clear it's still a work in progress. Marking things like this draft PRs in the future should hopefully help with the confusion about the state.

jbonofre:
I'm starting to refactore this one to provide a login module supporting oauth2 flow.

PR #1036, opened 2023-06-21
[AMQ-8149] Add tooling to create activemq docker image (including multi-platform support)

jbonofre:


assembly/src/docker/docker-compose.yml
```diff
@@ -0,0 +1,39 @@
...
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+# limitations under the License.
+################################################################################
+
+version: "2.1"
+services:
+  activemq:
+    image: ${ACTIVEMQ_DOCKER_IMAGE_NAME:-qctivemq}
```
mattrpav, line 22: Looks like minor type-o here.  Should be 'activemq' instead of 'qctivemq'?

jbonofre:
@mattrpav good catch, I will fix that. Thanks !

PR #1040, opened 2023-08-01
AMQ-9293 - Upgrade to Spring 5.3.29

jbonofre:


mattrpav:
@jbonofre please note that 'main' is now Jakarta, JDK 17, and Spring 6 based.

PR #1045, opened 2023-08-30
WIP: [AMQ-8324] JMS 2.0 Producer CompletionListener methods

mattrpav:


mattrpav:
@tabish121 Yep, agreed. This is PR is still WIP. Just starting to flush out the methods and figure out the best place to house the signaling needed b/w the Context and the CompletionListener. 

ActiveMQ already has the AsyncCallback wired all the way down to the transport send w/ async support, so I think it can mostly piggy back on that approach in terms of where and when to invoke the callback.

The heavy lifting will come here shortly to cover the thread awareness and other fun checks as you mentioned.

kenliao94:
@mattrpav I have a draft PR that did exactly what you suggested "mostly piggy back on that approach in terms of where and when to invoke the callback" I will raise it and get back to this thread with link to the pull request.

kenliao94:
^ https://github.com/apache/activemq/pull/1303

WDYT?

PR #1046, opened 2023-08-30
[AMQ-8464] JMS 2.0 JMSConsumer recevieBody method implementations

mattrpav:


activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
```diff
@@ -675,6 +705,49 @@ public Message receive(long timeout) throws JMSException {
...
+            Message message = createActiveMQMessage(md);
+
+            try {
+                // This throws MessageFormatException if body is not of bodyType
+                T body = message.getBody(bodyType);
+                afterMessageIsConsumed(md, false);
+                return body;
+            } catch (MessageFormatException e) {
+                synchronized (unconsumedMessages.getMutex()) {
+                    unconsumedMessages.enqueueFirst(md);
```
cshannon, line 743: This doesn't make any sense to me. Why would we add a bad message that had an exception back to the unconsumedMessages queue? I haven't looked at this in a long time but I think that would attempt to redispatch again as it's meant for redlivery so we would just error over and over.

This error should just be propagated back up like a nomral JMSException I would think, no reason to catch and re-add it here

cshannon: I would think the method should just simply be this:

```java
<T> T receiveBody(Class<T> bodyType) throws JMSException {
  checkClosed();
  checkMessageListener();

  sendPullCommand(0);
  MessageDispatch md = dequeue(-1);
  if (md == null) {
      return null;
  }

  beforeMessageIsConsumed(md);
  afterMessageIsConsumed(md, false);

  Message message = createActiveMQMessage(md);
  return message.getBody(bodyType);
}
```

cshannon (changes requested):
Edit again:

Apparently I need to read the spec :) It says that it should be redelivered for only some modes so I still think there is some work to be done here.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
```diff
@@ -589,6 +590,35 @@ public Message receive() throws JMSException {
...
+
+        beforeMessageIsConsumed(md);
+        Message message = createActiveMQMessage(md);
+        try {
+            // This throws MessageFormatException if body is not of bodyType
+            T body = message.getBody(bodyType);
+            afterMessageIsConsumed(md, false);
+            return body;
+        } catch (MessageFormatException e) {
+            synchronized (unconsumedMessages.getMutex()) {
```
cshannon, line 615: Ok so based on the spec I think we should only be re-enqueing for certain ack modes: https://jakarta.ee/specifications/messaging/3.0/apidocs/jakarta/jms/jmsconsumer#receiveBody-java.lang.Class-

For example this is qpid jms: 

https://github.com/apache/qpid-jms/blob/149eeb01027120d84af3a62eb83858c86d244b90/qpid-jms-client/src/main/java/org/apache/qpid/jms/JmsMessageConsumer.java#L257-L264

 ```java
 // Should behave as if receiveBody never happened in these modes.
  if (acknowledgementMode == Session.AUTO_ACKNOWLEDGE ||
      acknowledgementMode == Session.DUPS_OK_ACKNOWLEDGE) {


      envelope.setEnqueueFirst(true);
      onInboundMessage(envelope);
      envelope = null;
  }
```

**Review by cshannon — CHANGES_REQUESTED** [review]

File: activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java [hunk 3]
```diff
@@ -589,6 +590,35 @@ public Message receive() throws JMSException {
...
+        checkClosed();
+        checkMessageListener();
+
+        sendPullCommand(0);
+        MessageDispatch md = dequeue(-1);
+        if (md == null) {
+            return null;
+        }
+
+        beforeMessageIsConsumed(md);
```
cshannon, line 607: This line is also seems like a problem as the spec says that the method has to act like the receive never happened in certain modes (auto akck and dups ok) and calling this has side effects such as adding to deliveredMessages and updating next delivery id

cshannon:
Overall my impressions are that this seems more complicated than it appears to be at first glance because you need to properly handle rolling things back and making sure that in certain modes the client acts like the receive was never called. I am not sure that simply re-adding the message to the unconsumed list is enough. There's a lot of code in rollback() to handle things like duplicate detection so the message can be received again.

So that being said I think this needs more work and a lot more testing and might need to wait for 6.1.0

PR #1048, opened 2023-09-07
[AMQ-9305] Adapt the docker-related files to Java 17

essobedo:
Fixes https://issues.apache.org/jira/browse/AMQ-9305

## Motivation

The Dockerfile and corresponding README need to be adapted to Java 17 in order to be able to launch a snapshot of ActiveMQ 5.19 inside a Docker container.

## Modifications

* Use `eclipse-temurin:17-jre` as new root image
* Modify the documentation to refer to the new root image used

### DIFF
--- a/assembly/src/docker/build.sh
+++ b/assembly/src/docker/build.sh
@@ -27,7 +27,7 @@ Usage:
 
   If the --image-name flag is not used the built image name will be 'activemq'.
   Check the supported build platforms; you can verify with this command: docker buildx ls
-  The supported platforms (OS/Arch) depend on the build's base image, in this case [eclipse-temurin:11-jre](https://hub.docker.com/_/eclipse-temurin).
+  The supported platforms (OS/Arch) depend on the build's base image, in this case [eclipse-temurin:17-jre](https://hub.docker.com/_/eclipse-temurin).
 
 HERE
   exit 1
--- a/assembly/src/docker/Dockerfile
+++ b/assembly/src/docker/Dockerfile
@@ -16,7 +16,7 @@
 # limitations under the License.
 ################################################################################
 
-FROM eclipse-temurin:11-jre
+FROM eclipse-temurin:17-jre
 
 # ActiveMQ environment variables
 ENV ACTIVEMQ_INSTALL_PATH /opt
--- a/assembly/src/docker/README.md
+++ b/assembly/src/docker/README.md
@@ -32,7 +32,7 @@ On macOS, an easy way to install `buildx` is to install [Docker Desktop Edge](ht
 
 ## Build
 
-Images are based on the Docker official [Eclipse Temurin 11 JRE](https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=11-jre) image. If you want to
+Images are based on the Docker official [Eclipse Temurin 17 JRE](https://hub.docker.com/_/eclipse-temurin/tags?page=1&name=17-jre) image. If you want to
 build the ActiveMQ image you have the following choices:
 
 1. Create the docker image from a local distribution package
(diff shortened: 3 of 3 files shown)


jbonofre:
ActiveMQ 5.18.x is guaranteed on Java 11. That's why it's temurin 11 by default. I would keep 11 for now and upgrade for 5.19.x.

jbonofre:
`main` is the future 5.19.0 but the Dockerfile is used to create also 5.18.x images (you can see the `build.sh` can download different ActiveMQ version using `--from-release` option).
So I would:
1. either wait 5.19.0 out
2. either limit the release version to 5.19.x in the `build.sh`

mattrpav:
How about we introduce a Dockerfile for each "LTS", that can then have the JDK level set? The JDK revisions are going to continue to rapidly iterate, so having an approach to support multiple will keep things simple to manage.

Dockerfile-5.18.x
Dockerfile-5.19.x
..

Thoughts?

essobedo:
OK I better understand let me try to improve it a bit more

jbonofre:
IMHO, the best approach is to have a specific Dockerfile per branch and scope release-version option to only the major version of the branch.
It's what I planned to do after I added the Dockerfile.

E.g. it means:
- `activemq-5.18.x` should have a Dockerfile scoped for 5.18.x (jdk11)
- `main` (5.19.x) should have a Dockerfile scoped for 5.19.x (jdk17)

essobedo:
So your idea is to have a specific version of the folder assembly/src/docker in each LTS branch?

jbonofre:
Yup

essobedo:
I added a commit to show how it could be managed with a new parameter to define the target java version. The idea is the following:
* If the AMQ version is set then if it is a version before 5.19, it uses Java 11 other it uses Java 17
* Otherwise it is by default Java 17 unless it has been provided explicitly with `--target-java-version <version>`

WDYT?
If you prefer to have a special copy per branch, I will revert this commit and we can just wait for 5.19 to be released, let me know what you prefer.

essobedo:
Example of output:

```
./build.sh --from-release --activemq-version 2.18.2  
Downloading apache-activemq-2.18.2-bin.tar.gz from https://dlcdn.apache.org/activemq/2.18.2/
Using 11 as target Java version
```

```
./build.sh --from-local-dist                                    
Using ActiveMQ dist: ../../target/apache-activemq-*.tar.gz
Using 17 as target Java version
```

```
 ./build.sh --from-local-dist --target-java-version 11
Using ActiveMQ dist: ../../target/apache-activemq-*.tar.gz
Using 11 as target Java version
```

jbonofre:
Thanks I gonna take a look. But I think we can do that in a easy way: as the Java version is specific to a branch, we can have specified in the dockerfile.

essobedo:
Just note that even if the folder is duplicated in each branch, if in branch 5.19, I execute `./build.sh --from-release --activemq-version 2.18.2`, the Java version will be 17 while the expected version is still 11.
In other words, the flexibility of the script can have side effects if the target Java version is not fully managed.

essobedo:
Ok, I reverted the proposal, let's wait for 5.19

mattrpav:
Yeah, image names change too much. I don't think we can rely on a a simple '11', '17' parameter.

Perhaps we should move the full container image name up to the parent pom right next to JDK version, so we can remember to update them at the same time.

jbonofre:
To summarize:
1. I will merge this PR on `main` branch as the target is 5.19.x for JDK 17
2. I will do another change to update the `--release-version` option to square the version to the branch version
3. I keep JDK 11 (temurin) on `activemq-5.18.x` with the same change for `--release-version`option

assembly/src/docker/build.sh
```diff
@@ -100,7 +100,7 @@ elif [ -n "${FROM_LOCAL}" ]; then
   if [ -n "${ARCHIVE}" ]; then
      DIST_DIR=${ARCHIVE}
   else
-     DIST_DIR="target/apache-activemq-*.tar.gz"
+     DIST_DIR="../../target/apache-activemq-*.tar.gz"
```
jbonofre, line 103: Any reason to change the path here ?

jbonofre (changes requested):

assembly/src/docker/build.sh
essobedo: Good question, since I was not sure from where to launch the script because I did not find any doc about it, I may have changed it incorrectly. I will revert it asap.

essobedo: Fixed thx

jbonofre (approved):

PR #1050, opened 2023-09-08
[AMQ-9296] Add support of credentials to the docker image

essobedo:
Fixes https://issues.apache.org/jira/browse/AMQ-9296

## Motivation

Since Docker images for Apache ActiveMQ are published, it would be great to support env vars for USER and PASSWORD.

## Modifications:

* Add an entry point script to generate the configuration files if the environment variable `ACTIVEMQ_USER` is set
* Describe the supported environment variables

### DIFF
--- a/assembly/src/docker/Dockerfile
+++ b/assembly/src/docker/Dockerfile
@@ -28,6 +28,8 @@ ENV PATH $PATH:$ACTIVEMQ_HOME/bin
 # activemq_dist can point to a directory or a tarball on the local system
 ARG activemq_dist=NOT_SET
 
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+
 # Install build dependencies and activemq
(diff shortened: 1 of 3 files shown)


jbonofre (changes requested):
I'm not sure adding an authenication plugin is what's expected.
IMHO, the user (in Jira) is asking for authentication on the webconsole, so updating `jetty-realm.properties`.  The same can apply to `credentials.properties` and `users.properties`. I can explain to you the use of those different files.
So I think, we could have different env variables, depending the credentials we want to set.

essobedo:
I understand the differences with the authentication on the web console, I simply believed that it is what the OP was expecting. I guess the OP needs to clarify a bit more because it is really unclear/misleading

jbonofre:
Agree, the ticket should provide some details. However, I think it would be acceptable to have different env variables, depending what kind of credentials we want to provide.

jbonofre:
Superseded by #1114 1114

PR #1052, opened 2023-09-15
[NO-JIRA-6.0] Update poms to 6.0.0-SNAPSHOT

mattrpav:


jbonofre:
`poms` are not enough IMHO. `Jenkinsfile` should be updated too.

I proposed to do it on the mailing list, please can I take it ?

PR #1057, opened 2023-09-28
WIP: [AMQ-8391] Consolidate JAAS for messaging and web layers

mattrpav:
- [x] Decide activemq-jaas to use java.util.logger vs shade log4j2 into activemq-run (slf4j bootstrapping issue)
 - [ ] Solve for Jetty path spec / .action not working


mattrpav:
@jbonofre @cshannon 

This is what is done:

cshannon:
Can this wait until 6.1.0? The issue has been open for a while and doesn't seem like it would be a big deal if it was done in 6.1.0

I took a quick look but i am not familiar with any of the stuff here and i would need more time than i have this week likely to look into it if you need help fixing it and i know we were trying to get 6.0.0 out soon

cshannon:
And I should add it's still fine to go into 6.0.0 if it can be figured out this week, I just didn't want to hold things up for a long time because of it.

mattrpav:
@cshannon yeah, I think the only things left are minor.

1. Any objections to moving activemq-jaas to jul for logging?
2. There is a syntax issue w/ jetty.xml path spec causing authz to go wonky (I'll be able to get into debugger this week)

cshannon:
The main problem with the bridge is the performance penalty so I wouldn't include it. If you read the documentation there can be a pretty big performance penalty so i don't know that we would want to include it by default, not sure. So probably just make it JUL and let other people decide if they want to add a bridge.

jbonofre:
I agree this kind of change is acceptable in 6.0.0. If we don't include in 6.0.0, then, we will have to wait 7.0.0 because it's a huge major change for the users: the user/password file will completely different, etc.

About logging, I don't think it's a major concern. For instance, I know a lot of users using jetty-jaas/jetty-ldap and they struggled to have log messages in `activemq.log` due to the backend (jetty using JUL). The performance impact is "limited" imho as we talk about log from specific areas. 

I would also test that we can use different realms per scope: one for broker, one for jetty, one for JMX. Most of our users are using different authentication backend (at least between broker and jetty/admin console).

So, if we want to include in 6.0.0, I would:
1. do large tests (manual) with multiple realms, different login modules
2. according to (1), it needs time. I would rather see that change in 7.0.0 and document (on website) how to secure the JMX and admin console (it's already what we do with 5.x).

mattrpav:
Proposal:

1. Include in v6.0.0 (this is a makes-life-easier-for-new users feature)
2. Add example configs for how to separate the realms back out into the dist assembly
3. Convert activemq-jaas to jul
4. Update website w/ logging for activemq-jaas and how to use the breakout-realm configs

jbonofre:
I agree with the proposal. I just add (for me):

5. Extensive tests with different realms (for broker, jetty/adminconsole and JMX) and login modules (LDAP, JDBC, ...), including eventually unit tests

jbonofre:
I already noticed two potential issues to address:
1. web console authentication seems to not work out of the box (I will double check)
2. jolokia authentication seems also broken

I'm fixing that.

jbonofre:
@mattrpav that's my point, we have to document/state that on website/Release Notes. It's a breaking change.

jbonofre:
About Jolokia, I think it would be better to upgrade to Jolokia 2.0.0-M4 which support Jakarta Servlet API. I'm on it.

mattrpav:
Reworked PR to just consolidate web and messaging JAAS for 6.0.0:

https://github.com/apache/activemq/pull/1076

PR #1065, opened 2023-10-06
[AMQ-9324] Convert activemq-jdbc-store to use j.util.Base64

mattrpav:
- No need to drag JAXB dependency just for a base64 converter


jbonofre:
Fully agree, it's definitely a good idea.

jbonofre (approved):

PR #1075, opened 2023-10-18
AMQ-9343 - Reduce memory used for in flight transactions

cshannon:
This commit will reduce the memory required in KahaDB for long running transactions and transactions with a lot of pending message sends by clearing out the message memory when no longer needed instead of keeping it tracked in the pending map


jbonofre (approved):

jbonofre:
The Jenkins execution failed as there's no space left on the executor.
I will test locally on my machine. If you want, you can trigger a new build (creating an empty commit and rebasing for instance).

cshannon:
I can't get CI to run so I will just merge it and we can let CI run whenever it works again against main

PR #1103, opened 2023-10-31
AMQ-7338: Add mariadb JDBC adapter

jbonofre:


jbonofre:
mariadb is LGPL licensed.

PR #1115, opened 2023-11-10
AMQ-9384: Fix "read" securityConstraint and authentication (including logout)

jbonofre:


mattrpav (approved):
LGTM. Tested in Chrome

PR #1117, opened 2023-11-13
AMQ-9388 - Exclude activemq-client-jakarta from camel-activemq

cshannon:
The current version of camel pulls in the activemq-client-jakarta jar which is not necessary as it no longer exists with ActiveMQ 6.0.0

Furthermore the version being pulled in is 5.18.2 which contains a critical CVE that was fixed in 5.18.3

### DIFF
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -165,6 +165,15 @@
     <dependency>
       <groupId>org.apache.camel</groupId>
       <artifactId>camel-activemq</artifactId>
+      <exclusions>
+        <!-- Current version of Camel has a dependency on the removed
+        activemq-client-jakarta module. This exclusion can eventually be removed
+        when we upgrade to a version of camel-activemq that depends on 6.x -->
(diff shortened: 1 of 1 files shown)


mattrpav:
While activemq-client-jakarta is a relocation to 6.0.0, the way the camel-activemq pulls it in, Maven reactor isn't up-leveling it to 6.0.0 for us-- which would have mad this a non-issue. Alas, we'll need to recut the release.

I see activemq BOM's in our future.

% mvn dependency:tree snippet
```
[INFO] +- org.apache.camel:camel-activemq:jar:4.1.0:compile
[INFO] |  \- org.apache.activemq:activemq-client-jakarta:jar:5.18.2:compile
```

mattrpav (approved):
LGTM, good catch.

cshannon:
I can merge this and then we need to revert the 6.0.0 preparation on the main branch and redo the release, I will let @jbonofre handle that.

PR #1120, opened 2023-11-15
Update entrypoint.sh for console authentication

lionelraymundi:


mattrpav:
@lionelraymundi jetty realm is merged into a single JAAS realm in 6.x. This change appears to revert that.

A single JAAS realm simplifies management overall, and we shouldn't need completely separate JAAS realms to do access control across services. 

If you want to separate access for web layer services, simply change the roles in the conf/jetty.xml file and conf/users.properties.

lionelraymundi:
Thanks for your quick feedback!

I had to add it because the user passed by environment variable ACTIVEMQ_WEB_USER (and its password) was not  having any effect when accessing the web console, as the configuration file jetty.xml points to jetty-realm.properties and not to the file which the sed command was modifying...

`    <bean id="securityLoginService" class="org.eclipse.jetty.security.HashLoginService">
        <property name="name" value="ActiveMQRealm" />
        <property name="config" value="${activemq.conf}/jetty-realm.properties" />
    </bean>
`
I dont need separate access, I just need to change the default user and password. Should I do something different?

jbonofre:
@lionelraymundi you should take a look on the `Dockerfile` from `activemq-5.18.x` branch.

PR #1121, opened 2023-11-17
WIP: [AMQ-9394] Tech Preview: Virtual Thread support

mattrpav:
TODO:
- [x] Throw exception if virtualThreadTaskRunner is enabled and JDK 21 (or higher) is not available
- [x] Breakout the Virtual Thread factory init so it only logs on JDK 17
- [x] Add webpage with instructions and implementation progress status (https://activemq.apache.org/virtual-threads)
- [ ] Add a VirtualThreadTaskRunner
- [ ] Consider separate config flag for persistenceAdapter TaskRunner
- [ ] Replace Thread.yield() usage with VirtualThread friendly alternative

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -222,6 +222,7 @@ public class BrokerService implements Service {
     private boolean monitorConnectionSplits = false;
     private int taskRunnerPriority = Thread.NORM_PRIORITY;
     private boolean dedicatedTaskRunner;
+    private boolean virtualThreadTaskRunner;
     private boolean cacheTempDestinations = false;// useful for failover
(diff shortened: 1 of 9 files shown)


PascalSchumacher:
Great to see work done on this! 👍 

What about other threads like `ActiveMQ Data File Writer`, `ActiveMQ Journal *`, can/should these be virtual too or is that a bad idea?

mattrpav:
> Great to see work done on this! 👍
> 
> What about other threads like `ActiveMQ Data File Writer`, `ActiveMQ Journal *`, can/should these be virtual too or is that a bad idea?

Those are single threaded processing right now. They would most benefit from being converted (or create a configurable alternative) to use Linux's IO_uring (aka kqueue in macOS) vs Virtual Thread (which would block on I/O anyway).

mattrpav:
Newer PR here: #1172

PR #1123, opened 2023-11-21
[AMQ-9398] Fix activemq-ra/activemq-rar namespaces for Jakarta EE

mattrpav:


activemq-ra/src/main/java/org/apache/activemq/ra/ActiveMQActivationSpec.java
```diff
@@ -147,25 +147,19 @@ public void validate() throws InvalidPropertyException {
...
-        try {
-            new Boolean(useRAManagedTransaction);
-            return true;
-        } catch (Throwable e) {
-            //
+        boolean retVal = Boolean.valueOf(useRAManagedTransaction);
+        if(!retVal) {
+            errorMessages.add("useRAManagedTransaction must be set to: true or false.");
         }
-        errorMessages.add("useRAManagedTransaction must be set to: true or false.");
```
cshannon, line 156: This seems wrong as this will add an error message even for "false"

activemq-ra/src/main/java/org/apache/activemq/ra/ActiveMQActivationSpec.java
```diff
@@ -147,25 +147,19 @@ public void validate() throws InvalidPropertyException {
     }
 
     public boolean isValidUseRAManagedTransaction(List<String> errorMessages) {
-        try {
-            new Boolean(useRAManagedTransaction);
-            return true;
-        } catch (Throwable e) {
-            //
+        boolean retVal = Boolean.valueOf(useRAManagedTransaction);
+        if(!retVal) {
```
cshannon, line 151: This seems wrong as this will add an error message even for "false"

cshannon (approved):
LGTM

PR #1139, opened 2024-01-11
[AMQ-9344] Add configurable limit to uncommitted transaction

mattrpav:


activemq-broker/src/main/java/org/apache/activemq/broker/TransactionBroker.java
```diff
@@ -291,13 +294,43 @@ public void send(ProducerBrokerExchange producerExchange, final Message message)
...
+            verifyUncommittedCount(producerExchange, transaction, message);
             next.send(producerExchange, message);
         } finally {
             context.setTransaction(originalTx);
         }
     }
 
+    protected void verifyUncommittedCount(ProducerBrokerExchange producerExchange, Transaction transaction, Message message) throws Exception {
+        // maxUncommittedCount <= 0 disables
+        int maxUncommittedCount = this.getBrokerService().getMaxUncommittedCount();
```
cshannon, line 309: You should be able to just read this in the constructor and store it as a final variable on the TransactionBroker as it won't be changed later since there's no dynamic config for the property

activemq-broker/src/main/java/org/apache/activemq/broker/TransactionBroker.java
```diff
@@ -291,13 +294,43 @@ public void send(ProducerBrokerExchange producerExchange, final Message message)
             transaction = getTransaction(context, message.getTransactionId(), false);
         }
         context.setTransaction(transaction);
+
         try {
+            // [AMQ-9344] Limit uncommitted transactions by count
+            verifyUncommittedCount(producerExchange, transaction, message);
```
cshannon, line 300:
```suggestion
            if (maxUncommittedCount > 0) {
                verifyUncommittedCount(producerExchange, transaction, message);
            }
```
I would move the check here for enabled, it's a small performance hit (likely not really noticable) but still I don't see a reason to always invoke a method for every single message when this is disabled.

activemq-broker/src/main/java/org/apache/activemq/broker/TransactionBroker.java
```diff
@@ -291,13 +294,43 @@ public void send(ProducerBrokerExchange producerExchange, final Message message)
...
             next.send(producerExchange, message);
         } finally {
             context.setTransaction(originalTx);
         }
     }
 
+    protected void verifyUncommittedCount(ProducerBrokerExchange producerExchange, Transaction transaction, Message message) throws Exception {
+        // maxUncommittedCount <= 0 disables
+        int maxUncommittedCount = this.getBrokerService().getMaxUncommittedCount();
+        if (maxUncommittedCount > 0 && transaction.size() >= maxUncommittedCount) {
```
cshannon, line 310: Can remove maxUncommittedCount > 0 here if moved up as my other comment says

activemq-broker/src/main/java/org/apache/activemq/broker/TransactionBroker.java
```diff
@@ -291,13 +294,43 @@ public void send(ProducerBrokerExchange producerExchange, final Message message)
...
+                // the client to reset to a new transaction so we need to clean up
+                transaction.rollback();
+
+                // Send ResourceAllocationException which will translate to a JMSException
+                final ResourceAllocationException e = new ResourceAllocationException(
+                    "Can not send message on transaction with id: '" + transaction.getTransactionId().toString()
+                      + "', Transaction has reached the maximum allowed number of pending send operations before commit of '"
+                      + maxUncommittedCount + "'", "42900");
+                if(LOG.isDebugEnabled()) {
+                    LOG.warn("ConnectionId:{} exceeded maxUncommittedCount:{} for destination:{} in transactionId:{}", (producerExchange.getConnectionContext() != null ? producerExchange.getConnectionContext().getConnectionId() : "<not set>"), maxUncommittedCount, message.getDestination().getQualifiedName(), transaction.getTransactionId().toString(), e);
```
cshannon, line 323: This is logging at the warn level which doesn't make sense since you are checking if debug is enabled.

The formatting here is also bad and should be multiple lines and not one line so it's readable

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountExceededTest.java
```diff
@@ -0,0 +1,217 @@
...
+
+        return step1[0] + "." + step3[0].split("=", 2)[1] + "." + step3[1].split("=", 2)[1];
+    }
+
+    protected QueueViewMBean getQueueViewMBean(ActiveMQDestination destination) throws Exception {
+        return JMX.newMBeanProxy(mbeanServer, BrokerMBeanSupport.createDestinationName(BrokerMBeanSupport.createBrokerObjectName(DEFAULT_JMX_DOMAIN_NAME, DEFAULT_JMX_BROKER_NAME).toString(), destination), QueueViewMBean.class);
+    }
+
+    @Test
+    public void testUncommittedCountExceededSync() throws Exception {
```
cshannon, line 156: A lot  of the code in these two tests are identical and copy and pasted. This could be refactored to re-use some of the code or maybe even using parameterized tests to reduce the bloat.

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountExceededTest.java
```diff
@@ -0,0 +1,217 @@
...
+                Message message = session.createBytesMessage();
+                message.setIntProperty("IDX", i);
+                messageProducer.send(message);
+            }
+        } catch (ResourceAllocationException e) { 
+            caught = true;
+            caughtException = e;
+        }
+
+        //Thread.sleep(10_000l);
```
cshannon, line 180: Can remove

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountExceededTest.java
```diff
@@ -0,0 +1,217 @@
...
+        Queue queue = session.createQueue(methodNameDestinationName);
+        messageProducer = session.createProducer(queue);
+        mbeanServer = ManagementFactory.getPlatformMBeanServer();
+        brokerService.getAdminView().addQueue(methodNameDestinationName);
+        queueViewMBean = getQueueViewMBean(new ActiveMQQueue(methodNameDestinationName));
+    }
+
+    @After
+    public void tearDown() throws Exception {
+        if (messageProducer != null) {
```
cshannon, line 99: The code here to close the producer and session seems un-necessary as closing the connection will also close these. You also don't really need to set the reference to null or any references to null as they will be reset in the setUp() method when running in the future or just cleaned up by the garbage collector when the test is done. 

It's not a huge deal but the code base has a ton of this kind of extra code laying around that is unnecessary and just leads to more bloat. We really should just create utilities or  move this clean up stuff to a common section anyways.

cshannon (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/TransactionBroker.java
mattrpav: When DEBUG is enabled, the log is at the same level, but includes the exception

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountExceededTest.java
mattrpav: Agreed, I'd like to build a JUnit 5 set of tools to provide all this boilerplate code.

activemq-broker/src/main/java/org/apache/activemq/broker/TransactionBroker.java
cshannon: That should be fine since it's just a getter I was just figuring it's pretty unlikely it would ever be changed, it would be rare as likely you are going to set the limit to something reasonable in the first place.

mattrpav: Correct. Changing the value at runtime would be 'emergency playbook' or 'testing different scenarios' for the most part.

activemq-broker/src/main/java/org/apache/activemq/broker/TransactionBroker.java
mattrpav: Secondary logging removed. We are the _thrower_ no need to log the exception here

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountExceededTest.java
mattrpav: Fixed

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountExceededTest.java
mattrpav: Fixed.

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountExceededTest.java
mattrpav: Fixed. Test converted to parameterized

activemq-broker/src/main/java/org/apache/activemq/broker/TransactionBroker.java
```diff
@@ -291,13 +294,39 @@ public void send(ProducerBrokerExchange producerExchange, final Message message)
...
             next.send(producerExchange, message);
         } finally {
             context.setTransaction(originalTx);
         }
     }
 
+    protected void verifyUncommittedCount(ProducerBrokerExchange producerExchange, Transaction transaction, Message message) throws Exception {
+        // maxUncommittedCount <= 0 disables
+        int maxUncommittedCount = this.getBrokerService().getMaxUncommittedCount();
+        if (maxUncommittedCount > 0 && transaction.size() >= maxUncommittedCount) {
```
cshannon, line 310: Shouldn't this be a greater than check and not >= ? We only want to fail if we go past the max size not if we are at it. This is why the test has 11 failed ops but should only be 10 since your test you set the max to 10 so we should allow 10 and then fail on 11 - 20

cshannon: Nevermind, equals makes sense since you've already sent the max previous (this is on the send of the next one) so i'm still wondering why the test assertion is 11 and not 10 though.

cshannon (approved):
LGTM, probably need to do some more real world testing with it but it's off by default for now so people can experiment if they want

PR #1142, opened 2024-01-23
AMQ-9420 - Don't decrement KahaDB durable sub metrics on duplicate ack

cshannon:
This adds a check in case a duplicate ack is passed to the store to make sure that the subscription statistics (if enabled) for a durable sub do not have the metrics decremented a second time

I also noticed one small spot where we should probably increase the metrics for the store after adding to the order index (not before) just in case of exception and someone doesn't have kahadb configured to shutdown on io exception

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
@@ -1545,8 +1545,9 @@ public abstract class MessageDatabase extends ServiceSupport implements BrokerSe
         if (previous == null) {
             previous = sd.messageIdIndex.put(tx, command.getMessageId(), id);
             if (previous == null) {
-                incrementAndAddSizeToStoreStat(tx, command.getDestination(), location.getSize());
                 sd.orderIndex.put(tx, priority, id, new MessageKeys(command.getMessageId(), location));
+                // increment after putting into the order index first
+                incrementAndAddSizeToStoreStat(tx, command.getDestination(), location.getSize());
                 if (sd.subscriptions != null && !sd.subscriptions.isEmpty(tx)) {
                     addAckLocationForNewMessage(tx, command.getDestination(), sd, id);
                 }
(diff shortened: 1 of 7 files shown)


activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
```diff
@@ -1545,8 +1545,9 @@ long updateIndex(Transaction tx, KahaAddMessageCommand command, Location locatio
         if (previous == null) {
             previous = sd.messageIdIndex.put(tx, command.getMessageId(), id);
             if (previous == null) {
-                incrementAndAddSizeToStoreStat(tx, command.getDestination(), location.getSize());
                 sd.orderIndex.put(tx, priority, id, new MessageKeys(command.getMessageId(), location));
```
jbonofre, line 1548: Shouldn't we use the `IndexLock` here to have the `orderIndex` and store stat updated atomically ? Just wondering.

cshannon: The entire method is already protected by the index write lock here: https://github.com/apache/activemq/blob/0b95f93449e58a855efd2fb03731432f396b18e1/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java#L1342

This change is probably not even necessary since it's already under a lock and any exception should abort and shut things down normally but it just seemed to make more sense to not increment until adding to the index. but really it doesn't much matter since the index isn't commuted anyways if it isn't successful.

jbonofre: Good point.

cshannon: Yeah your comment was good as it made me realize that change was pointless as I forgot the index isn't even committed until after the transaction completes. One could argue we could wait to update the counters until after transaction finishes but that would be a much larger change and tricky to guarantee things are in fact still atomic since you process it later outside of where you have the information you need and the lock. Plus the default behavior is to shutdown the broker on IoException anyways. The current behavior has been fine for many years so don't need to change it I don't think.

cshannon:
This is a pretty simple and straight forward change so I'm going to merge (the majority of code is testing updates) and backport. I will keep an eye on the full CI builds

cshannon:
I forgot to squash my two commits into 1 but i will fix that and squash when i cherry-pick

PR #1147, opened 2024-01-30
[AMQ-9425] Upgrade slf4j 2.0.11

mattrpav:


pom.xml
```diff
@@ -71,7 +71,7 @@
     <jaxb-version>4.0.2</jaxb-version>
     <jaxb-bundle-version>2.3.2_1</jaxb-bundle-version>
     <jetty-version>11.0.18</jetty-version>
-    <jmdns-version>3.5.8</jmdns-version>
+    <jmdns-version>3.5.9</jmdns-version>
```
jbonofre, line 74: Can you please remove this change from this PR (it's unrelated) ?

pom.xml
```diff
@@ -94,7 +94,7 @@
     <regexp-version>1.4</regexp-version>
     <rome-version>2.1.0</rome-version>
     <shiro-version>1.13.0</shiro-version>
-    <slf4j-version>2.0.9</slf4j-version>
+    <slf4j-version>2.0.11</slf4j-version>
```
jbonofre, line 97: I have to double check the package version in the OSGi/Karaf part. It should be OK but I would like to double check (as I did some other updates).

jbonofre (changes requested):

PR #1150, opened 2024-01-30
[AMQ-9428] Upgrade commons-io 2.15.1

mattrpav:


jbonofre (approved):

jbonofre (changes requested):
I'm checking because this upgrade may break some parts of `activemq-http`. I'm double checking to be sure.

jbonofre (approved):
I checked and all is OK.

PR #1153, opened 2024-02-15
AMQ-9435 - Ensure orderIndex next id is rolled back on duplicates

cshannon:
This commit fixes a bug in KahaDB that caused gaps in sequence ack tracking for durables that would lead to the appearance of stuck messages on durable subs if duplicate messages were detected. The sequence is now correctly rolled back so that there is no gap if the message is not added to the order index


jbonofre (approved):
Good one, thanks @cshannon !

PR #1154, opened 2024-02-15
AMQ-9436 - Ensure message audit in queue store cursor is shared

cshannon:
This commit fixes the initialization of the StoreQueueCursor message audit object to make sure it's shared between the persistent and non persistent cursors. It also adds a check to ensure that duplicate calls to start will not try and init more than once.


mattrpav (approved):
LGTM

PR #1155, opened 2024-02-16
AMQ-9432 Disable default Jetty WebSocket idle timeout

thezbyg:


mattrpav:
This is a good improvement! @jbonofre how about re-using the inactivity monitor value? Not sure we need a separate config flag.

thezbyg:
I do not mind, but should this WebSocket idle timeout still be active after higher level protocol (MQTT, STOMP) completes keep-alive negotiation with client and inactivity timer is started? Clients could also request to not have any keep-alive. I think that this WebSocket idle timeout is only useful until higher level protocol inactivity timer is started.

jbonofre:
@mattrpav good idea but I'm not sure it would work for WS (it has to happen before the neg).
I will try this way first and fallback to a `ws.idle-property` like property.

jbonofre:
@mattrpav I agree. Let me update the PR with this.

jbonofre:
As the idle timeout set to 30s by default has been introduced in Jetty 10. So it makes sense to force it in ActiveMQ to be consistent with previous version. Being able to configure the timeout can be in another PR.

jbonofre (approved):

PR #1156, opened 2024-02-23
[AMQ-9437] Add policyEntry advancedNetworkStatisticsEnabled flag

mattrpav:
PolicyEntry option advancedNetworkStatisticsEnabled (default: false)

[AdvancedNetworkStatistics]
- [x] networkEnqueueCount
- [x] networkDequeueCount
- [x] Update unit tests to validate policyEntry and runtimeConfigurationPlugin
- [x] Update unit tests to validate statistics are working as expected
- [x] Add null guard for null context/connection scenarios
- [x] Add queue unit test
- [x] Improve unit test to confirm non-network use case counting is correct

This change introduces the configuration guard for _advancedNetworkStatistics_ as well as two simple metrics to provide fine-grained counting for network enqueue and dequeue counts that are enabled behind this config flag.

Usage notes:

* ActiveMQ.Advisory.> topics should never have advancedNetworkStatisticsEnabled set to 'true'
* High-performing non-persistent topic workflows may not want this enabled due to potential for latency penalties.


cshannon (changes requested):
The problem I have with this is creating a single flag for all of these extra metrics, and also the name is just very generic with no specifics. If I'm a user and i see "advancedStatistics" I have no idea what that means.

Grouping together a bunch of random statistics with a single on/off doesn't make sense to me because all of these extra statistics will have different levels of performance hits and users might only want some of them. Someone might want extra networking stats but doesn't care about the other stuff you mentioned. 

So I think that we probably need multiple flags to make it more fine grained. I view it similarly to the advisory configuration on a policy where you can control which advisories are sent. I think different stats that are unrelated should be grouped together under a different flag to turn them on or off. Obliviously this is bridge related so the flag could be called some like advancedBridgeStats or advancedNetworkStats or whatever.

mattrpav:
> This change is breaking `activemq-broker` tests:
> 
> ```
> Caused by: java.lang.NullPointerException: Cannot invoke "org.apache.activemq.broker.Connection.isNetworkConnection()" because the return value of "org.apache.activemq.broker.ConnectionContext.getConnection()" is null
>         at org.apache.activemq.broker.region.Queue.messageSent(Queue.java:1978)
>         at org.apache.activemq.broker.region.Queue.doPendingCursorAdditions(Queue.java:843)
>         at org.apache.activemq.broker.region.Queue$CursorAddSync.afterCommit(Queue.java:861)
>         at org.apache.activemq.transaction.Transaction.fireAfterCommit(Transaction.java:140)
>         at org.apache.activemq.transaction.Transaction.doPostCommit(Transaction.java:211)
> ```
> 
> I'm fixing.

Null guard added in latest commit [026b8ec](https://github.com/apache/activemq/pull/1156/commits/026b8ece426749dd84217bedf1b571e8248aabc8)

mattrpav:
> The problem I have with this is creating a single flag for all of these extra metrics, and also the name is just very generic with no specifics. If I'm a user and i see "advancedStatistics" I have no idea what that means. Also there's nothing that really makes this advanced per se, it's just more stats.

Parameter renamed in latest commit [026b8ec](https://github.com/apache/activemq/pull/1156/commits/026b8ece426749dd84217bedf1b571e8248aabc8)

cshannon (approved):
LGTM, the new property name is more descriptive and a bit better

PR #1192, opened 2024-03-31
[AMQ-9447] Added Logic to check underlying transport connector limits

AM-19:
Hello Team,

With Regards to Transport Connector, Have added a check based on the connection count.

Also in the ticket, With regards to JDBC Persistence Check, We may need to have JDBC persistence associated classes in activemq-broker,jar, Post that we can have a logic something like following:

`        if (brokerService != null && brokerService.getPersistenceAdapter() != null 
        		&& brokerService.getPersistenceAdapter() instanceof JDBCPersistenceAdapter ) {
        	JDBCPersistenceAdapter jdbcAdapter = (JDBCPersistenceAdapter) brokerService.getPersistenceAdapter();
			String message = "Issue with Persistence layer";
            try {
            	ResultSet rs = null;
            	Statement stmt = jdbcAdapter.getDataSource().getConnection().createStatement();
            	rs = stmt.executeQuery("SELECT 1");
				String message = 
            	if(rs.next())
					if(rs.getInt(1)!=1)
						answer.add(new HealthStatus("org.apache.activemq.store.jdbc.JDBCPersistenceAdapterIssue", "ERROR", message, jdbcAdapter.toString()));
            } catch(Exception e) {
            	answer.add(new HealthStatus("org.apache.activemq.store.jdbc.JDBCPersistenceAdapterIssue", "ERROR", message, jdbcAdapter.toString()));
            }            
        }`

Ive Tested the above and it works as expected. But again that would require whole jdbc persistence adapter code.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/HealthView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/HealthView.java
@@ -167,6 +167,25 @@ public class HealthView implements HealthViewMBean {
             }
         }
 
+        /**
+         * Check The Transport Connector limits
+         */
+
+        if (brokerService != null && !brokerService.getTransportConnectors().isEmpty()) {
+        	for(TransportConnector tc: brokerService.getTransportConnectors()) {
+        		if(tc.getServer() instanceof TcpTransportServer) {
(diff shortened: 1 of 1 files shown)


jbonofre:
That's a good start, on Tcp transport. I will add additional more "specific" transport (like http/https/ws transport).

AM-19:
Hello @jbonofre 

Have Re-opened as #1194 
Sorry for the inconvenience.

Thanks
Anubhav

PR #1193, opened 2024-03-31
AMQ-9460: Don't override ACTIVEMQ_OPTS when provided by user

jbonofre:


assembly/src/release/bin/setenv
```diff
@@ -33,9 +33,7 @@
 if [ -z "$ACTIVEMQ_OPTS_MEMORY" ] ; then
     ACTIVEMQ_OPTS_MEMORY="-Xms64M -Xmx1G"
 fi
-if [ -z "$ACTIVEMQ_OPTS" ] ; then
-    ACTIVEMQ_OPTS="$ACTIVEMQ_OPTS_MEMORY -Djava.util.logging.config.file=logging.properties -Djava.security.auth.login.config=$ACTIVEMQ_CONF/login.config"
-fi
+ACTIVEMQ_OPTS="$ACTIVEMQ_OPTS_MEMORY $ACTIVEMQ_OPTS -Djava.util.logging.config.file=logging.properties -Djava.security.auth.login.config=$ACTIVEMQ_CONF/login.config"
```
gemmellr, line 36: Seems like this maybe just broke the ability to set your own settings for these props in ACTIVEMQ_OPTS, since the later ones will take precedence, and it generally just looks to mess with the behaviour of ACTIVEMQ_OPTS handling. Given it seems it worked the way it did for 10+ years, that seems a questionable change in behaviour to aim at a patch release.

This overall change actually seems weird in general as the title is "_Don't override ACTIVEMQ_OPTS when provided by user_" and yet it looks more like it mostly does the opposite of that for folks not using Docker, given this suggests it actually specifically overrides ACTIVEMQ_OPTS provided by the user and replaces it with its own munged value (that incorporates the users value, somewhat...but in a way it can still override what they wanted to do...and in a way that changes it significantly from its historic behaviour of using only exactly what the user specified, rather than merging it with other config).

jbonofre: The point is that logging.config and auth.config should be provided: if the user removes it, it breaks the authentication, etc. So it's "confusing".

I can move this to another property and is always added and let the user define ACTIVEMQ_OPTS still.

jbonofre: Another option is to populate logging.config and auth.config in Dockerfile.

gemmellr: A different change that doesnt alter the long standing existing behaviour or break anyone, might be to give a more general way of _adding_ options without needing to define them _all_. E.g adding a new env variable that only appends to the options stuff rather than define it all, and then using that in the docker file to pass the additional option. E.g put it back the way it was, and add a line below here which appends the value of a new variable if set.  Artemis did something similar a while back to allow appending specific additional args rather than needing to set them all (which worked the same way as these scripts did before the change...either you specify them all, or the defaults are used).

jbonofre: Let me do this way.

gemmellr: https://github.com/apache/activemq-artemis/commit/45ffea9ef824bd009bc47c94e6c9e6285cd83e3a is the Artemis change I was remembering, added 'JAVA_ARGS_APPEND' to align to its existing 'JAVA_ARGS' variable. So e.g ACTIVEMQ_OPTS_APPEND might be a candidate in this case.

PR #1194, opened 2024-03-31
[AMQ-9447] Added Logic to check underlying transport connector limits

AM-19:
(Re-Opened)
Hello Team,

With Regards to Transport Connector, Have added a check based on the connection count.

Also in the ticket, With regards to JDBC Persistence Check, We may need to have JDBC persistence associated classes in activemq-broker,jar, Post that we can have a logic something like following:

`
        if (brokerService != null && brokerService.getPersistenceAdapter() != null 
        		&& brokerService.getPersistenceAdapter() instanceof JDBCPersistenceAdapter ) {
        	JDBCPersistenceAdapter jdbcAdapter = (JDBCPersistenceAdapter) brokerService.getPersistenceAdapter();
			String message = "Issue with Persistence layer";
            try {
            	ResultSet rs = null;
            	Statement stmt = jdbcAdapter.getDataSource().getConnection().createStatement();
            	rs = stmt.executeQuery("SELECT 1");
				String message = 
            	if(rs.next())
					if(rs.getInt(1)!=1)
						answer.add(new HealthStatus("org.apache.activemq.store.jdbc.JDBCPersistenceAdapterIssue", "ERROR", message, jdbcAdapter.toString()));
            } catch(Exception e) {
            	answer.add(new HealthStatus("org.apache.activemq.store.jdbc.JDBCPersistenceAdapterIssue", "ERROR", message, jdbcAdapter.toString()));
            }            
        }
`

Ive Tested the above and it works as expected. But again that would require whole jdbc persistence adapter code.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/HealthView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/HealthView.java
@@ -29,8 +29,10 @@ import javax.management.openmbean.TabularDataSupport;
 import javax.management.openmbean.TabularType;
 
 import org.apache.activemq.broker.BrokerService;
+import org.apache.activemq.broker.TransportConnector;
 import org.apache.activemq.broker.scheduler.JobSchedulerStore;
 import org.apache.activemq.store.PersistenceAdapter;
+import org.apache.activemq.transport.tcp.TcpTransportServer;
 import org.apache.activemq.usage.SystemUsage;
 
 public class HealthView implements HealthViewMBean {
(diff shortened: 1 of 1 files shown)


jbonofre (approved):

jbonofre:
This first step checking the transports looks good to me. I would also check the storage service.

AM-19:
> This first step checking the transports looks good to me. I would also check the storage service.

Ideally the above mentioned code for JDBC Persistence check will work, with the appropriate imports.

jbonofre:
For JDBC yes, but no test on KahaDB. We can check the system usage for instance to see if the filesystem is not almost full.

AM-19:
It seems we already have a check for File system usage.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/HealthView.java
```diff
@@ -167,6 +169,25 @@ public List<HealthStatus> healthList() throws Exception {
...
+        /**
+         * Check The Transport Connector limits
+         */
+
+        if (brokerService != null && !brokerService.getTransportConnectors().isEmpty()) {
+        	for(TransportConnector tc: brokerService.getTransportConnectors()) {
+        		if(tc.getServer() instanceof TcpTransportServer) {
+        			int connectionUsage = (int) (((TcpTransportServer)tc.getServer()).getCurrentTransportCount().get() * 100 ) / ((TcpTransportServer)tc.getServer()).getMaximumConnections();
+        			if(((TcpTransportServer)tc.getServer()).getCurrentTransportCount().get() >= ((TcpTransportServer)tc.getServer()).getMaximumConnections()) {
+        				String message = "Exceeded the maximum number of allowed client connections: " + ((TcpTransportServer)tc.getServer()).getMaximumConnections();
```
mattrpav, line 181: This check can’t happen since the connector will close a connection when it reaches the limit. An exceeded count isn’t possible. 

I think this branch of logic should just be removed and simply go with the limit check

mattrpav (changes requested):

mattrpav (changes requested):
Please add a quick unit test

jbonofre:
> It seems we already have a check for File system usage.

We have a check on the system usage, but not on the cursor.

It's not super critical.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/HealthView.java
jbonofre: Agree, just the limit is enough.

We can relay on the existing metrics for that.

AM-19:
> Please add a quick unit test

Sure, Let me work on that

AM-19:
> Please add a quick unit test

Have Added the same, Please have a look.

AM-19:
Hello Matt, 
Seems JB have approved and its waiting for your approval.
If something is missing, do let me know I will work on the same.

PR #1196, opened 2024-04-01
AMQ-9460: Fix authentication on Docker images

jbonofre:


jbonofre:
@gemmellr I changed the fix to be specific on Docker images, no impact on the scripts.

gemmellr:
Since this is now setting an override for defaulted variables, it could likely also now need updated any time the original values it overrides are updated. Seems like it has potential to go stale. I'd have gone the other way mentioned on #1193 given its useful both to fix this and also more generally, and doesnt add extra tied maintenance for later.

jbonofre:
@gemmellr as I don't want to hold the release longer, I will go ahead with this quick fix and create a Jira to track more generic improvement.

PR #1200, opened 2024-04-10
AMQ-9475 - ConsumerControl commands should not auto create wildcard dests

cshannon:
This fixes an issue where wildcard destinations could be inadvertently created by ConsumerControl commands which could lead to problems. The processing logic for this command now will only look up existing destinations if they exist.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
@@ -21,6 +21,8 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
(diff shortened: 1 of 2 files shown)


mattrpav (approved):
LGTM

cshannon:
I'll let the tests finish and then I'll merge and backport.

thezbyg:
@cshannon @mattrpav I rebuilt ActiveMQ with these changes and can confirm that wildcard topics are no longer created and everything works as expected.

Thank you.

cshannon:
> @cshannon @mattrpav I rebuilt ActiveMQ with these changes and can confirm that wildcard topics are no longer created and everything works as expected.
> 
> Thank you.

Thanks for the quick test, glad that worked for you. This will be included in the next bug fix releases for 6.1.x and 5.18.x. 5.18.4 is already under vote so this will be in 5.18.5 but obviously you can just build your own version as you did.

PR #1202, opened 2024-04-11
AMQ-9473: Fix IntrospectionSupport to deal with SSLSocket properties

jbonofre:


mattrpav:
We should add a unit test for this JIRA https://issues.apache.org/jira/browse/AMQ-9478 assigned to 6.2.0

PR #1203, opened 2024-04-11
AMQ-9330 - Return 204 code when polling empty destinations

cshannon:
Previously a timeout was thrown and complete was not called so Jetty was returning a 500 error


jbonofre (approved):
LGTM. I tested and it works fine. Thanks !

PR #1204, opened 2024-04-11
[AMQ-9430] RuntimeConfigPlugin wire in classloader to JAXBContext

mattrpav:


jbonofre (approved):
LGTM. I just tested and it works fine on JMX MBean. Thanks !

mattrpav:
I confirmed testing this locally using jconsole

```
 INFO | Manual configuration update triggered
 INFO | No material change to configuration in class path resource [activemq.xml] at: Thu Apr 11 09:31:19 CDT 2024
```

Background: The issue is activemq distribution uses a different classloader to bootstrap. There is a unit test for jmx invocation, but it is not like-for-like b/c of the runtime differences between the unit test and the apache activemq dist. 

This fix should work for all dists, including embedded (karaf, etc)

cshannon (approved):

PR #1206, opened 2024-04-21
AMQ-9481 - Correctly complete async servlet request on timeout

cshannon:
This fixes AsyncServletRequest to correctly call context.dispatch() when the async request times out so that the consumer can be re-used.


mattrpav (approved):
LGTM

PR #1207, opened 2024-04-23
WIP: [AMQ-9484] Support exporting kahadb messages from a queue with an offset - Option B

mattrpav:


cshannon:
I only took a quick glance but 2 things stand out to me immediately that are issues:

1. The current PR is using recoverNextMessages() which changes state of the cursor, so this definitely won't work in an online broker without breaking the queue if the intent is for a user to use this to export while a broker is online.
2. This scans through the index and just skips when you could just use MessageOrderIndex with the order iterator and skip ahead by setting the batch location on the index

This needs a bit more context with use cases to see how this is intended to be used to probably know the best way of doing some of this

cshannon (changes requested):
I only took a quick glance but 2 things stand out to me immediately that are issues:

The current PR is using recoverNextMessages() which changes state of the cursor, so this definitely won't work in an online broker without breaking the queue if the intent is for a user to use this to export while a broker is online.
This scans through the index and just skips when you could just use MessageOrderIndex with the order iterator and skip ahead by setting the batch location on the index
This needs a bit more context with use cases to see how this is intended to be used to probably know the best way of doing some of this

mattrpav:
One reason I thought iterating over the index was preferred vs seeking over the index was to skip over and not include any messages in the ackedAndPrepared collection.

```
    if (ackedAndPrepared != null && ackedAndPrepared.contains(entry.getValue().messageId)) {
        continue;
    }
```

mattrpav:
Closing in favor of https://github.com/apache/activemq/pull/1208

mattrpav:
Closing in favor of https://github.com/apache/activemq/pull/1208

PR #1209, opened 2024-04-24
[AMQ-9485] Host activemq-protobuf modules for modernization

mattrpav:


mattrpav:
I think the benefits are that it is versioned as the same with the broker. It is used to generate classes for KahaDB, so building utilities for working with KahaDB stores are wonky to follow-- v1.1 of activemq-protobuf for ActiveMQ 5.x and 6.x.  Then when there is a change is it v1.2 or 2.0 for ActiveMQ 7.x, etc.

Also, when there are some minor changes (like I have tee'd up for compiler deprecation WARNs, etc) it is one less thing to vote on and go through all the release iterations when they can simply be rolled into the next ActiveMQ Classic release.

cshannon:
I don't think that there's any problem having different versions, we have a ton of dependencies, there's no reason we need to have the same version match.

It's also not a huge deal to go through the release process. I agree with you that if this was something that changed every release that would be fine, but it's not. This is the first time I can even remember needing to make changes to this project in my 10 years being around. If there are a bunch of duplicated classes we can just fix it up in the exist repo and main classic base to de-duplicate and modernize.

So I don't really think this is necessary, it's like the OpenWire generator now being separate. I view this code in the same way that it should not change often so I don't see a need to bring it into the main project because it's something that should almost never change.

cshannon:
@jbonofre - What do you think?

cshannon:
If we do bring this into the main project then maybe we should bring Openwire as well, it would just be nice to be consistent. I view this project in much the same was as the new Openwire generator where it's not going to change often but it's tied to the broker so I think we should either bring them both in or keep them both separate.

jbonofre:
Originally, the reason to have this "outside" for the broker repo is to have a sharable artifact with its own release cycle.
If the purpose here is to "align" the broker with protobuf module, it makes sense.

I think we should check on the dev mailing list if we don't have other "consumers" of protobuf.

mattrpav:
Analysis from mvnrepository.com shows it is used in a boms from two organizations-- Mulesoft and Fuse/Fabric8.io. Looks like just packagers of ActiveMQ that would then get it 'for free' if the jar was in the distro.

ref: https://mvnrepository.com/artifact/org.apache.activemq.protobuf/activemq-protobuf/1.1

mattrpav:
Goals:

1. Align version number with the release
2. Easier alignment of JDK language and runtime compatibly (current code needs clean-up to remove WARNs on compile)
3. Future change to consolidate redundant I/O related classes across the project — ASCIIBuffer, UTF8Buffer, *Stream classes, etc. 
4. Ability to update the format as needed for future enhancements and features (ie add a serial number of active-active replication, etc)
5. Initially, no change to Maven coordinates. Only version number to align w/ broker release(s)

Note: #3 will also help with security and parsing consistency issuesetC)

cshannon (changes requested):
Why is there a separate module for test vs putting all the test stuff under src/test ? If you need to package test code into a jar you can build a test jar from that

mattrpav:
This PR is to move 'as-is' from the previous module. The "test" module uses the main module as a dependent maven module in the build to test processing the protobuf definitions.

cshannon (approved):

PR #1222, opened 2024-05-13
[Misc] Add README.md and code hygiene

kenliao94:
* Add README to help new developers with onboarding and navigating the code.
* Rewrote some for-loop and unnecessary conditional check for cleaner code.

### DIFF
--- a/activemq-broker/README.md
+++ b/activemq-broker/README.md
@@ -1,0 +1,6 @@
+## Description
+This package contains the ActiveMQ Message Broker implementation.
+
+## Contribution guide (WIP)
+#### Interesting classes to start diving into:
+`src/main/java/org/apache/activemq/broker/BrokerService.java`: Manages the life-cycle of an ActiveMQ Broker.
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -271,8 +271,8 @@ public class BrokerService implements Service {
     static {
 
         try {
-            Boolean bouncyCastleNotAdded = Boolean.getBoolean("org.apache.activemq.broker.BouncyCastleNotAdded");
-            if (bouncyCastleNotAdded == null || bouncyCastleNotAdded == false) {
+            boolean bouncyCastleNotAdded = Boolean.getBoolean("org.apache.activemq.broker.BouncyCastleNotAdded");
+            if (!bouncyCastleNotAdded) {
                 ClassLoader loader = BrokerService.class.getClassLoader();
                 Class<?> clazz = loader.loadClass("org.bouncycastle.jce.provider.BouncyCastleProvider");
                 Provider bouncycastle = (Provider) clazz.getDeclaredConstructor().newInstance();
@@ -286,7 +286,8 @@ public class BrokerService implements Service {
                 LOG.info("Loaded the Bouncy Castle security provider at position: {}", ret);
             }
         } catch(Throwable e) {
-            // No BouncyCastle found so we use the default Java Security Provider
+            // No BouncyCastle found, so we use the default Java Security Provider
+            LOG.info("Using the default Java security provider");
         }
 
         String localHostName = "localhost";
@@ -768,7 +769,7 @@ public class BrokerService implements Service {
         startAllConnectors();
 
         LOG.info("Apache ActiveMQ {} ({}, {}) started", getBrokerVersion(), getBrokerName(), brokerId);
-        LOG.info("For help or more information please see: http://activemq.apache.org");
+        LOG.info("For help or more information please see: https://activemq.apache.org");
 
         getBroker().brokerServiceStarted();
         checkStoreSystemUsageLimits();
@@ -831,10 +832,8 @@ public class BrokerService implements Service {
             this.scheduler.stop();
             this.scheduler = null;
         }
-        if (services != null) {
-            for (Service service : services) {
-                stopper.stop(service);
-            }
+        for (Service service : services) {
+            stopper.stop(service);
         }
         stopAllConnectors(stopper);
         this.slave = true;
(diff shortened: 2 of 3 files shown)


activemq-broker/README.md
```diff
@@ -0,0 +1,6 @@
+## Description
```
jbonofre, line 1: I'm not a big fan of having `README.md` in a module. It can lead to scattered the information in different place.
I would rather update the `README.md` at root level. We can eventually add a `CONTRIBUTING.md` file dedicate to contribution.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -831,10 +832,8 @@ public void stop() throws Exception {
             this.scheduler.stop();
             this.scheduler = null;
         }
-        if (services != null) {
-            for (Service service : services) {
-                stopper.stop(service);
-            }
+        for (Service service : services) {
```
jbonofre, line 835: I think we should still check if `services` is not `null` as it can be passed programmatically.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -2048,7 +2047,7 @@ protected void checkTmpStoreUsageLimits() throws Exception {
                 long maxJournalFileSize;
 
                 PListStore store = usage.getTempUsage().getStore();
-                if (store != null && store instanceof JournaledStore) {
+                if (store instanceof JournaledStore) {
```
jbonofre, line 2050: Same here, `store` can be `null` if temp usage is not set.

jbonofre (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
kenliao94: Yes. But `instanceof` should handle the nullchecking, so the first check is redundant.

jbonofre: Ah yeah, good catch. Thanks !

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
kenliao94: Interesting, because services is marked as final when it's instantiated in line 180. So services should never be null. Am I read the code correctly?

activemq-broker/README.md
kenliao94: Ok sounds good. Will remove it at the module level. Because this is work in progress so I don't want to increase its visibility. But I will push the final documentation to the root level README when it's mature and we can review that :)

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -286,7 +286,8 @@ public class BrokerService implements Service {
                 LOG.info("Loaded the Bouncy Castle security provider at position: {}", ret);
             }
         } catch(Throwable e) {
-            // No BouncyCastle found so we use the default Java Security Provider
+            // No BouncyCastle found, so we use the default Java Security Provider
+            LOG.info("Using the default Java security provider");
```
mattrpav, line 290: Does this need to be printed out on every startup?  Perhaps this is debug level?

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -1768,7 +1767,7 @@ public synchronized PListStore getTempDataStore() {
 
             try {
                 PersistenceAdapter pa = getPersistenceAdapter();
-                if( pa!=null && pa instanceof PListStore) {
+                if(pa instanceof PListStore) {
```
mattrpav, line 1770: Why is the null check being removed?

mattrpav (changes requested):
There are a bunch of varying changes here. Not sure it makes sense as a single commit. Perhaps the iteration Java-language updates go as one easy commit. 

I do not see a reason to remove null checks.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
kenliao94: Because `instanceof` already handles null checking. I.E if pa == null, instanceof will return false. So In this case, I believe the null checking is redundant. Do you see any concerns?

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
kenliao94: Good point. Changed it to debug.

kenliao94:
> There are a bunch of varying changes here. Not sure it makes sense as a single commit. Perhaps the iteration Java-language updates go as one easy commit.
> 
> I do not see a reason to remove null checks.
> 
> Is this PR part of a JIRA or NO-JIRA change?

This is part of a NO-JIRA change. I am just trying to clean up the code as I navigate the code :) I moved the removing redundant null checks from this PR to another I will create later to make this PR more about trivial updates.

kenliao94:
Hey Matt and JB, are there any more concerns about this PR? I checked the Jenkins failing tests they are about KahaDB

[INFO] ActiveMQ ........................................... SUCCESS [  0.852 s]

[INFO] ActiveMQ :: BOM .................................... SUCCESS [  0.248 s]

[INFO] ActiveMQ :: Openwire Generator ..................... SUCCESS [  1.091 s]

[INFO] ActiveMQ :: Client ................................. SUCCESS [02:50 min]

[INFO] ActiveMQ :: Openwire Legacy Support ................ SUCCESS [  2.411 s]

[INFO] ActiveMQ :: JAAS ................................... SUCCESS [01:46 min]

[INFO] ActiveMQ :: Broker ................................. SUCCESS [01:18 min]

[INFO] ActiveMQ :: KahaDB Store ........................... FAILURE [  03:30 h]


And the broker tests passed.

jbonofre (approved):

jbonofre:
I did a new pass and it looks good to me. If @mattrpav is OK (as he requested a change), I will merge it.

PR #1238, opened 2024-07-06
[AMQ-9545] Set cache-control to no-store by default for stronger security

kenliao94:
Unspecified Cache-Control HTTP header is vulnerable. Set it to no-store to avoid caching sensitive data for stronger security. It should be the default unless users override it.

Reference: https://www.virtuesecurity.com/kb/cache-controls-explained/


jbonofre (approved):

kenliao94:
Messed up branch history due to outdated branch and wrong operation with git.

PR #1239, opened 2024-07-09
[AMQ-9519] Remove runtime usage of commons-io

mattrpav:


jbonofre:
@mattrpav I consider like a behavior change as we don't check the stream length anymore. It's potentially a security/performance issue: we introduced stream limit to avoid to flood the REST client with long message.
I would rather keep the same behavior, but replace common-io with our own method to define the stream length.

jbonofre (changes requested):
If we want to remove commons-io, we need to provide a private method to compute stream length as this behavior should stay.

mattrpav:
> If we want to remove commons-io, we need to provide a private method to compute stream length as this behavior should stay.

done

jbonofre (approved):
I'm doing the final tests and it looks good to me. Thanks !

PR #1244, opened 2024-07-09
[AMQ-9525] Upgrade Camel 4.4.3

mattrpav:


jbonofre:
I think in the future, we can also focus only on LTS version. It will limit and "guarantee" the Camel updates.

mattrpav:
Right, Camel 4.4.x is an LTS release

PR #1247, opened 2024-07-09
[AMQ-9530]Fix SelectorAwareVirtualTopicInterceptor ClassCastException if next is not Topic.

Nikita-Shupletsov:
`next` is not guaranteed  to be Topic. If there is plugin that injects another `DestinationInterceptor`, the code will fail with `ClassCastException`.

Added a method to correctly unwrap next.

Ran:
* AMQ4899Test.java
* VirtualTopicSelectorTest.java
* JavaVirtualDestTest.java
* SelectorAwareVTThatDropsMessagesWhenNoConsumer.java
* TwoBrokerVirtualTopicSelectorAwareForwardingTest.java
* VirtualConsumerDemandTest.java
* VirtualTopicSelectorWithAnotherInterceptorTest.java
to make sure there is no regression

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/BaseVirtualDestinationFilter.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/BaseVirtualDestinationFilter.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.broker.region.virtual;
+
+import org.apache.activemq.broker.region.BaseDestination;
+import org.apache.activemq.broker.region.Destination;
+import org.apache.activemq.broker.region.DestinationFilter;
+
+import java.util.Optional;
+
+public class BaseVirtualDestinationFilter extends DestinationFilter {
+    
+    public BaseVirtualDestinationFilter(Destination next) {
+        super(next);
+    }
+
+    BaseDestination getBaseDestination(Destination virtualDest) {
+        if (virtualDest instanceof BaseDestination) {
+            return (BaseDestination) virtualDest;
+        } else if (virtualDest instanceof DestinationFilter) {
+            return ((DestinationFilter) virtualDest).getAdaptor(BaseDestination.class);
+        }
+        return null;
+    }
+}
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/MappedQueueFilter.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/MappedQueueFilter.java
@@ -16,12 +16,12 @@
  */
 package org.apache.activemq.broker.region.virtual;
 
+import java.util.Optional;
 import java.util.Set;
 
 import org.apache.activemq.broker.ConnectionContext;
 import org.apache.activemq.broker.region.BaseDestination;
 import org.apache.activemq.broker.region.Destination;
-import org.apache.activemq.broker.region.DestinationFilter;
 import org.apache.activemq.broker.region.IndirectMessageReference;
 import org.apache.activemq.broker.region.RegionBroker;
 import org.apache.activemq.broker.region.Subscription;
@@ -34,7 +34,7 @@ import org.apache.activemq.util.SubscriptionKey;
  * Creates a mapped Queue that can recover messages from subscription recovery
  * policy of its Virtual Topic.
  */
-public class MappedQueueFilter extends DestinationFilter {
+public class MappedQueueFilter extends BaseVirtualDestinationFilter {
 
     private final ActiveMQDestination virtualDestination;
 
@@ -85,15 +85,6 @@ public class MappedQueueFilter extends DestinationFilter {
                 }
             }
         }
-    }
-
-    private BaseDestination getBaseDestination(Destination virtualDest) {
-        if (virtualDest instanceof BaseDestination) {
-            return (BaseDestination) virtualDest;
-        } else if (virtualDest instanceof DestinationFilter) {
-            return ((DestinationFilter) virtualDest).getAdaptor(BaseDestination.class);
-        }
-        return null;
     }
 
     @Override
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/SelectorAwareVirtualTopicInterceptor.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/SelectorAwareVirtualTopicInterceptor.java
@@ -17,9 +17,9 @@
 package org.apache.activemq.broker.region.virtual;
 
 import org.apache.activemq.broker.Broker;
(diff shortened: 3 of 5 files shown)


mattrpav:
What is the use case for this?

Nikita-Shupletsov:
> What is the use case for this?

you mean the use case of having a DestinationFilter after SelectorAwareVirtualTopicInterceptor?

I my case I add an interceptor to all topics. and that interceptor coexists with the rest of interceptors. and some topics can be vistual topics. it's technically possible on the plugin side to check if there is a virtual topic interceptor and not add my interceptor. But I checked other interceptors([this one](https://github.com/apache/activemq/blob/main/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/MappedQueueFilter.java#L68) for example) and they handle this situation correctly, so I thought it would be a positive change in general. In case people decide to do something similar in the future, they will not need to work around the edge case.

kenliao94 (approved):

mattrpav:
A couple of points:

1. This change needs to be a JIRA
2. This needs a unit test with an example configured w/ another DestinationFilter on top of Virtual Topic
3. If we are going to fix it, we should adapt all the Virtual Destinations to use the same set of methods.

Nikita-Shupletsov:
>     1. This change needs to be a JIRA
Done. https://issues.apache.org/jira/browse/AMQ-9530
>     2. This needs a unit test with an example configured w/ another DestinationFilter on top of Virtual Topic
Updated `VirtualTopicSelectorTest,` as it's basically tests `SelectorAwareVirtualTopicInterceptor`. Also used it to initially reproduce the issue(see the ticket).
>     3. If we are going to fix it, we should adapt all the Virtual Destinations to use the same set of methods.
There are no issues like this in other pars of the Virtual Destination code. only in `SelectorAwareVirtualTopicInterceptor`

mattrpav:
> There are no issues like this in other pars of the Virtual Destination code. only in `SelectorAwareVirtualTopicInterceptor`

MappedQueueFilter has a nearly identical method:

ref: https://github.com/apache/activemq/blob/5309352e12c1d5f05c027331f49cc1caa6a62f2f/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/MappedQueueFilter.java#L90

Nikita-Shupletsov:
> > There are no issues like this in other pars of the Virtual Destination code. only in `SelectorAwareVirtualTopicInterceptor`
> 
> MappedQueueFilter has a nearly identical method:
> 
> ref:
> 
> https://github.com/apache/activemq/blob/5309352e12c1d5f05c027331f49cc1caa6a62f2f/activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/MappedQueueFilter.java#L90

are you proposing to extract this method to a helper class? If I create a new class, let's say VirtualTopicSupport or VistualTopicHelper, will it be okay? or do you have a particular place in mind fir it?

mattrpav:
Yes, a base class that contains the lookup method to unwrap the destination to get to the root (real) destination that can be used by VirtualTopic and MappedQueueFilter.

ie BaseVirtualDestination(Inteceptor|Filter)

Nikita-Shupletsov:
> Yes, a base class that contains the lookup method to unwrap the destination to get to the root (real) destination that can be used by VirtualTopic and MappedQueueFilter.
> 
> ie BaseVirtualDestination(Inteceptor|Filter)

sounds good! done

activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/BaseVirtualDestinationFilter.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.util.Optional;
+
+public class BaseVirtualDestinationFilter extends DestinationFilter {
+    
+    public BaseVirtualDestinationFilter(Destination next) {
+        super(next);
+    }
+
+    Optional<BaseDestination> getBaseDestination(Destination virtualDest) {
```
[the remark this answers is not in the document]
cshannon: I would probably avoid optional if it's a hot spot area of the code and just do a null check.

activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/BaseVirtualDestinationFilter.java
```diff
@@ -0,0 +1,23 @@
+package org.apache.activemq.broker.region.virtual;
```
jbonofre, line 1: The ASF header is missing here.

activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/BaseVirtualDestinationFilter.java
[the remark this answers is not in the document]
jbonofre: I don't think we should use `Optional` here. I would rather check the base object.

jbonofre (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/BaseVirtualDestinationFilter.java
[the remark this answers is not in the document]
Nikita-Shupletsov: Done!

jbonofre (approved):
The change looks good to me now.

@mattrpav as you requested changes, can you please take a new look ?

activemq-unit-tests/src/test/java/org/apache/activemq/broker/virtual/VirtualTopicSelectorTest.java
```diff
@@ -99,7 +103,25 @@ protected BrokerService createBroker() throws Exception {
         virtualTopic.setSelectorAware(true);
         VirtualDestinationInterceptor interceptor = new VirtualDestinationInterceptor();
         interceptor.setVirtualDestinations(new VirtualDestination[]{virtualTopic});
-        broker.setDestinationInterceptors(new DestinationInterceptor[]{interceptor});
+        TestDestinationInterceptor testInterceptor = new TestDestinationInterceptor();
```
mattrpav, line 106: Instead of modifying the existing test, please add a second test to cover the new use case. Changing the test removes the test for the out-of-the-box default behavior vs the custom behavior this fix solves.

mattrpav (changes requested):
The code changes look good.

o

activemq-unit-tests/src/test/java/org/apache/activemq/broker/virtual/VirtualTopicSelectorTest.java
jbonofre: @Nikita-Shupletsov do you mind to add a new test as proposed by @mattrpav ? I would like to include this PR in ActiveMQ 6.1.4 release.
Thanks !

jbonofre: It looks good to me. Thanks !

jbonofre: We could have create an abstract but it's really cosmetic, let's go like this for now.

jbonofre:
FYI, in some contexts, the `VirtualConsumerDemandTest` can be stuck. It's not related to this PR (known issue).

mattrpav:
This type of internal broker plugin and extension points (API/SPI) change should happen in a minor release 6.2.x and not a minor release.

jbonofre:
@mattrpav as it's a bug fix, I think it's ok for have it on all active branches.

activemq-broker/src/main/java/org/apache/activemq/broker/region/virtual/BaseVirtualDestinationFilter.java
```diff
@@ -0,0 +1,39 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.broker.region.virtual;
+
+import org.apache.activemq.broker.region.BaseDestination;
+import org.apache.activemq.broker.region.Destination;
+import org.apache.activemq.broker.region.DestinationFilter;
+
+import java.util.Optional;
```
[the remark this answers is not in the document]
Nikita-Shupletsov: Addressed here: https://github.com/apache/activemq/pull/1351

PR #1251, opened 2024-07-11
[AMQ-9515] Redact JMSText in the HTTP params

kenliao94:
As pointed out in https://issues.apache.org/jira/browse/AMQ-9515 message body is sensitive data and sending a text message using the web console shouldn't log the message content in the audit log.

### DIFF
--- a/activemq-web/src/main/java/org/apache/activemq/web/AuditFilter.java
+++ b/activemq-web/src/main/java/org/apache/activemq/web/AuditFilter.java
@@ -24,11 +24,13 @@ import org.slf4j.LoggerFactory;
 import jakarta.servlet.*;
 import jakarta.servlet.http.HttpServletRequest;
 import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
 
 public class AuditFilter implements Filter {
 
     private static final Logger LOG = LoggerFactory.getLogger("org.apache.activemq.audit");
-
+    private static final String[] SENSITIVE_PARAM_KEYS = {"JMSText"};
     private boolean audit;
     private AuditLogService auditLog;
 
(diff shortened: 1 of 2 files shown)


mattrpav:
I'm mostly -1 on this change.

I think this should be handled by configuring a filter at the logging framework level.

Some users may want body redacted, others certain header or properties fields.. or even just specific fields within the body.

Logging frameworks support configurable filters for this, I think this can be handled as a FAQ in ActiveMQ about how to add a logging filter.

kenliao94:
> I'm mostly -1 on this change.
> 
> I think this should be handled by configuring a filter at the logging framework level.
> 
> Some users may want body redacted, others certain header or properties fields.. or even just specific fields within the body.
> 
> Logging frameworks support configurable filters for this, I think this can be handled as a FAQ in ActiveMQ about how to add a logging filter.

Hi Matt, thanks for the feedback.

That's a good idea. Let me look into it today. Instead of making this change, we will ship a default configuration for audit log to mask that information and add a FAQ in ActiveMQ to instruct users on how to add their own or override the default. Does it sound good to you?

mattrpav:
> we will ship a default configuration for audit log ..

If by 'we' you mean your organization, good to go =)

If by 'we' you mean Apache, then I don't see us changing the default Apache distribution config for this.

kenliao94:
> > we will ship a default configuration for audit log ..
> 
> If by 'we' you mean your organization, good to go =)
> 
> If by 'we' you mean Apache, then I don't see us changing the default Apache distribution config for this.

I can see your concern. By "we" I meant developers in this community :) When developers deploy ActiveMQ for their customers, their customers may want sensitive data to be redacted. For instance, in `AuditLogEntry.java` fields that are annotated by `@Sensitive` (such as password) are redacted in the audit.log. So in my opinion, it is probably better to have it default redacted (possibly other fields in the HTTP parameters as well) to avoid potential compliance issues. I will focus on documenting the instructions to configure the log filter nevertheless.

PR #1256, opened 2024-07-12
[AMQ-9533] Add Backup command that uses activemq-protobuf

mattrpav:
eventually deprecate the 'export' command


mattrpav:
replaced by #1258

PR #1258, opened 2024-07-12
[AMQ-9533] Add offset support to Backup command

mattrpav:


jbonofre (changes requested):
Please, don't mix several fixes/changes (unrelated) in the same PR.
For instance, I would have move the hawtbuf dependency issue (on CLI) in a separate PR.
If you don't mind I will split in different PRs (keeping you as author).

activemq-broker/src/main/java/org/apache/activemq/store/memory/MemoryMessageStore.java
```diff
@@ -130,6 +130,32 @@ public void recoverNextMessages(int maxReturned, MessageRecoveryListener listene
         }
     }
 
+    @Override
+    public void recoverNextMessages(int offset, int maxReturned, MessageRecoveryListener listener) throws Exception {
+        synchronized (messageTable) {
+            boolean pastLackBatch = lastBatchId == null;
```
cshannon, line 136: looks like a small typo here

mattrpav: Misnamed variable aligns other method usage

PR #1261, opened 2024-07-13
AMQ-9531: Add hawtbuf-proto dependency for CLI

jbonofre:
Author: Matt Pavlovich <mattrpav@gmail.com>


mattrpav:
@jbonofre this change is incomplete and should be closed in favor of [#1257]. The Export command needed a patch b/c the introspection wasn't working with the method overload.

mattrpav (changes requested):
Use PR [#1257] instead

PR #1266, opened 2024-07-21
[AMQ-9537] Update user-guide.html

DressedUpZebra:
Updated things I've noticed that did not work when following the guide:

- Windows example of starting the broker did not work. The broker on windows can be started with activemq.bat start
- Removed the mentions of ANT in the guide, the ANT examples were removed in 5.16.0
- Updated instructions on how to start the broker on windows with web apps.
- Starting broker with the demo apps did not work, updated the configuration to use the config the broker ships with, ~saved the 5.x demo config for reference.~ 
- Updated in the guide where to change the default credentials.


assembly/src/release/docs/user-guide.html
```diff
@@ -98,7 +98,7 @@ <h2>Access the web demos</h2>
 </p>
 <pre>
     cd bin
-    activemq.bat console xbean:examples/conf/activemq-demo.xml
+    activemq.bat start xbean:../examples/conf/activemq-demo.xml
```
jbonofre, line 101: NB: `activemq console` is actually valid (at least for Unix shell script). `activemq.bat` doesn't really take any action arg. Maybe we should clean the arg.

assembly/src/release/examples/conf/jetty-demo-5.x.xml
```diff
@@ -0,0 +1,180 @@
...
+        "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+        implied. See the License for the specific language governing permissions and
+        limitations under the License.
+    -->
+    <!--
+        An embedded servlet engine for serving up the Admin consoles, REST and Ajax APIs and
+        some demos Include this file in your configuration to enable ActiveMQ web components
+        e.g. <import resource="jetty.xml"/>
+    -->
+    <beans xmlns="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```
jbonofre, line 20: I would completely remove this file as it's ActiveMQ 5.x specific (not 6).

jbonofre (changes requested):

assembly/src/release/docs/user-guide.html
AlexejTimonin: Thank you for the review!
You are correct that the activemq.bat file does not take any arguments directly, but it does pass the provided arguments to the jar using %*. For reference, see this line: 
https://github.com/apache/activemq/blob/6b6f5fbe24550452cd09aab9c4d7094cf782a04f/assembly/src/release/bin/activemq.bat#L99 
Running ``activemq.bat`` without ``start`` argument will not start the broker.

jbonofre (approved):

mattrpav:
@jbonofre the jetty-5.x.* files shouldn't have been deleted. These demonstrate how configure Jetty to use a separate JAAS realm, which was the default behavior in ActiveMQ 5.x.

jbonofre:
But it's ActiveMQ 6 and I tested: it doesn't work anymore with Jetty 11.

mattrpav:
Ah, ok, That's fine then, thanks

PR #1272, opened 2024-07-23
Code Janitor task: Replace use of deprecated Class.newInstance()

AlexejTimonin:
This is a no-jira task, just some cleaning in activemq-unit-tests to minimize deprecated warnings when building.


tabish121:
If you need to open five "no jira" PRs for cleaning up unit tests, then you probably need a JIRA for cleaning up unit tests.

AlexejTimonin:
> If you need to open five "no jira" PRs for cleaning up unit tests, then you probably need a JIRA for cleaning up unit tests.

Fair enough. At first it was going to be one PR, but I split it up to keep changes related, hence 5 PRs. I will create a ticket.

PR #1275, opened 2024-07-25
AMQ-9541 limit expensive queries on the database side

jeanouii:
…ere there is a setMaxRows on the statement.


activemq-jdbc-store/src/main/java/org/apache/activemq/store/jdbc/adapter/DefaultJDBCAdapter.java
```diff
@@ -620,12 +618,12 @@ public void doRecoverNextMessages(TransactionContext c, ActiveMQDestination dest
...
     public void doRecoverNextMessagesWithPriority(TransactionContext c, ActiveMQDestination destination, String clientId,
-            String subscriptionName, long seq, long priority, int maxReturned, JDBCMessageRecoveryListener listener) throws Exception {
+                                                  String subscriptionName, long seq, long priority, int maxReturned, JDBCMessageRecoveryListener listener) throws Exception {
 
         PreparedStatement s = null;
         ResultSet rs = null;
         try {
-            s = c.getConnection().prepareStatement(this.statements.getFindDurableSubMessagesByPriorityStatement());
+            s = c.getConnection().prepareStatement(this.limitQuery(this.statements.getFindDurableSubMessagesByPriorityStatement()));
             s.setMaxRows(Math.min(maxReturned * 2, maxRows));
```
jeanouii, line 627: I'm not sure why there is "* 2" for the setMaxRows

activemq-jdbc-store/src/main/java/org/apache/activemq/store/jdbc/adapter/DefaultJDBCAdapter.java
```diff
@@ -586,12 +584,12 @@ public void doRecoverSubscription(TransactionContext c, ActiveMQDestination dest
...
     public void doRecoverNextMessages(TransactionContext c, ActiveMQDestination destination, String clientId,
-            String subscriptionName, long seq, long priority, int maxReturned, JDBCMessageRecoveryListener listener) throws Exception {
+                                      String subscriptionName, long seq, long priority, int maxReturned, JDBCMessageRecoveryListener listener) throws Exception {
 
         PreparedStatement s = null;
         ResultSet rs = null;
         try {
-            s = c.getConnection().prepareStatement(this.statements.getFindDurableSubMessagesStatement());
+            s = c.getConnection().prepareStatement(this.limitQuery(this.statements.getFindDurableSubMessagesStatement()));
             s.setMaxRows(Math.min(maxReturned * 2, maxRows));
```
jeanouii, line 593: I'm not sure why there is "* 2" for the setMaxRows

activemq-jdbc-store/src/main/java/org/apache/activemq/store/jdbc/adapter/DefaultJDBCAdapter.java
AlexejTimonin: Not an expert, but I did some git blame digging, the * 2 got introduced in https://issues.apache.org/jira/browse/AMQ-2980, to take into account the messages from low and high priority queues, to ensure that none get lost. Down the road this happened ``java.sql.SQLException: setMaxRows() out of range. 50871918 > 50000000.`` (https://issues.apache.org/jira/browse/AMQ-1870). The solution to this was to pick the lowest of two values ``(maxReturner * 2) and maxRows``, to ensure such exception does not happen.

kenliao94: Sorry, not following this logic here. Isn't this.limitQuery a no-ops? It just returns the original query (line 1188)

kenliao94:
@jeanouii i Sorry, not following this logic here. Isn't this.limitQuery a no-ops? It just returns the original query (line 1188)

kenliao94:
^ Ah I guess this is the default. It will be over-written in the class that extends it.

kenliao94:
LGTM, however, this PR should be open against the main branch not a release branch. If you want the fix to only be in a release branch, it can be backported later.

jeanouii:
@kenliao94 reviving this PR a little bit.
I'm currently working on a deadlock issue with a fix.
If you are still willing to merge it, I can rebase and create a new PR against 6.1.x and main branch (6.2.x).

I can also backport to 5.19.x as it's still maintained

PR #1276, opened 2024-07-27
Initial declaration of OAuth and OIDC plugin and validator with unimp…

samirboudjebla:
Initial declaration of the dependencies, java classes and the corresponding methods to implement the following jira - https://issues.apache.org/jira/browse/AMQ-9400

Methods are not implemented yet.

### DIFF
--- a/activemq-web-console/src/main/webapp/WEB-INF/activemq.xml
+++ b/activemq-web-console/src/main/webapp/WEB-INF/activemq.xml
@@ -34,6 +34,9 @@
     <transportConnectors>
       <transportConnector name="openwire" uri="tcp://localhost:61616" />
       <transportConnector name="stomp" uri="stomp://localhost:61613" />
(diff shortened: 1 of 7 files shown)


activemq-web-console/src/main/webapp/WEB-INF/activemq.xml
```diff
@@ -34,6 +34,9 @@
     <transportConnectors>
       <transportConnector name="openwire" uri="tcp://localhost:61616" />
       <transportConnector name="stomp" uri="stomp://localhost:61613" />
+      <!-- Add Jetty Transport Connector for Web Console -->
```
jbonofre, line 37: That's not correct: the jetty transport is not for the webconsole. Jetty transport is for http transport connect.
The WebConsole is managed outside of the broker by inclusion of `jetty.xml`.

jbonofre (changes requested):

PR #1278, opened 2024-08-08
AMQ-9547 - Remove setLength() and usage from RecoverableRandomAccessFile

cshannon:
This method always throws an exception so it should be removed and no longer used. Places in `PageFile` that called the method have been updated to either remove the usage or throw an error.

Update: I was able to add a test in `PageFileTest` to verify there's no more exception if the recovery file is missing as the page file no longer calls setLength() and will continue


mattrpav (approved):
LGTM

PR #1282, opened 2024-08-17
AMQ-9550 for loop can be replaced with enhanced for connection manager adapter

levrun:
What problems does this PR solve?
```
Iteration over collections or arrays  can be replaced with an enhanced for loop (foreach iteration syntax). 
See more details why : https://docs.oracle.com/javase/8/docs/technotes/guides/language/foreach.html
```

Why is it beneficial to merge into ActiveMQ?
```
The for-each construct gets rid of the clutter and the opportunity for error.
When should you use the for-each loop? Any time you can. It really beautifies your code.
```

How do you make sure this PR is well tested?
run tests locally

### DIFF
--- a/activemq-ra/src/test/java/org/apache/activemq/ra/ConnectionManagerAdapter.java
+++ b/activemq-ra/src/test/java/org/apache/activemq/ra/ConnectionManagerAdapter.java
@@ -17,7 +17,6 @@
 package org.apache.activemq.ra;
 
 import java.util.ArrayList;
-import java.util.Iterator;
 
 import jakarta.resource.ResourceException;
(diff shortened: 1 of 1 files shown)


kenliao94 (approved):

kenliao94:
@jbonofre Hey JB, could you take a quick look at this PR? Thanks.

mattrpav:
The change is good -- I think the title could be clarified this is just a simple modernization and not a fundamental change in loop algorithm.

Something like:
activemq-ra: modernize loops to remove iterator

kenliao94:
@mattrpav  Good call out.

PR #1285, opened 2024-08-20
AMQ-9552: Add metric lastMessageTimestamp to StatisticsBroker based o…

ggkochanski:
Add metric lastMessageTimestamp to StatisticsBroker based on modification time of metric enqueues:

```
// NOTICE: Client-side, you may get the broker "now" Timestamp by msg.getJMSTimestamp()
// This allows for calculating inactivity.
statsMessage.setLong("lastMessageTimestamp", stats.getEnqueues().getLastSampleTime()); 

```

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/plugin/StatisticsBroker.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/plugin/StatisticsBroker.java
@@ -160,6 +160,9 @@ public class StatisticsBroker extends BrokerFilter {
                                 tempFirstMessage.clear();
                             }
                         }
+                        // NOTICE: Client-side, you may get the broker "now" Timestamp by msg.getJMSTimestamp()
+                        // This allows for calculating inactivity.
+                        statsMessage.setLong("lastMessageTimestamp", stats.getEnqueues().getLastSampleTime());
                         statsMessage.setJMSCorrelationID(messageSend.getCorrelationId());
                         sendStats(producerExchange.getConnectionContext(), statsMessage, replyTo);
                     }
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/plugin/BrokerStatisticsPluginTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/plugin/BrokerStatisticsPluginTest.java
@@ -218,6 +218,7 @@ public class BrokerStatisticsPluginTest extends TestCase{
         assertEquals(1, reply.getLong("size"));
         assertTrue(reply.getJMSTimestamp() > 0);
         assertTrue(reply.getLong("firstMessageTimestamp") > 0);
+        assertTrue(reply.getLong("lastMessageTimestamp") >= msg.getJMSTimestamp());
         // Assert that we got the brokerInTime for the first message in queue as value of key "firstMessageTimestamp"
         assertTrue(System.currentTimeMillis() >= reply.getLong("firstMessageTimestamp"));
         assertEquals(Message.DEFAULT_PRIORITY, reply.getJMSPriority());


ggkochanski:
Some examples per threshold of this metric from real life implementation:

image

activemq-client/src/main/java/org/apache/activemq/management/StatisticImpl.java
```diff
@@ -80,6 +80,9 @@ public synchronized long getLastSampleTime() {
         return this.lastSampleTime;
     }
 
+    public synchronized long getLastSampleTimeOrStartTime(){
+        return lastSampleTime == 0 ? startTime : lastSampleTime;
```
kenliao94, line 84: An alternative approach to setting the lastSampleTime to 0, we can keep the `this.lastSampleTime = this.startTime;` as originally implemented. But then we add an additional field in StatisticImpl class like "hasUpdated" to indicate if the lastSampleTime has been updated. In that way, it's up to the caller to decide what to make use of this "hasUpdated" field. I.E write 0, or not set at all ... etc. 

The advantage of avoiding giving special meaning to 0 is:
1. Not explicit in semantic and make the code a bit harder to follow (why 0?)
2. What if 0 means some other things in downstream code
3. More flexible and clear on the caller, it calls the object to see if the lastSampleTime has been updated since restart, rather than just set the value to whatever `stats.getEnqueues().getLastSampleTime()` returns

ggkochanski: wow, that's clever 😊 

it's much better,
we may even skip generation of
`statsMessage.setLong("lastMessageTimestamp", stats.getEnqueues().getLastSampleTime())`
when it's not "hasUpdated" 😋

activemq-broker/src/main/java/org/apache/activemq/plugin/StatisticsBroker.java
```diff
@@ -160,6 +160,11 @@ else if (dest instanceof Topic) {
                                 tempFirstMessage.clear();
                             }
                         }
+                        if ( stats.getEnqueues().hasUpdated() ) {
+                            // NOTICE: Client-side, you may get the broker "now" Timestamp by msg.getJMSTimestamp()
+                            // This allows for calculating inactivity.
+                            statsMessage.setLong("lastMessageTimestamp", stats.getEnqueues().getLastSampleTime());
```
ggkochanski, line 166: @mattrpav I can see your change is merged, so I'm refreshing the pull request.

I understand the name unification motivation. Should I change name **lastMessageTimestamp** into **lastEnqueTimestamp** here?

mattrpav: @ggkochanski The field naming is a broader conversation. The full scope is not one field as you have in this PR. We currently have a number of metrics (and config params) that are not fully aligned in naming convention, so I would press for an organized approach for these new metrics.

Related JIRAs:
ref: https://issues.apache.org/jira/browse/AMQ-4358
ref: https://issues.apache.org/jira/browse/AMQ-5497

ggkochanski: What is the meaning LastDequeueTime and FirstDequeueTime? I believe one of them cannot be determined (if **the last** means the newest enqueued message then it is not dequeued yet)

@mattrpav could you explain in details how you imagine this new approach to metrics based on some example preferably?

Maybe I try to use statistics in the wrong way. What I need to know to assess the condition of mq system is: enqueue count and last update timestamp, dequeue count and last update, last process time (iso "averageEnqueueTime") and the age of the oldest message in queue. What do you suggest me?

mattrpav: @ggkochanski yeah, the challenge is settling on the terminology for the metric values. 

For example, does 'last' mean last-as-in-time, or last-as-in-last-message-in-the-queue (most recently added)

mattrpav:
The advancedMessageFlow statistics are implemented and released in 5.19.x (and soon 6.2.0) 

Please review the added metrics: https://github.com/apache/activemq/blob/main/activemq-client/src/main/java/org/apache/activemq/management/MessageFlowStatsImpl.java#L21

I believe this PR is addressed with the _enqueuedMessageBrokerInTime_ or _enqueuedMessageTimestamp_ value.

ggkochanski:
yup, thank you, closing

PR #1286, opened 2024-08-22
[AMQ-9554] Allow session transacted queue browser to view messages that have been dlq'd due to max redeliveries

mattrpav:


cshannon (changes requested):
This doesn't make sense to me and seems to hide the root issue. That method is simply checking if redelivery was already exceeded and handling it if it is. It also checks if the message has been expired. There's no reason to not discard the message on browse if redelivery was exceeded.  Here is where expiratnoi is also done in the same section where we discard on delivery exceeded: https://github.com/apache/activemq/blob/e45ee4aae5b49986750d5a2feb171f36cf432fca/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java#L510 and https://github.com/apache/activemq/blob/e45ee4aae5b49986750d5a2feb171f36cf432fca/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java#L1432

And note we specifically check if we are a browser in this case: https://github.com/apache/activemq/blob/e45ee4aae5b49986750d5a2feb171f36cf432fca/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java#L1423

Is the actual problem that browsing is incrementing the counter when dispatched to a browser and shouldn't be?

cshannon:
so re-reading what you wrote, you said it applies to a DLQ which would be an edge case but your change seems like it changes the behavior for every consumer and every queue and not just DLQs which we would not want.

If the goal is to simply be able to browse the DLQ and preventing not being able to read messages that were previously exceeding the redelivery policy then for your DLQ consumer can you simply disable the redelivery policy? Seems like a simple solution.

mattrpav:
@cshannon the updated test has the full repro scenario.

The issue comes down to:

1. Should a browser send a poison ack if the message redeliveryCounter is on the edge of max redeliveries? 

Currently, ActiveMQMessageConsumer _does_ send a poison ack. The problem with that is it creates two messages, as the original stays on the source queue (dlq or otherwise) AND a message to a DLQ (or .DLQ.DLQ when browsing a .DLQ).

My argument is that browser should _not_ send a poison ack if the max redelivery count is reached.

cshannon:
If you don't want to create the poison ack then couldn't you simply disable the redelivery policy when browsing the DLQ since it's an edge case? This seems like a client side config issue because you are setting that policy....so just don't set it when browsing the DLQ.

Consumers (including browsers) are designed to process the messages and check things like the redelivery counter or even expiration.  A browser should not increment the redelivery counter, but I don't see how checking that counter and sending a poison ack is any different than checking expiration and expiring a message which it will do.

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/QueueBrowsingTest.java
```diff
@@ -217,4 +230,177 @@ public void testMemoryLimit() throws Exception {
...
+
+        Thread browseThread = new Thread() {
+            public void run() {
+                
+                QueueBrowser browser = null;
+                try {
+                    browser = browseSession.createBrowser(queueDLQ);
+                    Enumeration<?> enumeration = browser.getEnumeration();
+                    while (enumeration.hasMoreElements()) {
+                        Message message = (Message)enumeration.nextElement();
```
mattrpav, line 359: BUG: This blocks indefinitely

mattrpav:
Planned fix path forward:

1. Apply a new redeliveryPolicy option: queueBrowserIgnored = false (default behavior today)
2. In 7.x set queueBrowserIgnored = true (more consistent behavior for out-of-the-box JMS browse tools)

mattrpav:
> Is the actual problem that browsing is incrementing the counter when dispatched to a browser and shouldn't be?

Yes! With the caveat that it is *only* when browsing in a transacted session. In AUTO_ACK mode this behavior does not occur.

PR #1288, opened 2024-08-27
[AMQ-9548] Add attributes to display total number of queues and topics

kenliao94:
Currently, to get the total number of queues and topics on the broker using JMX, one needs to query the attribute values of Queues and Topics on the broker MBean then post-process it (finding the length of returned array) to get that data. This can be resource intensive (unnecessary copy of the data) on the customer application using JMX or Jolokia if user has tens or thousands of destinations and all the user wants is the number. Hence this PR added two attributes so customer application can just query the total number without getting the list of topics and queues first.

This will be beneficial for users with high number of destinations. They can now monitor the total number of queues and topics in a more performant way.

- Added unit test in the PR
- Tested with Jolokia (cross-checked with the web console) [see attachment below]
- Tested with Jconsole (cross-checked with the web console) [see attachment below]

image
image
image


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
```diff
@@ -280,11 +280,21 @@ public ObjectName[] getTopics() {
...
+    }
+
     @Override
     public ObjectName[] getQueues() {
         return safeGetBroker().getQueuesNonSuppressed();
     }
 
+    @Override
+    public int getTotalQueuesCount() {
+        return getQueues().length;
```
mattrpav, line 295: This still calls a method that creates a new array of objects. Why not just return the value of the size from the map?

mattrpav (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
kenliao94: Good callout, I believe the new array of objects you are referring to is the one created in `onlyNonSuppressed`  which copy over the non suppressed queues or topics already maintained in the `ManagedRegionBroker`.  Are you saying I should just simply return the size of `private final Map<ObjectName, DestinationView> queues = new ConcurrentHashMap<>();` in `ManagedRegionBroker`?

My concern is that it will not agree with the queue or topic list in the `queues` and `topics` MBean attribute, and those are non suppressed which should be the correct list.

The optimization done at this PR is mainly for user application. Because they will need to read that ObjectName array into their memory space even all they need is the count of `queues` or `topics`. So if this PR is merged as it is, the user application memory space will not be bloated (even tho as your pointed out, on the broker side, it will still need to create an array of those objects on the broker process memory space).

mattrpav: I'm saying the suppression info shouldn't be considered at all. If there are 128 queues, and 28 suppressed, the user is still going to want to know there are 128 queues and not 100.

jbonofre: I disagree @mattrpav, I think this method should be consistent with the `getQueues()` that uses `getQueuesNonSuppressed()`.

So, I think what @kenliao94 does makes sense.

I'm +1 with this change.

jbonofre (approved):

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
mattrpav: @jbonofre none of the other "Total*" metrics take in account the suppression, why would we start with this one?

mattrpav: The current implementation is a -1 for me. There is _zero_ reason to allocate an array, and then shove a bunch of objects into it just to get a count.

Proposed path forward:

The total queue count should simply come from the queue region and do a quick size on an existing collection.

jbonofre: What about:

```
    public ObjectName[] getQueues() {
        return safeGetBroker().getQueuesNonSuppressed();
    }
```

?

mattrpav: That creates an array of ObjectName[] unnecessarily. 

This is a zero allocation operation:

```
managedBroker.getQueueRegion().getDestinationMap().size()
```

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerViewMBean.java
```diff
@@ -179,9 +179,17 @@ public interface BrokerViewMBean extends Service {
...
+    int getTotalTopicsCount();
+
+
     @MBeanInfo("Standard Queues containing AIE messages.")
     ObjectName[] getQueues();
 
+    @MBeanInfo("Total number of queues")
+    int getTotalQueuesCount();
+
+
```
mattrpav, line 192: This should add getTotalTempQueueCount and getTotalTempTopicCount while we are here

mattrpav (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
jbonofre: I got your point, but it looks more "logic" to use the existing method to me. The `ObjectName[]` is not very costly.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerViewMBean.java
jbonofre: Can you elaborate ?

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
jbonofre: Just to explain my thought: currently, an user who wants to get the total number of queues gonna use `getQueues()` and count himself. So, the rationale of this change is to provide a convenient to directly provide the queues count.

It seems "logic" to me to use the same method as the user would use to do the count.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerViewMBean.java
mattrpav: There are 4 total sets of destinations -- Queues, Topics, TempQueues and TempTopics. If we are adding the metric for TotalQueueCount and TotalTopicCount, we should also add TotalTempQueueCount and TotalTempTopicCount.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
mattrpav: How is this logic? This is simply walking pointer references to get a size off a collection.

```
managedBroker.getQueueRegion().getDestinationMap().size()
```

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerViewMBean.java
jbonofre: Oh you mean adding `getTotalTempQueueCount()` and `getTotalTempTopicCount()`. Yeah, agree, it makes sense.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
jbonofre: The logic is the consistency between `getQueues()` and `getQueuesCount()` operations. An user might be confused if he counts with `getQueues()` and compare with `getQueuesCount()`: if he has different numbers, he would be confused.

mattrpav: Right! That is exactly my point. The getQueues pulls in what is _registered_ in JMX, not what is present in the broker. JMX suppression is to prevent overhead caused by registering objects into JMX -- specifically for environments with _lots_ of queues and queues that are one-time-use queues.

If a user enable suppression for performance reasons, they still need base metrics. The BrokerView metrics should be consistent and provide -- what is the in broker?

I think the confusion is around what does the getQueues() method mean? getQueues() _not_ "What queues are in the broker?" -- getQueues() is "what queues are _registered in JMX_?"

jbonofre: That's also my point: as `getQueues()` MBean operation returns the JMX registered queues, the `getQueuesCount()` should also return the JMX registered queues count.

mattrpav: @jbonofre I disagree. The behavior you are describing would be inconsistent with how other metrics at the broker-level operate. Specifically, the connection metrics.

jbonofre: I agree it's "different" compare to the behavior of other MBeans (like connection), BUT it's consistent inside this specific MBean.

mattrpav: getQueues() <-- returns ObjectName[], an array of JMX ObjectNames for queues that are registered, this isn't _all queues in the broker_. If anything, this method is mislabeled.

Use cases:

Use Case 1: 128 queues, no jmx mbean suppression
Use Case 2: 128 queues, 128 queues suppressed
Use Case 3: 128 queues, 28 queues suppressed

Current proposal:

| Use Case | # queues | # suppressed | getQueues() result | getTotalQueueCount() |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| 1 | 128 | 0 | 128 ObjectName[] | 128 |
| 2 | 128 | 128 | 0 ObjectName[] | 0 |
| 3 | 128 | 28 | 28 ObjectName[] | 28 |

jbonofre: I'm sorry, but using my user hat, I still think the current proposal makes more sense to me 😄 

@cshannon as we need an arbiter here, what do you think ? 
@kenliao94 what's your thoughts about that ?

jbonofre: Maybe it's more a matter of operation name. Both operations could make sense (depending of the user cases).

mattrpav: > Maybe it's more a matter of operation name. Both operations could make sense (depending of the user cases).

I'd be fine with adding the metric, while not allocating the array to do it ;-)

getRegisteredQueueCount()  <-- returns ObjectName[] length

```
public int getRegisteredQueueCount() {
   return managedBroker.getQueueViews().size()
```

jbonofre: It looks like a good consensus to me 😄 

@kenliao94 thoughts ?

mattrpav: Can I be greedy and request the getTotalQueueCount() also be added?

cshannon:
@jbonofre and @mattrpav -

I took a look at this and I agree with Matt for all the points he made. The suppression configuration here is because of performance reasons. You are suppressing the returned queue metrics over JMX to not overload things. That doesn't mean the queues don't exist or are not valid so I don't think it makes sense to hide suppressed queues from the metrics. On the other hand I get JB's point about it could be a little confusing if getQueues() is returning a different number because of the suppression.

So if I had to pick one I would agree with @mattrpav and say we should return total queues. However, why not just provide both and call it a day? Just provide both a total count for the broker of the over all queue count and also a suppressed count metric.

We could have something like getTotalQueueCount() and getSupressedQueueCount() or something like that. The total queues just can just use the map size with is a fast constant operation as @mattrpav pointed out. For suppressed queues you could iterate over the queues in the map and count them to avoid object allocation on the broker side, or we could just keep a counter that keeps a running total on queue adds/deletes.

cshannon:
Also the naming doesn't matter, can pick whatever makes sense. The point is just provide both metrics as they could both be useful but make sure we compute them without a bunch of memory allocation

jbonofre:
Ok fair enough. I propose to add the two operations as I'm sure users will need both.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
kenliao94: Thanks for the input @mattrpav and @jbonofre  I see the points of having different names to avoid confusion. I will rename, avoid allocating an array and add the new count metrics.

kenliao94: Done.

jbonofre (approved):
New MBean operation names look good to me. @mattrpav do you mind to take a new look and approve the PR ?

kenliao94:
> New MBean operation names look good to me. @mattrpav do you mind to take a new look and approve the PR ?

Hi @mattrpav, do you have any more concerns about this PR?

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerViewMBean.java
kenliao94: Btw, this is already resolved :)

activemq-unit-tests/src/test/java/org/apache/activemq/broker/jmx/BrokerViewTest.java
```diff
@@ -0,0 +1,68 @@
...
+        tempTopicProducer.send(tempTopicProducerSession.createTextMessage("testMessage"));
+
+        assertTrue(Wait.waitFor(() -> (brokerService.getAdminView()) != null));
+        final BrokerView view = brokerService.getAdminView();
+        assert(view.getTotalTopicsCount() == 11); // Including advisory topics
+        assert(view.getTotalNonSuppressedTopicsCount() == 11);
+        assert(view.getTotalTemporaryTopicsCount() == 1);
+        assert(view.getTotalQueuesCount() == 1);
+        assert(view.getTotalNonSuppressedQueuesCount() == 1);
+        assert(view.getTotalTemporaryQueuesCount() == 1);
```
mattrpav, line 66: 1. Please use the more appropriate assert* methods. 

2. These tests are dodgy. The value expect is 1 and the same across the different types. 

3. Multiple commits need to be squashed for this change

mattrpav (changes requested):

activemq-unit-tests/src/test/java/org/apache/activemq/broker/jmx/BrokerViewTest.java
cshannon: I'll just point out that squashing commits is fine but it is also not necessary. I know I've mentioned before but I prefer it if people do not force push and just push new commits so you can see the difference easier between versions. It's easy to do a squash merge several commits at the end into 1 commit when merging back to main.

kenliao94: Done. Thanks for the feedback @mattrpav

kenliao94:
Hey @jbonofre do you think we should put this in 6.1.5?

cshannon:
> Hey @jbonofre do you think we should put this in 6.1.5?

No, we should not be adding changes like this to a bug fix release. Releases like 6.1.5 or 6.1.6 should just be bug fixes, security fixes, minor dependency updates, etc. 

There was a discussion started on SEMVER on the dev list that we probably need to get back to so that it's more clear what we change in different versions.

If new features are being developed quickly and we end up with lots of stuff to release we can always quickly do releases. There doesn't have to be a long wait in between 6.2.0 and 6.3.0 for example if there's a new JMS 2.0 feature that is ready. Artemis already does this, if you look at past Artemis [releases](https://activemq.apache.org/components/artemis/download/past_releases), they release relatively frequently.

jbonofre:
No it should be for
6.2.0, not 6.1.5 as it's not a bug fix.

jbonofre:
This PR looks good to go in 6.2.0 for me.

@mattrpav you still have "change request"  on the PR but I think @kenliao94 addressed your comments. Can you confirm ? Thanks !

mattrpav:
@kenliao94 The changes look good, thanks!  

I want to take one last minute to discuss the naming, b/c these are public interfaces and once go out we are living with them for a long time.

I think using a term that says what 'TotalUnsuppressedQueueCount' _is_ would be better than what it _isn't_.

How about 'TotalManagedQueueCount'? -- I think this terminology better describes what that number means.

kenliao94:
Thanks for the comment. I caught a bug. The TotalManaged counts suppressed one as well. I,E `safeGetBroker().getTopicViews().size()` returns suppressed ones as well
However, when I do more deep dive, It turns out that there is no map that stores managed destinations. The getQueues()  and getTopics()  are doing the filtering everytime it is called because of that. When a MBean is registered, it internally invokes isAllowedToRegister to check if a given MBean can be registered and register it if allowed. But instead of storing different maps categorized by destination, it is a single map of all registered MBeans. Hence, to get the total number of managed queue or topic, we need to iterate a list of queue or topic and invoke isAllowedToRegister again every time (similar to how onlyNonSuppressed work, it is not retrieving a cached value, but iterate over a set and check everytime).

Furthermore, why `safeGetBroker().getTopicViews().size()` didn't work? Because registerDestination didn't check if that ObjectName is allow to register before putting it into the map

```
    protected void registerDestination(ObjectName key, ActiveMQDestination dest, DestinationView view) throws Exception {
        if (dest.isQueue()) {
            if (dest.isTemporary()) {
                temporaryQueues.put(key, view);
            } else {
                queues.put(key, view);
            }
        } else {
            if (dest.isTemporary()) {
                temporaryTopics.put(key, view);
            } else {
                topics.put(key, view);
            }
        }
        try {
            if (AsyncAnnotatedMBean.registerMBean(asyncInvokeService, mbeanTimeout, managementContext, view, key) != null) {
                registeredMBeans.add(key);
            }
        } catch (Throwable e) {
            LOG.warn("Failed to register MBean {}", key);
            LOG.debug("Failure reason: ", e);
        }
    }
```

For now, I agree with your suggestion on exposing the APIs and what do that suppose to mean. I changed the name of those metrics "NonSuppressed" -> "Managed" and added the testcase to test for suppressed destination. The only thing I rollback to, is that I need to first generate the array first and reads the size, the inefficiency you initially pointed out.
```
  @Override
  public int getTotalManagedTopicsCount() {
      return safeGetBroker().getTopicsNonSuppressed().length;
  }
```

That said, I can follow up with another PR that changes the way how the ManagementContext stores the registered destination, such that getTotalManagedQueuesCount and getTotalManagedTopicsCount can be done by reading a map size and it won't be a API / breaking change, so it can be done in a patch version. WDYT? @mattrpav

mattrpav (approved):
LGTM

jbonofre:
That looks good to me and reasonable.

PR #1290, opened 2024-08-28
AMQ-8398 - Fix Stomp to OpenWire UTF-8 translation

cshannon:
This commit fixes conversion of messages that are sent using Stomp when the body contains characters that require 4 bytes for encoding using standard UTF-8.

ActiveMQ and OpenWire currently use a modified UTF-8 encoding that only uses 3 bytes so the conversion previously was breaking because the body was encoded using standard JDK UTF-8 encoding and set directly on an ActiveMQText message which was leading to decoding errors later when the ActiveMQMEssage tried to decode using the modified encoder.

The fix now makes sure to correctly decode the Stomp message back to a String first and set that on the ActiveMQ message so it can be re-encoded correctly. Only the conversion from Stomp -> OpenWire was fixed because the other directionly already worked correctly of Stomp -> OpenWire as it already did this decoding/encoding process in reverse.


cshannon:
Something else is I didn't look to see if the conversion is correct for other protocols such as AMQP or MQTT. We should probably look at that too and apply the same or similar fix if needed. Also, we should check things like Websockets although websockets should just delegate to the same conversion logic I would think for STOMP/MQTT.

tabish121:
Change seems reasonable to me.  The idea that you need to change the openwire bits might be going to far given the expectation is that those bits are using modified UTF8 and has been for ages.  What really matters is the protocols on the periphery which are less likely to be expecting modified UTF8 vs the standard UTF8 encoding like AMQP.

For AMQP the proton-j bits will be doing that work to encode and decode and should only be populating the ActiveMQ message types with the decoded String content on ingress and then encoding the Strings from TextMessage and message properties as they move out of the broker (If the transformer is configured to 'jms').  

For MQTT the protocol is a binary encoding and as far as I'm aware there is nothing internal that's creating TextMessage views of those MQTT messages so that should be safe (probably want to check for some weird option that does though).

tabish121 (approved):

PR #1291, opened 2024-08-31
AMQ-8122 - Fix DataByteArrayInputStreamTest

azotcsit:
As a part of https://github.com/apache/activemq/pull/1284, I re-wrote `DataByteArrayInputStreamTest`. Since we decided to take another route for the fix and closed that PR, I'd like to merge these changes separately.

This PR addresses problems mentioned in https://issues.apache.org/jira/browse/AMQ-8122.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/util/DataByteArrayOutputStream.java
+++ b/activemq-client/src/main/java/org/apache/activemq/util/DataByteArrayOutputStream.java
@@ -210,7 +210,6 @@ public final class DataByteArrayOutputStream extends OutputStream implements Dat
         ensureEnoughBuffer((int)(pos + encodedsize + 2));
         writeShort((int)encodedsize);
 
-        byte[] buffer = new byte[(int)encodedsize];
         MarshallingSupport.writeUTFBytesToBuffer(text, (int) encodedsize, buf, pos);
         pos += encodedsize;
     }
--- a/activemq-client/src/test/java/org/apache/activemq/util/DataByteArrayInputStreamTest.java
+++ b/activemq-client/src/test/java/org/apache/activemq/util/DataByteArrayInputStreamTest.java
@@ -17,64 +17,119 @@
 package org.apache.activemq.util;
 
 import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertThrows;
 
+import java.io.UTFDataFormatException;
 import org.junit.Test;
 
+/**
+ * https://issues.apache.org/jira/browse/AMQ-1911
+ * https://issues.apache.org/jira/browse/AMQ-8122
+ */
 public class DataByteArrayInputStreamTest {
 
-    /**
-     * https://issues.apache.org/activemq/browse/AMQ-1911
-     */
     @Test
-    public void testNonAscii() throws Exception {
-        doMarshallUnMarshallValidation("mei\u00DFen");
-
-        String accumulator = new String();
-
-        int test = 0; // int to get Supplementary chars
-        while(Character.isDefined(test)) {
-            String toTest = String.valueOf((char)test);
-            accumulator += toTest;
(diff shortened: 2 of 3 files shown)


azotcsit:
Even though the build failed, the failure is not related to these changes: https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/PR-1291/2/cloudbees-pipeline-explorer/#line-11324-11325

azotcsit:
@cshannon could you please take a look at this PR when you have a chance.

mattrpav (changes requested):
@azotcsit these are good tests to add in terms of breaking out the 1, 2, 3 and 4-byte ranges.

Would you please add some failure scenario tests to confirm the exception handling? For example, request to encode characters that are out of range.

cshannon:
The extra tests here will be nice to have

azotcsit:
@mattrpav @cshannon 
I tried adding tests for a negative scenario, however, it does not seem to be possible for this particular test/functionality.

Basically, the idea is to generate an invalid String and pass it to `doMarshallUnMarshallValidation` method, so writeUTF/readUTF fails somewhere. However, after quite a bit of trying, I realized that Java itself does not allow you to create a string with invalid UTF characters. Any invalid sequence of bytes is getting replaced or removed (or an exception thrown) based on https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/charset/CodingErrorAction.html (default is REPLACE). 

Here are two more links with details:
- https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/charset/CharsetEncoder.html
- https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/charset/CharsetDecoder.html

Here are a few examples where I had created an invalid string which was automatically corrected by Java:
1. out of range character (0X110000)
```
new String(new byte[]{-12, -112, -128, -128}).getBytes()
[-17, -65, -67, -17, -65, -67, -17, -65, -67, -17, -65, -67]
```

_* every byte replaced by [-17, -65, -67]_

2. invalid character (0xC3 0x28 which is out of range for 2-bytes characters, hence invalid)
```
new String(new byte[]{-61, 40}).getBytes()
[-17, -65, -67, 40]
```

_* -61 was replaced by [-17, -65, -67], 40 remained as is_

3. single high surrogate
```
"\uD83D".getBytes()
[63]
```

_* supplementary characters should be coming as surrogates (pairs of characters called high surrogate [U+D800 - U+DBFF] and low surrogate [U+DC00 to U+DFFF]; single surrogate character is invalid, it was replaced by 63_

4. double high surrogate
```
"\uD83D\uD83D".getBytes()
[63, 63]
```

_* every surrogate character was replaced by 63_

Notes:
- [-17, -65, -67] represent a replacement character U+FFFD. 
- 63 represents another replacement character "?".


Both U+FFFD and "?" are properly handled by the current logic. So, if it is impossible to trigger an exception using `writeUTF` and `readUTF` on `DataByteArrayOutputStream`/`DataByteArrayInputStream` while working with `String` objects. The only way to trigger an exception is to write invalid bytes manually and then try to read them, but I do not feel this test is the right place for doing that. `MarshallingSupportTest`seems to be a better place for covering such a scenario. I can write a test there. Could you please share your thoughts on this?

cshannon:
I think adding a test to MarshallingSupportTest seems ok but I will let @mattrpav since he was the one that requested changes here.

azotcsit:
@mattrpav 

I gave a though to "non-Java applications can send things over the wire" scenario that you had pointed out. I think you're right, and we need to have corresponding tests as a part of `DataByteArrayInputStreamTest`. Could you please take a look at the latest changes.

mattrpav (approved):
LGTM

jbonofre (approved):
LGTM thanks !

PR #1292, opened 2024-09-01
AMQ-8398 - Add missing super.tearDown() call to Stomp NIOSSLLargeMessageTest

azotcsit:
CI build started failing after https://github.com/apache/activemq/pull/1290 had been merged:
* this PR - https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/PR-1290/2/cloudbees-pipeline-explorer/#line-12002
* 5.18.x - https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/activemq-5.18.x/141/cloudbees-pipeline-explorer/#line-12166
* 6.11.x - https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/activemq-6.1.x/14/cloudbees-pipeline-explorer/#line-11963
* main - https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/main/482/cloudbees-pipeline-explorer/#line-11996

However, after some local troubleshooting, I believe the problem is not particularly related to those changes. The root cause seems to be related to a missing tear down logic in one of tests.

### DIFF
--- a/activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/StompNIOSSLLargeMessageTest.java
+++ b/activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/StompNIOSSLLargeMessageTest.java
@@ -103,6 +103,7 @@ public class StompNIOSSLLargeMessageTest extends StompTestSupport {
 
     @Override
     public void tearDown() throws Exception {
+        super.tearDown();
         // unregister Log4J appender
         org.apache.logging.log4j.core.Logger rootLogger = (org.apache.logging.log4j.core.Logger) org.apache.logging.log4j.LogManager.getRootLogger();
         rootLogger.removeAppender(appender);


azotcsit:
Hmm.. PR failed with the same error. This changed helped in my local, but not on CI server.

jbonofre:
@azotcsit I don't think it's related to #1290 : the test is flaky for a while afair.
I can take a look as well.

jbonofre:
The test failure seems indeed related to client still connected:

```

[ERROR] org.apache.activemq.transport.stomp.StompNIOSSLTest.testSend4ByteUtf8StompToStomp -- Time elapsed: 0.123 s <<< FAILURE!
--
  | 11779 | 06:56:59 AM | org.junit.ComparisonFailure:
  | 11780 | 06:56:59 AM | expected:<[!®౩🙂]> but was:<[jakarta.jms.InvalidClientIDException: Broker: localhost - Client: ID:asf928.gq1.ygridcore.net-39889-1725339354177-3:33 already connected from tcp://127.0.0.1:37786
```

Let me try on my machine.

jbonofre:
@azotcsit weird, it works without problem on my machine (even without this fix). I'm trying to increate the maven threads to see if it's race condition.

cshannon:
There is probably just a race condition based on timing, I agree it's not related to the changes I made but an existing issue to resolve with the tests

activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/StompTestSupport.java
```diff
@@ -342,6 +342,8 @@ protected void addTranportConnectors() throws Exception {
             autoNioSslPort = connector.getConnectUri().getPort();
             LOG.debug("Using auto+nio+ssl port " + autoNioSslPort);
         }
+
+        connector.setAllowLinkStealing(true);
```
jbonofre, line 346: That one can help in the case of a race condition because it can reuse the link from another client.

mattrpav: Note: This closes the _other_ client and allows the new connection to use the clientId. 

While this would work, these smells more like a hack vs a fix. Why/Where is the other client connection coming from?

azotcsit: I totally agree that this is a workaround rather than a proper fix. I'm just exploring different options.

> Why/Where is the other client connection coming from?

This is the right question! I'm puzzled by this and I'm not clear how to understand what test/client keeps the connection open as it is not reproducible in my local.

azotcsit: Link stealing did not help:
https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/PR-1292/3/cloudbees-pipeline-explorer/#line-12035

cshannon:
The root cause is dangling sockets that are not being properly closed by some tests and those sockets contain data on the wire, and it seems like on restart the frames get sent along to the broker. I noticed multiple connection attempts being created by previous tests.

I am going to push up a fix in a minute.

cshannon:
Hmm nevermind, something else weird is still going on. I thought I had it fixed but it's not with the NIO ssl tests, even closing sockets is still an issue so I don't think it's that. I'm still trying to figure it out...it's super weird but clearly there's another test interfering because if I run the new test in a loop by itself it doesn't have issues.

cshannon:
This StompNIOSSLTest was breaking every time for me locally and now passes after I pushed this fixed to main: https://github.com/apache/activemq/commit/f5fb6c91b2f302459ab13b247630d4869ed8ce75

The receive for the CONNECTED frame was happening out of order, fixing that and now the test seems to pass reliably.

azotcsit:
Good catch @cshannon ! 

> This StompNIOSSLTest was breaking every time for me locally

It is interesting as the test worked fine in my local.

This is the only change from this PR that still seems to be useful: https://github.com/apache/activemq/pull/1292/files#diff-258971427cd19a158ba8586ba4977966b86d26771749b265355e7781463fc9e8R106. Do you think we can/should merge that?

cshannon:
The call to the teardown for super makes sense, if you want to create a quick PR for that you can or I can make a quick commit, up to you.

azotcsit:
@cshannon 
I updated this PR, so it has the aforementioned change only. Please, take a look when you have a chance.

jbonofre:
It looks good to me now that @cshannon fixed it.

jbonofre (approved):

cshannon:
Yeah, the PR is fine, thanks for mergin @jbonofre

PR #1321, opened 2024-10-08
Add original message to JobScheduler remove signature

kana-dowjones:
Currently, JobScheduler doesn't have a reference to the original message being deleted.
This doesn't allow for analysis and possibly retaining parameters of the message being removed

Proposed: add an original message to JobScheduler remove method


mattrpav:
Hi @kana-dowjones-  Thank you for the PR. Please rebase your branch, so only your commit is present in the PR.

Would you please elaborate on the use case for this change? Could this change be implemented with _new_ method overloads vs modifying the existing methods (SPI-breaking change).

Thanks!

PR #1324, opened 2024-10-18
[AMQ-9595] Fix recoverNextMessages when there are messages consumed farther than maxBatchSize.

Nikita-Shupletsov:
Fixes a bug when a durable consumer can received messages which it's already acknowledged

It fixes a bug

I couldn't find any tests that test that part of code directly. I wrote a higher level test that covers that edge case.


mattrpav:
Do you have a unit test the reproduces the issue?

Nikita-Shupletsov:
> Do you have a unit test the reproduces the issue?

the test I added reproduces the issue. without the fix the test will fail

jbonofre (approved):
Good catch and nice work !

Thanks @Nikita-Shupletsov !

PR #1325, opened 2024-10-19
[AMQ-9591] Fix file not exist error when running jetty demo in 6.1.3

evanylem:
What problems does this PR solve?
Fixes a bug where BeanCreationException is thrown for not finding jetty-realm-5.x.properties when running jetty demo in 6.1.3

Why is it beneficial to merge into ActiveMQ?
It fixes a bug

How do you make sure this PR is well tested?
The demo can run without an exception after adding the file.

### DIFF
--- a/activemq-all/pom.xml
+++ b/activemq-all/pom.xml
@@ -14,7 +14,7 @@
   <parent>
     <groupId>org.apache.activemq</groupId>
     <artifactId>activemq-parent</artifactId>
-    <version>6.1.3-SNAPSHOT</version>
+    <version>6.1.3</version>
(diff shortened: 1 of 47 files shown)


mattrpav:
This PR is not properly re-based and the approach appears to be incorrect.

What is the error stacktrace of the error? 

note: ActiveMQ 6.x uses a combined JAAS realm, which is different from 5.x where Jetty had its own JAAS realm and user+password+group file. If there is an issue with the demo, the demo jetty.xml file should be updated, not a reference to the jetty-5.x-realm.properties.

PR #1329, opened 2024-10-23
[AMQ-8463] Add advancedMessageStatistics feature

mattrpav:
Reviewer notes:
- [x] New AdvancedMessageStatisticsEnabled flag per-policyEntry (default: false)
- [x] No new timestamp sampling added (ie no new System.currentTimeMillis() calls)
- [x] Values reset on resetStatistics
- [x] Values exposed via JMX  
- [x] Enqueue Message BrokerInTime, ClientID, MessageID and MessageTimestamp fields
- [x] Dequeue Message BrokerInTime, BrokerOutTime, ClientID, MessageID and MessageTimestamp fields
- [x] Unit test for policyEntry runtime configuration modification
- [x] Unit test for functional testing including network include/exclude verification
- [x] Non-durable topics do not have brokerInTime or messageTimestamp in scope during dequeue, so those values will always be zero (0)
- [x] Convert to using generics for UnsampledStatistic
- [x] Add unit test for UnsampledStatisticImpl (done: 98.3% coverage)

advanced-message-statistics

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/DestinationView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/DestinationView.java
@@ -30,6 +30,8 @@ import jakarta.jms.Connection;
 import jakarta.jms.InvalidSelectorException;
 import jakarta.jms.MessageProducer;
 import jakarta.jms.Session;
+import java.util.function.Function;
+import java.util.function.Supplier;
 import javax.management.MalformedObjectNameException;
 import javax.management.ObjectName;
 import javax.management.openmbean.CompositeData;
@@ -52,6 +54,8 @@ import org.apache.activemq.command.ActiveMQTextMessage;
 import org.apache.activemq.command.Message;
 import org.apache.activemq.filter.BooleanExpression;
 import org.apache.activemq.filter.NonCachedMessageEvaluationContext;
+import org.apache.activemq.management.MessageFlowStats;
+import org.apache.activemq.management.UnsampledStatistic;
 import org.apache.activemq.selector.SelectorParser;
 import org.apache.activemq.store.MessageStore;
 import org.apache.activemq.util.URISupport;
@@ -620,4 +624,63 @@ public class DestinationView implements DestinationViewMBean {
         return destination.getDestinationStatistics().getNetworkDequeues().getCount();
     }
 
+    @Override
+    public boolean isAdvancedMessageStatisticsEnabled() {
+        return destination.isAdvancedMessageStatisticsEnabled();
+    }
+
+    @Override
+    public void setAdvancedMessageStatisticsEnabled(boolean advancedMessageStatisticsEnabled) {
+        destination.setAdvancedMessageStatisticsEnabled(advancedMessageStatisticsEnabled);
+    }
+
+    @Override
+    public long getEnqueuedMessageBrokerInTime() {
+        return getMessageFlowStat(MessageFlowStats::getEnqueuedMessageBrokerInTime, 0L);
+    }
+
+    @Override
+    public String getEnqueuedMessageClientId() {
+        return getMessageFlowStat(MessageFlowStats::getEnqueuedMessageClientID, null);
+    }
+
+    @Override
+    public String getEnqueuedMessageId() {
+        return getMessageFlowStat(MessageFlowStats::getEnqueuedMessageID, null);
+    }
+
+    @Override
+    public long getEnqueuedMessageTimestamp() {
+        return getMessageFlowStat(MessageFlowStats::getEnqueuedMessageTimestamp, 0L);
+    }
+
+    @Override
+    public long getDequeuedMessageBrokerInTime() {
+        return getMessageFlowStat(MessageFlowStats::getDequeuedMessageBrokerInTime, 0L);
+    }
+
+    @Override
+    public long getDequeuedMessageBrokerOutTime() {
+        return getMessageFlowStat(MessageFlowStats::getDequeuedMessageBrokerOutTime, 0L);
+    }
+
+    @Override
+    public String getDequeuedMessageClientId() {
+        return getMessageFlowStat(MessageFlowStats::getDequeuedMessageClientID, null);
+    }
+
+    @Override
+    public String getDequeuedMessageId() {
+        return getMessageFlowStat(MessageFlowStats::getDequeuedMessageID, null);
+    }
+
+    @Override
+    public long getDequeuedMessageTimestamp() {
+        return getMessageFlowStat(MessageFlowStats::getDequeuedMessageTimestamp, 0L);
+    }
+
+    private <T> T getMessageFlowStat(Function<MessageFlowStats, UnsampledStatistic<T>> f, T defVal) {
+        final var stats = destination.getDestinationStatistics().getMessageFlowStats();
+        return stats != null ? f.apply(stats).getValue() : defVal;
+    }
 }
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
@@ -14,41 +14,43 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package org.apache.activemq.broker.region;
 
+import java.util.Optional;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.activemq.management.CountStatisticImpl;
 import org.apache.activemq.management.PollCountStatisticImpl;
 import org.apache.activemq.management.StatsImpl;
 import org.apache.activemq.management.*;
 
 /**
- * The J2EE Statistics for the a Destination.
- *
- *
+ * The Statistics for a Destination.
  */
 public class DestinationStatistics extends StatsImpl {
 
-    protected CountStatisticImpl enqueues;
-    protected CountStatisticImpl dequeues;
-    protected CountStatisticImpl forwards;
-    protected CountStatisticImpl consumers;
-    protected CountStatisticImpl producers;
-    protected CountStatisticImpl messages;
-    protected PollCountStatisticImpl messagesCached;
-    protected CountStatisticImpl dispatched;
-    protected CountStatisticImpl duplicateFromStore;
-    protected CountStatisticImpl inflight;
-    protected CountStatisticImpl expired;
-    protected TimeStatisticImpl processTime;
-    protected CountStatisticImpl blockedSends;
-    protected TimeStatisticImpl blockedTime;
-    protected SizeStatisticImpl messageSize;
-    protected CountStatisticImpl maxUncommittedExceededCount;
+    private final CountStatisticImpl enqueues;
+    private final CountStatisticImpl dequeues;
+    private final CountStatisticImpl forwards;
+    private final CountStatisticImpl consumers;
+    private final CountStatisticImpl producers;
+    private final CountStatisticImpl messages;
+    private final PollCountStatisticImpl messagesCached;
+    private final CountStatisticImpl dispatched;
+    private final CountStatisticImpl duplicateFromStore;
+    private final CountStatisticImpl inflight;
+    private final CountStatisticImpl expired;
+    private final TimeStatisticImpl processTime;
+    private final CountStatisticImpl blockedSends;
+    private final TimeStatisticImpl blockedTime;
+    private final SizeStatisticImpl messageSize;
+    private final CountStatisticImpl maxUncommittedExceededCount;
 
-    // [AMQ-9437] Advanced Statistics are optionally enabled
-    protected CountStatisticImpl networkEnqueues;
-    protected CountStatisticImpl networkDequeues;
+    // [AMQ-9437] Advanced Network Statistics
+    private final CountStatisticImpl networkEnqueues;
+    private final CountStatisticImpl networkDequeues;
+
+    // [AMQ-8463] Advanced Message Statistics are disabled by default
+    private final AtomicReference<MessageFlowStatsImpl> messageFlowStats = new AtomicReference<>();
 
     public DestinationStatistics() {
 
@@ -75,25 +77,6 @@ public class DestinationStatistics extends StatsImpl {
 
         networkEnqueues = new CountStatisticImpl("networkEnqueues", "The number of messages that have been sent to the destination via network connection");
         networkDequeues = new CountStatisticImpl("networkDequeues", "The number of messages that have been acknowledged from the destination via network connection");
-
-        addStatistic("enqueues", enqueues);
-        addStatistic("dispatched", dispatched);
-        addStatistic("dequeues", dequeues);
-        addStatistic("duplicateFromStore", duplicateFromStore);
-        addStatistic("inflight", inflight);
-        addStatistic("expired", expired);
-        addStatistic("consumers", consumers);
-        addStatistic("producers", producers);
-        addStatistic("messages", messages);
-        addStatistic("messagesCached", messagesCached);
-        addStatistic("processTime", processTime);
-        addStatistic("blockedSends",blockedSends);
-        addStatistic("blockedTime",blockedTime);
-        addStatistic("messageSize",messageSize);
-        addStatistic("maxUncommittedExceededCount", maxUncommittedExceededCount);
-
-        addStatistic("networkEnqueues", networkEnqueues);
-        addStatistic("networkDequeues", networkDequeues);
     }
 
     public CountStatisticImpl getEnqueues() {
@@ -132,10 +115,6 @@ public class DestinationStatistics extends StatsImpl {
         return messages;
     }
 
-    public void setMessagesCached(PollCountStatisticImpl messagesCached) {
-        this.messagesCached = messagesCached;
-    }
-
     public CountStatisticImpl getDispatched() {
         return dispatched;
     }
@@ -170,6 +149,10 @@ public class DestinationStatistics extends StatsImpl {
         return networkDequeues;
     }
 
+    public MessageFlowStats getMessageFlowStats() {
+        return messageFlowStats.get();
+    }
+
     public void reset() {
         if (this.isDoReset()) {
             super.reset();
@@ -186,6 +169,8 @@ public class DestinationStatistics extends StatsImpl {
             maxUncommittedExceededCount.reset();
             networkEnqueues.reset();
             networkDequeues.reset();
+            Optional.ofNullable(messageFlowStats.get())
+                .ifPresent(MessageFlowStatsImpl::reset);
         }
     }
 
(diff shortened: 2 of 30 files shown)


cshannon (commented):
One thing I noticed is the new statistic impl classes all are copy and pasted code, it seems like you could just re-use all of that if you used generics.

cshannon:
I know this would be a much bigger effort but if we intend to keep adding more and more metrics, from a performance and maintainability standpoint, we should really look into moving everything to Micrometer which adds more more flexibility and should be better performance. That kind of change would likely end up being a major version, ie AMQ 7.0 if we wanted to do that.

activemq-client/src/main/java/org/apache/activemq/management/UnsampledStatisticImpl.java
```diff
@@ -0,0 +1,45 @@
...
+ * limitations under the License.
+ */
+package org.apache.activemq.management;
+
+/**
+ * Base class for a UnsampledStatistic implementation
+ * 
+ * 
+ */
+public abstract class UnsampledStatisticImpl extends StatisticImpl implements UnsampledStatistic {
```
cshannon, line 24: Do we actually need this abstract class? Does it make sense to just have default methods in UnsampledStatistic?

cshannon: Ah nevermind, I guess it extends StatisticImpl so we need it

activemq-client/src/main/java/org/apache/activemq/management/StringStatisticImpl.java
```diff
@@ -0,0 +1,58 @@
...
+ */
+package org.apache.activemq.management;
+
+/**
+ * A string statistic implementation
+ */
+public class StringStatisticImpl extends UnsampledStatisticImpl implements StringStatistic {
+
+    // Note: [AMQ-8463] Adding volatile to 'value' would be more accurate, but performance impact
+    private String value = null;
```
cshannon, line 25: We should not be compromising correctness for performance. If this is multi-threaded you need to make it thread safe. Furthermore, the impact of volatile is going to be negligible in the real world.

You could use AtomicReference and make this final

activemq-client/src/main/java/org/apache/activemq/management/LongStatistic.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+package org.apache.activemq.management;
+
+/**
+ * A statistic to store a single long value that is not incremented
+ * 
+ * Example: Store a timestamp value of a recent message
+ * 
+ */
+public interface LongStatistic extends UnsampledStatistic {
```
cshannon, line 27: You don't need these interfaces if you just use generics as I mentioned in my other comment.

activemq-client/src/main/java/org/apache/activemq/management/StringStatisticImpl.java
mattrpav: I'm thinking 'volatile' over AtomicReference. Same thread-safety, fewer objects created.

activemq-client/src/main/java/org/apache/activemq/management/LongStatistic.java
mattrpav: Yeah, I think this could be changed, but I'm wondering if it gets verbose to the consumer of the metric. I felt my original approach was "following the as-is" vs getting to the optimal target state.

If we go generic, do we add a generic parameter for all the dimensions? Sampled, Summed, Ranged, Fixed value, etc?

I think the interfaces do a good job of communicating the type. Feels like adding generics just pushes the type info to the consuming service code.

```
   public UnsampledStatisticImpl<Long> getDequeuedMessageBrokerOutTime() {
        return dequeuedMessageBrokerOutTime;
    }
```

I'm not married to either approach.

cshannon: I think it mostly depends on if the different types will have any logic that is different or needs to be overriden. You could also move common code into an abstract class if needed, I was just noticing a lot of copy/paste and seemed like generics could be useful.

activemq-client/src/main/java/org/apache/activemq/management/StringStatisticImpl.java
cshannon: I don't think the extra object creation matters in this case and the extra features of the atomic classes are nice. One thing i just realized was that if we want to use other Atomic classes like AtomicLong, which would be nice, then we may not be able to use generics after all as that atomic type needs to be different and we'd have slightly different code

tabish121: Depending on the use case this is what static atomic field updater types are for, all the features like CAS but only one static allocation.

activemq-client/src/main/java/org/apache/activemq/management/UnsampledStatisticImpl.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.util.concurrent.atomic.AtomicReference;
+
+/**
+ * UnsampledStatistic<T> implementation
+ * 
+ */
+public class UnsampledStatisticImpl<T> extends StatisticImpl implements UnsampledStatistic<T> {
+
+    protected final AtomicReference<T> value = new AtomicReference<>();
```
cshannon, line 27: Is the intent to have another class subclass this at some point? I'm wondering why these are protected and not private.

mattrpav: Good catch. I’ll tidy that up

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
```diff
@@ -76,6 +87,17 @@ public DestinationStatistics() {
...
         networkDequeues = new CountStatisticImpl("networkDequeues", "The number of messages that have been acknowledged from the destination via network connection");
 
+        enqueuedMessageBrokerInTime = new UnsampledStatisticImpl<>("enqueuedMessageBrokerInTime", "ms", "Broker in time (ms) of last enqueued message to the destination", Long.valueOf(0l));
+        enqueuedMessageClientID = new UnsampledStatisticImpl<>("enqueuedMessageClientID", "id", "ClientID of last enqueued message to the destination", null);
+        enqueuedMessageID = new UnsampledStatisticImpl<>("enqueuedMessageID", "id", "MessageID of last enqueued message to the destination", null);
+        enqueuedMessageTimestamp = new UnsampledStatisticImpl<>("enqueuedMessageTimestamp", "ms", "Message timestamp of last enqueued message to the destination", Long.valueOf(0l));
+
+        dequeuedMessageBrokerInTime = new UnsampledStatisticImpl<>("dequeuedMessageBrokerInTime", "ms", "Broker in time (ms) of last dequeued message to the destination", Long.valueOf(0l));
```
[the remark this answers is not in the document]
mattrpav: That’s a good idea. It’ll add a bit more logic to support dynamically enabling/disabling advanced statistics. 

I can add that for both advanced message and network statistics.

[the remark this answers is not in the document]
mattrpav: I experimented with this. I ran into a condition where supporting dynamic config update (good feature) would require a lock around all advanced network and advanced message stats operations.

Possible paths to move forward:

Option 1: Do not optimize for memory, allocate all metrics at startup time

Option 2: Lazy init advanced stats fields. Support only allocating statistics when moving from enabled = false to enabled = true. Moving from enabled = true to enabled = false would leave metrics fields present (or perhaps we could reset them)

Option 3: Add locking around all the advanced[Network|Message]Statistics operations (bleh)

mattrpav:
Local performance testing showed no degradation in performance

64x producers sending in async to a single topic sent 1M messages ea in ~24s on a 2023 14" Apple M3 Pro

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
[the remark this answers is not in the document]
mattrpav: Using a wrapper object for MessageFlowStatistics to support enable/disable and not require memory allocation for unused metrics

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/DestinationView.java
```diff
@@ -620,4 +621,67 @@ public long getNetworkDequeues() {
...
+    }
+
+    @Override
+    public String getDequeuedMessageId() {
+        MessageFlowStats tmpMessageFlowStats = destination.getDestinationStatistics().getMessageFlowStats();
+        return (tmpMessageFlowStats != null ? tmpMessageFlowStats.getDequeuedMessageID().getValue() : null);
+    }
+
+    @Override
+    public long getDequeuedMessageTimestamp() {
```
cshannon, line 683: I was thinking all of these getters could be a bit cleaner and less copy/paste with using lambdas. IE.

```java
@Override
public long getDequeuedMessageTimestamp() {
    return getMessageFlowStat(MessageFlowStats::getDequeuedMessageTimestamp, 0L);
}

private <T> T getMessageFlowStat(Function<MessageFlowStats, UnsampledStatistic<T>> f, T defVal) {
    final var stats = destination.getDestinationStatistics().getMessageFlowStats();
    return stats != null ? f.apply(stats).getValue() : defVal;
}
```

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
```diff
@@ -186,6 +177,11 @@ public void reset() {
             maxUncommittedExceededCount.reset();
             networkEnqueues.reset();
             networkDequeues.reset();
+
+            MessageFlowStatsImpl tmpMessageFlowStats = messageFlowStats;
```
cshannon, line 181: Nit: I usually like to make all of these final just to prevent any future mistakes

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
```diff
@@ -252,7 +255,22 @@ public void setParent(DestinationStatistics parent) {
             maxUncommittedExceededCount.setParent(null);
             networkEnqueues.setParent(null);
             networkDequeues.setParent(null);
+            // [AMQ-9437] Advanced Message Statistics do not parent.
         }
     }
 
+    public synchronized void setAdvancedMessageStatisticsEnabled(boolean advancedMessageStatisticsEnabled) {
```
cshannon, line 262: There is a lot of thread safety issues in this entire class. By making all of this stuff protected someone could easily screw up the state outside a lock. Furthermore you are synchronizing this method but not other places like reset() or anywhere else.

You should not be allowing the ability to set messageFlowStats or to set advancedMessageStatisticsEnabled() outside this method as you could get them out of sync and right now it's easy to do because they are not private

I need to think about it more but I really don't think the current approach is great, especially if you are making member variables protected and not private.

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -1971,10 +1978,17 @@ private final boolean tryCursorAdd(final Message msg) throws Exception {
...
     final void messageSent(final ConnectionContext context, final Message msg) throws Exception {
         pendingSends.decrementAndGet();
+
         destinationStatistics.getEnqueues().increment();
         destinationStatistics.getMessages().increment();
         destinationStatistics.getMessageSize().addSize(msg.getSize());
 
+        MessageFlowStats tmpMessageFlowStats = destinationStatistics.getMessageFlowStats();
+
+        if(isAdvancedMessageStatisticsEnabled() && tmpMessageFlowStats != null) {
```
cshannon, line 1988: This is something where both of these should ALWAYS be in sync. you should never have the stats enabled but the object is null.

Maybe the solution is you drop the flag entirely. Maybe stats being enabled is really just a "is this object not null" and disabling is making the stats null so we don't have to check 2 objects.

activemq-client/src/main/java/org/apache/activemq/management/StatisticImpl.java
```diff
@@ -83,32 +89,31 @@ public synchronized long getLastSampleTime() {
     /**
      * @return the enabled
      */
-    public boolean isEnabled() {
+    public synchronized boolean isEnabled() {
```
cshannon, line 92: There is no benefit to making these methods synchronized, the variables are already volatile and reading or setting the boolean is atomic

activemq-client/src/main/java/org/apache/activemq/management/StatsImpl.java
```diff
@@ -38,44 +35,35 @@ public StatsImpl(Set<StatisticImpl> set) {
...
+        return this.set.toArray(new Statistic[this.set.size()]);
     }
 
-    protected void addStatistic(String name, StatisticImpl statistic) {
+    @Deprecated(forRemoval = true, since = "6.2.0")
+    protected synchronized void addStatistic(String name, StatisticImpl statistic) {
         this.set.add(statistic);
     }
+
+    protected synchronized void addStatistics(Collection<StatisticImpl> statistics) {
```
cshannon, line 62: This also doesn't make a lot of sense, you don't need to synchronize and also use a CopyOnWriteArrayList.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/DestinationView.java
mattrpav: Completed w/ cshannon's change

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
mattrpav: Resolved w/ cshannon's commit

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
mattrpav: Resolved w/ cshannon's commit

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
mattrpav: Resolved w/ cshannon's commit

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
```diff
@@ -233,6 +222,7 @@ public void setParent(DestinationStatistics parent) {
             maxUncommittedExceededCount.setParent(parent.maxUncommittedExceededCount);
             networkEnqueues.setParent(parent.networkEnqueues);
             networkDequeues.setParent(parent.networkDequeues);
+            // [AMQ-9437] Advanced Message Statistics do not parent.
```
cshannon, line 225:
```suggestion
            // [AMQ-9437] Advanced Message Statistics does not have a parent.
```

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
```diff
@@ -252,7 +242,29 @@ public void setParent(DestinationStatistics parent) {
             maxUncommittedExceededCount.setParent(null);
             networkEnqueues.setParent(null);
             networkDequeues.setParent(null);
+            // [AMQ-9437] Advanced Message Statistics do not parent.
```
cshannon, line 245:
```suggestion
            // [AMQ-9437] Advanced Message Statistics does not have a parent.
```

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
```diff
@@ -252,7 +242,29 @@ public void setParent(DestinationStatistics parent) {
...
         }
     }
 
+    // This is the only method that can mutate the messageFlowStats state
+    // so we only need to synchronize this method to make sure we don't have 2 threads
+    // trying to set at the same time.
+    public synchronized void setAdvancedMessageStatisticsEnabled(boolean enabled) {
+        // we can just use the get() here on the reference as no other spot can
```
cshannon, line 253: You can get rid of this comment here after this was refactored

cshannon (approved):
made a few comments to look at but for the most part it LGTM

activemq-client/src/main/java/org/apache/activemq/management/StatisticImpl.java
cshannon: This is still not changed as well

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
mattrpav: Changes to comments applied via commit c079d1bd03df5a2faf35f2ec9625e7e51b2a4973

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
mattrpav: Changes to comment applied via commit c079d1bd03df5a2faf35f2ec9625e7e51b2a4973

activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationStatistics.java
mattrpav: Comment removed via commit c079d1bd03df5a2faf35f2ec9625e7e51b2a4973

activemq-client/src/main/java/org/apache/activemq/management/StatisticImpl.java
```diff
@@ -15,22 +15,19 @@
...
-    protected boolean enabled;
+    protected volatile boolean enabled;
 
-    private String name;
-    private String unit;
-    private String description;
-    private long startTime;
-    private long lastSampleTime;
-    private boolean doReset = true;
+    protected final String name;
```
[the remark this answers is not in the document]
mattrpav: Changes applied via commit c079d1bd03df5a2faf35f2ec9625e7e51b2a4973

activemq-client/src/main/java/org/apache/activemq/management/StatsImpl.java
mattrpav: Changes applied via commit c079d1bd03df5a2faf35f2ec9625e7e51b2a4973

activemq-client/src/main/java/org/apache/activemq/management/StatisticImpl.java
mattrpav: Changes applied via commit c079d1bd03df5a2faf35f2ec9625e7e51b2a4973

PR #1354, opened 2024-11-15
[WIP]  Add preserveDeliveryMode to deadLetterStrategy

pa-deasy:
By default non persistent messages are set to persistent when being sent to a dead letter queue. This improvement adds `preserveDeliveryMode` to deadLetterStrategy which allows the original non-persistent delivery mode to be preserved.

This is a WIP progress, I'm looking for early feedback from the community to ensure my changes are reasonable.
I've opted to move dead letter queue message preparation logic to the `AbstractDeadLetterStrategy` class. I believe this makes sense as the strategy influences how the message is prepared. It also reduces responsibility in the `RegionBroker` which is already quite large.

I will add tests if I get feedback that the approach is reasonable.


activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/DeadLetterStrategy.java
```diff
@@ -60,6 +71,16 @@ public interface DeadLetterStrategy {
      */
     public void setProcessNonPersistent(boolean processNonPersistent);
 
+    /**
+     * @return the PreserveDeliveryMode
```
kenliao94, line 75: The comment needs to be updated.

pa-deasy: > @return the PreserveDeliveryMode

This is a new method which does return `PreserveDeliveryMode`. What exactly would you like to update?

pa-deasy: Yeah, I noticed the "is" style is used for boolean methods in a number of places so I chose to be consistent.

kenliao94:
@pa-deasy I assume the code that actually calls "setPreserveDeliveryMode" which is read from destination configuration will come later?

kenliao94:
I sort of see the points of moving the code to `AbstractDeadLetterStrategy#prepareMessageForDeadLetterQueue`for readability in the chunky `RegionBroker`. My concern of that approach is that it effectively shifts the responsibility of preparing the message (setting correct attributes of the message object) to the this abstract class, which imo shouldn't need to be concern of it. This preparation is necessary before you send the message, but by moving to the strategy class, it feels like it's more a "util" then a necessary step.  IMO `DeadLetterStrategy` should have logic that only perform mutable action on the policy itself and shouldn't perform mutable actions on the `Message` object. 

Hence I vote -1  on "move dead letter queue message preparation logic to the AbstractDeadLetterStrategy class" part of this PR. The getter and setter of perserveDeliveryMode on the `Strategy` class makes sense to me.

mattrpav:
I'm not in favor of refactoring RegionBroker in a point release. Ultimately, the DeadLetterStrategy should re-use the MessageInterceptorStrategy interface to it so admins can configure or change anything about the message on the way to the DLQ. All these one-off config params and methods creates bloat and all this logic should be pushed off into a handler.

pa-deasy:
> @pa-deasy I assume the code that actually calls "setPreserveDeliveryMode" which is read from destination configuration will come later?
Yes that should be straightforward. I just wanted feedback on the `RegionBroker` and DeadLetterStrategy` for the draft.

Thanks folks for your feedback on "move dead letter queue message preparation logic to the AbstractDeadLetterStrategy class". Those are great points. I will update my PR appropriately and add tests. 

Thank you.

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/DeadLetterStrategy.java
```diff
@@ -60,6 +60,16 @@ public interface DeadLetterStrategy {
...
+     */
+    public boolean isPreserveDeliveryMode();
+
+    /**
+     * @param PreserveDeliveryMode the PreserveDeliveryMode to set
+     */
+    public void setPreserveDeliveryMode(boolean PreserveDeliveryMode);
```
Nikita-Shupletsov, line 71: preserveDeliveryMode

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/DeadLetterStrategy.java
Nikita-Shupletsov: yeah, is for boolean is the convetion

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/DeadLetterStrategy.java
pa-deasy: Good spot, not sure how I made this mistake.

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/AbstractDeadLetterStrategy.java
```diff
@@ -91,6 +129,14 @@ public void setProcessNonPersistent(boolean processNonPersistent) {
         this.processNonPersistent = processNonPersistent;
     }
 
+    @Override
+    public boolean isPreserveDeliveryMode() { return this.preserveDeliveryMode; }
```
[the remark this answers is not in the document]
pa-deasy: Strange, my IDE(ItelliJ) made the other getters in `AbstractDeadLetterStrategy` appear to be single line.
I'll update this.

activemq-broker/src/main/java/org/apache/activemq/broker/region/RegionBroker.java
```diff
@@ -781,23 +781,8 @@ public boolean sendToDeadLetterQueue(ConnectionContext context, MessageReference
...
-                            message.setExpiration(dlqExpiration);
-                            if (!message.isPersistent()) {
-                                message.setPersistent(true);
-                                message.setProperty("originalDeliveryMode", "NON_PERSISTENT");
-                            }
-                            if (poisonCause != null) {
-                                message.setProperty(ActiveMQMessage.DLQ_DELIVERY_FAILURE_CAUSE_PROPERTY,
-                                        poisonCause.toString());
-                            }
+                            message = prepareMessageForDeadLetterQueue(message, deadLetterStrategy, poisonCause);
```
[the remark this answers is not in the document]
cshannon: Well, while the method is in fact descriptive about what it is doing, it actually makes it harder to understand what you changed because by copying it all to a new method it's not clear what exactly the difference is from the old code without looking at it carefully as the entire method shows up as a new change which is the biggest reason why I made the comment because i as staring at it for a while to even figure out what was different.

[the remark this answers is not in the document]
pa-deasy: Makes sense, I've removed the `prepareMessageForDeadLetterQueue` method.

cshannon:
@mattrpav - Do you have any issues with the changes here? It seems ok to me, it preserves the previous behavior by default and I could see a use case for this.

jbonofre:
I will create a Jira associated with the change. The PR description is well detailed though.

PR #1355, opened 2024-11-20
AMQ-9625 - Prevent queue messages from becoming stuck

cshannon:
Fixes a race condition bug that can lead to a message being missed on dispatch and stuck on a Queue until restart when caching and concurrentStoreAndDispatch are enabled on a Queue and the cache becomes disabled.

This is pretty much impossible to reproduce or test in a unit test so there's no test for this but the risk of breaking anything is very low because all this is doing is just waiting for the future to complete to ensure the task finishes before we continue.


cshannon:
I updated the PR slightly as we needed the check/wait for the async to be done not only when the store is being synced, but also for any outstanding in flight async messages that were in progress by other threads when the mode switched.

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/AbstractStoreCursor.java
```diff
@@ -255,12 +255,20 @@ public synchronized boolean tryAddMessageLast(MessageReference node, long wait)
             disableCache = true;
         }
 
-        if (disableCache && isCacheEnabled()) {
+        // AMQ-9625 - use this.cacheEnabled directly because the method isCacheEnabled() is overriden
+        // to try to re-enable the cache which we don't want at this point as we already skipped
+        // adding it to the cache
+        if (disableCache && this.cacheEnabled) {
             if (LOG.isTraceEnabled()) {
```
mattrpav, line 262: Thoughts on follow-on JIRA to add a metric or state flag when this occurs? Trace log is good for debugging, but when managing lots of brokers, its seems this could be a good thing to bubble up so admins know it occurs or if it is occurring frequently, some tuning may need to be required.

cshannon: I don't see value in a metric because we already know if this is occurring frequently based on memory usage and adding a metric wouldn't really help with tuning because it doesn't have anything to do with the problem. 

If this is happening, the problem that needs tuning is consumers are not fast enough and memory is full which is already easily determined by existing logging and tracking for when the cache is disabled and memory thresholds are hit. The admins should be paying attention to the memory being full as that is the root cause. If the memory doesn't fill up then the cache won't be disabled.

This fix is just for the race condition on the pending writes and adding another metric here doesn't really help an admin with any real useful information they would not already have. We are essentially just fixing a bug that happens because memory is full but it doesn't really effect anything with tuning.

PR #1358, opened 2024-11-29
AMQ-9627 Don't fail on start if cachedLDAPAuthorizationMap is used and the LDAP server is not available.

Nikita-Shupletsov:
it fixes a crash on start if cachedLDAPAuthorizationMap is used and the LDAP server is not available.

it's rather inconvenient, as cachedLDAPAuthorizationMap is usually used to avoid issues with the LDAP server being not always available

I added a test that reproduces the error without the change.


activemq-broker/src/main/java/org/apache/activemq/security/SimpleCachedLDAPAuthorizationMap.java
```diff
@@ -936,7 +936,11 @@ public void namingExceptionThrown(NamingExceptionEvent namingExceptionEvent) {
 
     // Init / Destroy
     public void afterPropertiesSet() throws Exception {
-        query();
+        try {
+            query();
+        } catch (Exception e) {
+            LOG.error("Error updating authorization map.  Partial policy may be applied until the next successful update.", e);
```
Nikita-Shupletsov, line 942: I copied the existing error message from https://github.com/apache/activemq/blob/main/activemq-broker/src/main/java/org/apache/activemq/security/SimpleCachedLDAPAuthorizationMap.java#L679

please let me know what you think

kenliao94: I would also include a string that suggest ways for the user to mitigate it. Something like "The LDAP server might not be reachable, check ..." even tho the root cause can be many reason. Because "authorization map" is an internal concept, user might not get it and they don't know how to get themselves unstuck. However, It needs to be phrase in such a way that this is one possible root cause, but not necessarily THE root cause.

kenliao94:
Question @Nikita-Shupletsov, once the user fixed the issue with the configuration or their LDAP server, does the user need to restart the server? In another word, is this code part of the code path that get triggered everytime the broker performs AuthN/AuthZ?

If that latter case is true, the activemq.log be flooded with error message. How do we make it less noisy?

Nikita-Shupletsov:
> Question @Nikita-Shupletsov, once the user fixed the issue with the configuration or their LDAP server, does the user need to restart the server? In another word, is this code part of the code path that get triggered everytime the broker performs AuthN/AuthZ?
> 
> If that latter case is true, the activemq.log be flooded with error message. How do we make it less noisy?

No, they will not need to. afaik, that's the whole idea of this Cached Map is to be able to withstand connection failures. it will try to fetch data on every request until is succeeds

Nikita-Shupletsov:

PR #1369, opened 2024-12-13
[AMQ-9637] Add web socket connection limit.

Nikita-Shupletsov:
It adds `maximumConnections` support for WS transport.

all other transports have this limit. being able to limit the number of connections for WS transport brings in in par with the rest of transports.

I updated the existing amqp test to test amqp+ws and amqp+wss, added new tests for stomp+ws and mqtt+ws as well.


Nikita-Shupletsov:
> On the surface, this looks like it might be a couple of commits-- ws socket handling clean-up and then the limit added.

sounds good. I will split the PR into two

PR #1370, opened 2024-12-13
AMQ-9162 Fix WS transport start via JMX after it was closed.

Nikita-Shupletsov:
As it was discussed in [the ticket](https://issues.apache.org/jira/browse/AMQ-9162), the issue exists only in 5.x versions, hence the PR is only to the 5.18.x branch.

Also added a test that reproduces the issue without the fix.

### DIFF
--- a/activemq-http/src/main/java/org/apache/activemq/transport/ws/WSTransportServer.java
+++ b/activemq-http/src/main/java/org/apache/activemq/transport/ws/WSTransportServer.java
@@ -96,7 +96,13 @@ public class WSTransportServer extends WebTransportServerSupport implements Brok
 
         contextHandler.setAttribute("acceptListener", getAcceptListener());
 
-        server.start();
+        ClassLoader original = Thread.currentThread().getContextClassLoader();
+        Thread.currentThread().setContextClassLoader(WSTransportServer.class.getClassLoader());
+        try {
+            server.start();
+        } finally {
+            Thread.currentThread().setContextClassLoader(original);
+        }
 
         // Update the Connect To URI with our actual location in case the configured port
         // was set to zero so that we report the actual port we are listening on.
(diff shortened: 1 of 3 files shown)


cshannon (changes requested):
The err log file should obviously not be part of this PR.

activemq-http/src/main/java/org/apache/activemq/transport/ws/WSTransportServer.java
```diff
@@ -96,7 +96,13 @@ protected void doStart() throws Exception {
 
         contextHandler.setAttribute("acceptListener", getAcceptListener());
 
-        server.start();
+        ClassLoader original = Thread.currentThread().getContextClassLoader();
+        Thread.currentThread().setContextClassLoader(WSTransportServer.class.getClassLoader());
```
jbonofre, line 100: This could be problematic in OSGi/Karaf generally speaking. In this specific it should be ok as the `WSTransportServer` is in the `activemq-osgi` bundle.

jbonofre (changes requested):
`hs_err_pid*.log` should not be part of this PR. Please remove it.

Nikita-Shupletsov:
> The err log file should obviously not be part of this PR.


> `hs_err_pid*.log` should not be part of this PR. Please remove it.


sorry for that, removed

jbonofre:
@cshannon I did a new pass (including OSGi) and this looks good to me. Do you mind to take a new look and maybe remove your -1 ?

mattrpav:
@jbonofre should this be re-targeted to activemq-5.19.x?

jbonofre:
@mattrpav yes, that makes sense to include the fix in 5.19.x as well.

Nikita-Shupletsov:

PR #1376, opened 2025-01-13
[AMQ-9645] Fix double slash in path

hailengc:
### Problem 
When ActiveMQ starts up with `./bin/activemq console`, there is double slash (`//`) shows in the output, example:
 
*  ... Configurations are loaded in the following order: (omit...)/apache-activemq-6.2.0-SNAPSHOT//bin/setenv
*  .... INFO: Creating pidfile  (omit...)/apache-activemq-6.2.0-SNAPSHOT//data/activemq.pid
*  ... -Dactivemq.classpath= (omit...)/apache-activemq-6.2.0-SNAPSHOT//conf

### Description of Change
This fix uses  `Parameter Expansion` which is  a POSIX standard feature in shell ( reference [doc](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02)) . The syntax  `"${ACTIVEMQ_BASE%/}/data"` meaning it removes the trailing `/` if  `$ACTIVEMQ_BASE` ends with a `/`, otherwise it does nothing.

* Apply this to both  `$ACTIVEMQ_HOME` and `$ACTIVEMQ_BASE` where they are concatenated with sub path
* Not touching anything regarding non-linx platform ("cygwin" )

### Test
* Running it locally and seeing no double slash in the path in the output.  

* Verified Activemq starts up normally with pid file created. 

* I also made changes in below config files and they are correctly reflected

      * users.properties and login.properties 
      * acitvemq.xml and jolokia-access.xml
      * log4j2.properties

### DIFF
--- a/assembly/src/release/bin/activemq
+++ b/assembly/src/release/bin/activemq
@@ -108,14 +108,14 @@ if [ -z "$ACTIVEMQ_USER_CLASSPATH" ] ; then
 fi
 
 # ActiveMQ Classpath configuration
-ACTIVEMQ_CLASSPATH="$ACTIVEMQ_BASE/../lib/:$ACTIVEMQ_USER_CLASSPATH"
+ACTIVEMQ_CLASSPATH="${ACTIVEMQ_BASE%/}/../lib/:$ACTIVEMQ_USER_CLASSPATH"
 
 # Active MQ configuration directory
 if [ -z "$ACTIVEMQ_CONF" ] ; then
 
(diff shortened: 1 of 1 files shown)


kenliao94 (approved):
Thanks!

cshannon:
This should have a Jira issue created so it can be tracked before it is merged

hailengc:
Hey @cshannon , yes I'll create one.

hailengc:
JIRA issue created:  https://issues.apache.org/jira/browse/AMQ-9645

jbonofre (approved):

PR #1380, opened 2025-01-17
[AMQ-9647]  2025-01 Maven Plugin version updates

mattrpav:
apache parent pom 33
maven-javadoc-plugin-version 3.11.2
maven-install-plugin-version 3.1.3
maven-dependency-plugin-version 3.8.1
maven-bundle-plugin 6.0.0 (JDK 17 required, not back-ported)

### DIFF
--- a/activemq-client/src/test/java/org/apache/activemq/thread/TaskRunnerTest.java
+++ b/activemq-client/src/test/java/org/apache/activemq/thread/TaskRunnerTest.java
@@ -91,7 +91,6 @@ public class TaskRunnerTest {
                         for (int i = 0; i < enqueueCount / workerCount; i++) {
                             queue.incrementAndGet();
                             runner.wakeup();
-                            yield();
                         }
                     } catch (BrokenBarrierException e) {
                     } catch (InterruptedException e) {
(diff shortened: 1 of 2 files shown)


jbonofre:
It does't work so easily: it needs additional change for release, reproducible build, optional import, etc. 
The OSGi distribution doesn't work ootb with that as maven-bundle-plugin 6.0.0 changes the default dynamic import behavior.
As you can see I created the corresponding Jira ([AMQ-9492](https://issues.apache.org/jira/browse/AMQ-9492) etc

So please
1. Check the existing Jira
2. We agreed to target 6.2.0 (according to semver) for this kind of "impactul" changes (for maven-bundle-plugin and Apache POM 33 here).

Do you mind if I close this PR in favor of dedicated ones including other required changes ?

mattrpav:
@jbonofre note the fix in this commit for Apache pom 33. They explicitly use the compiler maven properties, so they need to be overwritten.

```
<maven.compiler.source>11</maven.compiler.source>
<maven.compiler.target>17</maven.compiler.target>
<source-version>${maven.compiler.source}</source-version>
<target-version>${maven.compiler.target}</target-version>
```

PR #1386, opened 2025-02-03
[AMQ-9656] Cleanup kahadb compiler warnings

mattrpav:


activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MultiKahaDBTransactionStore.java
```diff
@@ -427,15 +427,16 @@ private void recoverPendingLocalTransactions() throws IOException {
...
-        DataByteArrayInputStream is = new DataByteArrayInputStream(journal.read(location));
-        byte readByte = is.readByte();
-        KahaEntryType type = KahaEntryType.valueOf(readByte);
-        if (type == null) {
-            throw new IOException("Could not load journal record. Invalid location: " + location);
-        }
-        JournalCommand<?> message = (JournalCommand<?>) type.createMessage();
-        message.mergeFramed(is);
-        return message;
+        try(DataByteArrayInputStream is = new DataByteArrayInputStream(journal.read(location))) {
```
cshannon, line 430: Do we actually want to close this? We don't close it in regular KahadDB and the javadocs for DataByteArrayInputStream can be used more than once so I am wondering if there will be any unintentional side effects or if this should have always been closed after loading.

cshannon: Looking at it more carefully i don't think it matters either way. The source value returned is a ByteSequence which just wraps a byte array and that is passed to DataByteArrayInputStream. So in this case there's nothing to actually close, it will just be a no-op.

cshannon (approved):
changes look fine, as i said the auto closeable change is a no-op as close() isn't implemented so it doesn't really matter either way if you leave it or revert that

PR #1389, opened 2025-02-04
AMQ-9658 - Properly increment transport receive counter

cshannon:
Switch to using an AtomicInteger for tracking bytes received in a TcpTransport. This makes incrementing the counter an atomic operation. Previously a volatile int was used and incrementing volatiles is not atomic because it's a 3 step process of read, update, set.

This also makes a small fix to ensure that the full initialization buffer will always be entirely read and processed when using the auto+nio+ssl transport. Previous the code assumed only the first command was stored in the initialization buffer but technically more bytes could exist for a future command (even if unlikely with the current Java implementation).


jbonofre (commented):
It looks good to me. Sorry for the late review. I'm now back on AMQ 😄

cshannon:
> It looks good to me. Sorry for the late review. I'm now back on AMQ 😄

No problem, i figured the tests would have caught any issues so it was safe to merge as it was just mostly a minor refactoring

PR #1390, opened 2025-02-19
[AMQ-8324] Add design doc for Jakarta Messaging 3.1 asynchronous send

kenliao94:
Add design doc for implementing Jakarta Messaging 3.1 asynchronous send per the latest process discussed on dev list.

Note to reviewers, GitHub rich diff doesn't render image properly. You can click "view file" from the hamburger button to see the rendered markdown.

### DIFF
--- a/docs/proposal/jakarta_messaging_3_1_async_producer/jakarta_messaging_3_1_async_producer.md
+++ b/docs/proposal/jakarta_messaging_3_1_async_producer/jakarta_messaging_3_1_async_producer.md
@@ -1,6 +1,5 @@
 # Jakarta Messaging 3.1: asynchronous send design doc
 
-Date: Jan 5th 2025
 
 The purpose of this document is to discuss the design of implementing asynchronous send (with CompletionListener) that is JMS 2.0 / Jakarta Messaging 3.1 compliant.
 


jbonofre (changes requested):
Image is fine for design diagram/schema.
However, it should not be used for code.
Instead we should use the markdown code block: the huge advantage is that we can suggest change during the review.

docs/proposal/jakarta_messaging_3_1_async_producer/jakarta_messaging_3_1_async_producer.md
```diff
@@ -0,0 +1,225 @@
+# Jakarta Messaging 3.1: asynchronous send design doc
+
+Date: Jan 5th 2025
```
jbonofre, line 3: I would not add date here (the PR date is good enough 😄 ).

cshannon:
> Image is fine for design diagram/schema. However, it should not be used for code. Instead we should use the markdown code block: the huge advantage is that we can suggest change during the review.

+1 for this

kenliao94:
Use codeblock instead of image.

cshannon:
I haven't had a chance to get back to looking at this yet but still plan to, @mattrpav have you had a change to review this yet?

docs/proposal/jakarta_messaging_3_1_async_producer/jakarta_messaging_3_1_async_producer.md
```diff
@@ -0,0 +1,227 @@
...
+
+Cons:
+
+- It sort of defies the purpose of async send because the application thread needs to wait before the send is completed (can lead to performance issues). Even though it doesn’t violate the spec.
+
+This approach is recommended due to its simplicity. Will follow up with a plan to optimize such behaviour.
+
+**Approach 2: Application thread submits the async send to a work queue, a single background thread will dequeue a message from the queue, send it asynchronously and wait for it to finish, before dequeuing the next.**
+
+This is similar to approach 1, the async message is sent one at a time. The difference being instead of doing it in the application thread, the `Session` object will have a `SingleThreadExecutor` to handle sending of the messages on a background thread. The advantage of doing that is the application thread is not blocked by sending an async message, it simply enqueues the message to the work queue (ExecutorService) then proceeds. Allowing it to batch send a lot of async messages at a time without blocking.
```
clebertsuconic, line 186: each send on the executor would still need a round trip to the server on this case.
you are only unblocking the sender's thread., right?

clebertsuconic: wouldn't be possible to have a completion callback on the storage, sending responses to the client upon callback?

PR #1398, opened 2025-02-25
AMQ-9644: Don't allow create a queue when browsing (force to create the queue first) in the webconsole

jbonofre:


mattrpav (approved):
Confirmed the change works as designed by building this PR locally.

kenliao94 (approved):

PR #1400, opened 2025-02-26
[AMQ-9643] Add design doc for supporting hashed credential in PropertiesLoginModule

kenliao94:
Design doc for supporting AMQ-9643

### DIFF
--- a/docs/proposal/support_hashing_in_jaas_module/design_doc.md
+++ b/docs/proposal/support_hashing_in_jaas_module/design_doc.md
@@ -1,0 +1,105 @@
+# \[ActiveMQ Classic\] Support hashed credential in PropertiesLoginModule
+
+## Problem statement
+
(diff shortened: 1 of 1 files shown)


jbonofre:
I would suggest to mimic what I did in Apache Karaf (https://github.com/apache/karaf/blob/main/jaas/modules/src/main/java/org/apache/karaf/jaas/modules/encryption/EncryptionSupport.java).
Basically, I have on thread that scan the properties file, and if it detects non encrypted password (e.g without the prefix), then it does the encryption and "flag" the password with the prefix (indicating it's a hash).

kenliao94:
> I would suggest to mimic what I did in Apache Karaf (https://github.com/apache/karaf/blob/main/jaas/modules/src/main/java/org/apache/karaf/jaas/modules/encryption/EncryptionSupport.java).
Basically, I have on thread that scan the properties file, and if it detects non encrypted password (e.g without the prefix), then it does the encryption and "flag" the password with the prefix (indicating it's a hash).

Hey @jbonofre thanks for the review. I am not following. If a password doesn't have the prefix, then we assume it is just plaintext.

PR #1402, opened 2025-02-27
[AMQ-9672] Add auto start config flag for connectors

mattrpav:


cshannon:
Oops, I forgot to add that license header (I fixed my IDE to add it in the future), i pushed up a commit to main, 6.1.x and 5.19.x to fix the test class and add in the license

cshannon (changes requested):
I think it makes more sense to make the property the opposite and simplify the naming.

just call it something like `autoStart` and default it to true.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -2735,7 +2743,9 @@ public TransportConnector startTransportConnector(TransportConnector connector)
             connector = registerConnectorMBean(connector);
         }
         connector.getStatistics().setEnabled(enableStatistics);
-        connector.start();
+        if(!connector.isAutoStartDisabled()) {
```
cshannon, line 2746:
```suggestion
        if(connector.isAutoStart()) {
```

As an example of what i mean

cshannon (approved):
LGTM, assuming tests are all good.

PR #1410, opened 2025-03-13
AMQ-9685 - Virtual topic name should have at least one character to avoid Exception

graben:


mattrpav:
Can this check also be added to other virtual destination types?

graben:
@mattrpav: I think virtual topics are the only with this kind of wildcard.

cshannon (changes requested):
This change doesn't make sense to me. If you try and subscribe with the exact prefix name of a virtual topic but don't include anything after the period, why would we want to allow that and not break? We _should_ throw an exception as this is very likely a mistake and the current behavior makes sense to me.

graben:
Well, like most issues it is Murphy's law who is causing misconfigurations. Once a wrong named queue is created the broker is broken and not restartable any more. This might cause full data loss if you aren't able to manipulate your persistent backend! IMHO it doesn't make sense to allow an empty pattern by wildcard that is causing an exception because it is empty. Misconfigurations happen and shouldn't have that bad impact!

cshannon:
> Once a wrong named queue is created the broker is broken and not restartable any more. 

Ah ok, that was the part I was missing. So one thing to keep in mind when creating a Jira or a PR is the more details for the motivation to help explain why the change is necessary, the better. The jira only makes mention of the exception when trying to make the consumer which makes a lot of sense.
 
I still do not see a reason to make any change here and things are working as intended because:

1. The main point I was trying to make is creating a consumer or a producer that matches the configured VT prefix and suffix, but is missing the actual topic name, is a mistake and just continuing on without any error and allowing the client to dynamically create that destination doesn't make any sense and should be blocked. We shouldn't just hide the issue and allow things to carry on.
2. The broker itself will also block you from creating a destination like this if it is wrong. If the VirtualTopicInterceptor is created and the broker config tries to create a destination like this it will break on start up and should not create it. It would also break if trying to create it with JMX or some other means so we prevent mistakes there which is good.
3. The only issue seems to be the edge case when the destination was created by mistake somehow, probably before the interceptor was configured which would mean the broker would no longer start up. While this is a problem obviously, in that case it's a pretty easy fix and there is no risk of data loss. All you need to do is disable the VirtualDestinationInterceptor configuration in the broker config and then the broker would start up fine and you could delete the bad destination before reconfiguring the interceptor.

cshannon:
I was testing things out with the different destination creation scenarios and saw the exception pop up each time so I assumed it was short circuiting the creation attempt. It was only doing so partially and leaving things half initialized, I went back and looked and it is first initializing and adding the destination to the store and then exploding with the errors when the interceptor is later applied which is why you are seeing the errors.

The code first creates the destination including adding it to the datastore [here](https://github.com/apache/activemq/blob/9964c95ffef07c73ee4d702c817da6642ff1938e/activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java#L157) and later when the interceptor wraps the destination and throws an exception and blows up we run into this same problem. It's breaking at [this](https://github.com/apache/activemq/blob/9964c95ffef07c73ee4d702c817da6642ff1938e/activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java#L161) line which causes the rest of the code block not to process...so the destination is not started and it's not added into the map but the store is still created.

The real fix would be to just prevent destination creation entirely in this scenario. This is actually a bigger issue in that if any other interceptor breaks we run into the same problem. I have a small prototype I'm working on that I will push up and see what you think.

cshannon:
@graben - Ok I think you win, i made a proof of concept that catches the error and cleans up the initialized destination in the store but the issue that I realized is that there might be existing data that we could be blowing away as it may not be a brand new destination. It would be too dangerous to assume any error means we can just go dispose of the destination in the store as we don't know the exact issue. In this case the error is fatal on creation, but there could be lots of other interceptors that might have a more transient error that could be recovered from.

In the grand scheme of things, your change is probably fine. The side effect is if someone creates the wrong destination name with a missing virtual topic name, it's not great but there's not really any harm other then you end up with an extra destination with a typo and at least that is easily recovered from.

cshannon (approved):

PR #1411, opened 2025-03-14
[AMQ-9673] refactored JDBCIOExceptionHandlerMockeryTest to use Mockito and re-enabled the test

kenliao94:
The original test used JMock and it has issue with newer JDK. Refactored it with Mockito to modernize this test and re-enabled it


kenliao94:
@mattrpav

cshannon (approved):

PR #1413, opened 2025-03-17
[AMQ-9683] java.io.IOException on idle HTTP/XA connection

colinwerner:
As described in https://issues.apache.org/jira/browse/AMQ-9683 an EMPTY payload was received in response to a GET, as an intermediate POST sent a shutdown.

This PR implements solution 2, where we simply do not send an empty payload.

Solution 1 is in https://github.com/apache/activemq/pull/1414. We only need one or the other...


jbonofre (changes requested):
I don't think this change make sense: the logic looks good to me here (if the package received is a shutdown, we just exist the `doGet` method on the tunnel servlet).

Imho, #1414 is better.

colinwerner:
Agreed, will close this once #1414 is merged.

jbonofre:
Fixed with #1414

PR #1414, opened 2025-03-17
[AMQ-9683] java.io.IOException on idle HTTP/XA connection

colinwerner:
As described in https://issues.apache.org/jira/browse/AMQ-9683 an EMPTY payload was received in response to a GET, as an intermediate POST sent a shutdown.

This PR implements solution 1, where we ignore the exception IFF we are shutting down or we are shutdown.

Solution 2 is in https://github.com/apache/activemq/pull/1413. We only need one or the other...

### DIFF
--- a/activemq-http/src/main/java/org/apache/activemq/transport/http/HttpClientTransport.java
+++ b/activemq-http/src/main/java/org/apache/activemq/transport/http/HttpClientTransport.java
@@ -206,7 +206,11 @@ public class HttpClientTransport extends HttpTransportSupport {
                     stream.close();
                 }
             } catch (Exception e) { // handle RuntimeException from unmarshal
-                onException(IOExceptionSupport.create("Failed to perform GET on: " + remoteUrl + " Reason: " + e.getMessage(), e));
+                if (isStopped() || isStopping()) {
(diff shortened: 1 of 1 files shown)


jbonofre (approved):

jbonofre:
I think some tests can break with this change. I will double check.

colinwerner:
@jbonofre any luck here?

jbonofre:
@colinwerner on it 😄

jbonofre:
I'm checking one test that can be problematic (but not a blocker).

cshannon (approved):
I think this makes sense, if the transport is already stopping or stopped then exceptions are expected so probably best to log it vs throwing an exception, as long as the empty response is also handled without issue on the return and the tests are all passing.

mattrpav:
IMO: This is the _ideal_ scenario to add a counter metric. Logging can get clogged up/rotated/etc and having a counter makes it easy to monitor if this is happening in a on-off, or consistently.

jbonofre:
I think this first PR is fine as a quick fix. We can always improve later. The debug log message is good enough for now.

colinwerner:
Thank you @jbonofre -- much appreciated! I'll follow-up if we see any other issues!

PR #1418, opened 2025-03-30
Fix antrun for openwire-generate and allow (require) JDK8 for now

jbonofre:


jbonofre:
@kenliao94 as discussed on the mailing list.

kenliao94:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (default-cli) on project activemq-client: An Ant BuildException has occured: java.lang.UnsupportedClassVersionError: org/apache/activemq/openwire/tool/JavaGeneratorTask has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
[ERROR] around Ant part ...<taskdef classname="org.apache.activemq.openwire.tool.JavaGeneratorTask" name="generate" classpathref="maven.compile.classpath" />... @ 5:135 in /Users/qrl/amzn/personal_workspace/activemq/activemq-client/target/antrun/build-main.xml
[ERROR] -> [Help 1]
org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.apache.maven.plugins:maven-antrun-plugin:3.1.0:run (default-cli) on project activemq-client: An Ant BuildException has occured: java.lang.UnsupportedClassVersionError: org/apache/activemq/openwire/tool/JavaGeneratorTask has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 52.0
around Ant part ...<taskdef classname="org.apache.activemq.openwire.tool.JavaGeneratorTask" name="generate" classpathref="maven.compile.classpath" />... @ 5:135 in /Users/qrl/amzn/personal_workspace/activemq/activemq-client/target/antrun/build-main.xml
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute2 (MojoExecutor.java:375)
    at org.apache.maven.lifecycle.internal.MojoExecutor.doExecute (MojoExecutor.java:351)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
```
Hey @jbonofre it might be due to my local environment but since I configured Java 17 to work on `main` branch. I seem to can't get rid of this error. I tried to use mvn cli and configured the java_home to point to java 8 and it showed me this error. Do I need to be 5.16 branch and configure java 8 to generate?

jbonofre:
As you can see in my change, jdk8 is still required here. I will update to jdk17+ in a separate PR.

cshannon:
As mentioned on the dev list and on the Jira Issue, there is a new project for future OpenWire versions that is being modernized. It's fine to fix this to require JDK 8, but I don't think it's necessary or make sense to bother with fixing the current version for JDK 17.

I think we should probably just remove the old generator entirely from the 6.x build as it doesn't work on newer JDKs and we plan to migrate to the new project for v13 and future updates.

jbonofre:
I rebased. This is a simple change just to set JDK 8 for OpenWire generator (just in case someone would like to build it).

activemq-openwire-generator/pom.xml
```diff
@@ -51,6 +55,39 @@
...
+          </execution>
+          <execution>
+            <id>enforce-java-version</id>
+            <goals>
+              <goal>enforce</goal>
+            </goals>
+            <configuration>
+              <rules>
+                <requireJavaVersion>
+                  <version>[8,)</version>
```
cshannon, line 83: This looks like it will allow java 8 or higher, is that what you want? Doesn't this break when running with newer JDKs which was the problem?

jbonofre: That's a good point. We should probably be more "restrictive" here. Let me update.

PR #1419, opened 2025-04-10
AMQ-9689 - Network of Broker durable sync TTL fixes and improvements

cshannon:
This commit makes several improvements and fixes for syncing durable subscriptions when a network bridge connects.

1. A bug was fixed during durable sync that would cause the clientId to not always be included for durables in the subscription list which could cause a loop to be created as the other broker would not be able to tell where the network subscription came from.
2. During reactivation when dynamicOnly is false and durable sync is set to true, we make sure to include the TTL information (full broker path) from the online consumer attached to durables so that TTL info is properly
 propagated so we don't incorrectly create demand. Thisonly works if
consumers are online, so for TTL > 1 it is still recommended to set dynamicOnly to true and allow only online consumers drive demand.
3. For TTL 1, we can handle sync correctly on restarts even if durables are offline and missing consumer TTL info because we know that we should ignore proxy durables (bridge durables for other bridges) entirely because they will be > 1 hop away.
4. Some other minor improvements were made like filtering everything if TTL is 0 and also consolidating logic.


activemq-broker/src/main/java/org/apache/activemq/util/NetworkBridgeUtils.java
```diff
@@ -16,10 +16,7 @@
  */
 package org.apache.activemq.util;
 
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
```
mattrpav, line 19: import pkg.*

activemq-unit-tests/src/test/java/org/apache/activemq/JmsMultipleBrokersTestSupport.java
```diff
@@ -105,7 +105,7 @@ protected NetworkConnector bridgeBrokers(String localBrokerName, String remoteBr
 
     // Overwrite this method to specify how you want to bridge the two brokers
     // By default, bridge them using add network connector of the local broker
-    // and the first connector of the remote broker
+    // and the first connector of the remote brokerz
```
mattrpav, line 108: looks like type-o 'brokerz' vs preferred 'broker' in the comment

cshannon:
Looks like all the tests have passed

PR #1423, opened 2025-05-05
AMQ-9698 - Fix message expiration on durable subs

cshannon:
This commit fixes multiple problems with handling message expiration on durable topic subscriptions.

1) Memory usage tracking is fixed on expiration by correctly decrementing the counter inside AbstractStoreCursor when calling the remove(message) method, which was previously missed. 2) A new refrence type is used to wrap references in TopicStorePrefetch so that if multiple subscriptions share a reference in their cursors each one can expire the message. Previously only one would expire as the message would be marked as expired and skipped.
3) On client expiration, the references are properly decremented so memory tracking is correct.
4) The expiration thread for Topics has been improved to be much more efficient for KahaDB by only scanning for expired messages if there are durables eligible for expiration. The thread also now checks the index to see if expired messages are associated with the subs still so we don't expire the same sub multiple times. Only messages that need to still be processed are returned which further cuts down memory usage.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Topic.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Topic.java
@@ -21,6 +21,8 @@ import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -29,6 +31,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantReadWriteLock;
 
+import java.util.stream.Collectors;
 import org.apache.activemq.advisory.AdvisorySupport;
 import org.apache.activemq.broker.BrokerService;
 import org.apache.activemq.broker.ConnectionContext;
@@ -53,6 +56,7 @@ import org.apache.activemq.filter.MessageEvaluationContext;
 import org.apache.activemq.filter.NonCachedMessageEvaluationContext;
 import org.apache.activemq.management.MessageFlowStats;
 import org.apache.activemq.store.MessageRecoveryListener;
+import org.apache.activemq.store.MessageStore.StoreType;
 import org.apache.activemq.store.NoLocalSubscriptionAware;
 import org.apache.activemq.store.PersistenceAdapter;
 import org.apache.activemq.store.TopicMessageStore;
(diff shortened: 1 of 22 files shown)


cshannon:
This fix should be complete and tests have been updated/written but I still plan to add a couple more, specifically to make sure the KahaDB new method works correctly that is used by the expiration thread. I went ahead and pushed this up to get the full build to run and make sure the current tests work.

activemq-broker/src/main/java/org/apache/activemq/broker/region/Topic.java
```diff
@@ -826,14 +830,51 @@ protected void dispatch(final ConnectionContext context, Message message) throws
...
     private final AtomicBoolean expiryTaskInProgress = new AtomicBoolean(false);
-    private final Runnable expireMessagesWork = new Runnable() {
-        @Override
-        public void run() {
-            List<Message> browsedMessages = new InsertionCountList<Message>();
-            doBrowse(browsedMessages, getMaxExpirePageSize());
+    private final Runnable expireMessagesWork = () -> {
+        try {
+            final TopicMessageStore store = Topic.this.topicStore;
+            if (store != null && store.getType() == StoreType.KAHADB) {
```
mattrpav, line 836: Would it be suitable to check the class instanceof/isassingable/isinstance vs having the type flag? Anyone extending KahaDB would still have this functionality.

cshannon: @mattrpav - I actually forgot that even ignoring the proxy stuff (Because you could check delegates) it's not possible to do an instance of or class check. The code lives in the activemq-broker module and that module has no dependency on the data stores like KahaDB. So the KahaDB specific interfaces are not even on the classpath at compilation time.

mattrpav: That makes sense, thanks.

activemq-broker/src/main/java/org/apache/activemq/broker/region/Topic.java
```diff
@@ -826,14 +830,61 @@ protected void dispatch(final ConnectionContext context, Message message) throws
     }
 
     private final AtomicBoolean expiryTaskInProgress = new AtomicBoolean(false);
-    private final Runnable expireMessagesWork = new Runnable() {
-        @Override
-        public void run() {
-            List<Message> browsedMessages = new InsertionCountList<Message>();
-            doBrowse(browsedMessages, getMaxExpirePageSize());
+    private final Runnable expireMessagesWork = () -> {
+        try {
```
mattrpav, line 834: How about adding a fast-exit check here if there are zero messages and/or zero offline durable subscribers?

cshannon: See https://github.com/apache/activemq/pull/1423/commits/baff56ed8a4aaf0eb849f771c438284f79f1adfc

mattrpav: Great!

PR #1425, opened 2025-05-07
add instructions to copy jetty.xml to example/conf to access web demos

gurpartap0306:
### Summary
This PR addresses a configuration issue that caused the demo to fail during startup due to an XML parsing error in jetty.xml.

### Issue Fixed

Before this change, starting the demo threw the following error:

``` 
ERROR | Failed to load: file [/Users/user/workspace/activemq-brokers/apache-activemq-6.1.6/examples/conf/activemq-demo.xml], reason: IOException parsing XML document from file [/Users/user/workspace/activemq-brokers/apache-activemq-6.1.6/examples/conf/jetty.xml]
```
This was caused by the application not being able to resolve jetty.xml from the filesystem path. ~~By updating the reference to use the classpath and ensuring the correct path is included, this issue is now resolved.~~

Added instructions to copy `jetty.xml` to `example/conf` to access web demos that fixes the issue.

### DIFF
--- a/assembly/src/release/docs/user-guide.html
+++ b/assembly/src/release/docs/user-guide.html
@@ -97,6 +97,7 @@ You would need to start the broker with the demos included, which you do as foll
 If you're using Windows, just type
 </p>
 <pre>
+    copy .\conf\jetty.xml .\examples\conf\
     cd bin
     activemq.bat start xbean:../examples/conf/activemq-demo.xml
 </pre>
@@ -104,6 +105,7 @@ If you're using Windows, just type
 On Unix-like systems, type
 </p>
 <pre>
+    cp ./conf/jetty.xml ./examples/conf/
     ./bin/activemq console xbean:examples/conf/activemq-demo.xml
 </pre>
 


jbonofre (changes requested):
I don't think we should add `conf` in classpath.

The default `activemq.xml` using related path (not classpath). Demo can do the same.

gurpartap0306:
Thanks for the feedback! Initially, we used a relative path like `<import resource="jetty.xml" />`, which worked during testing because the test setup moved the demo files from `example/conf` to `conf`. However, this caused the DemoWebApp to fail at runtime since `jetty.xml` doesn't exist in `example/conf`.

To resolve this, I added `conf` to the test classpath to align it with the runtime environment. I did not modify the runtime classpath, keeping the production setup unchanged.

Other option is to add copy of `jetty.xml` in `example/conf` and use relative path. Let me know if you'd prefer this approach.

jbonofre:
@gurpartap0306 yeah, I would prefer to copy (or document the copy) `jetty.xml` in `example/conf`. I don't want to add `conf` in the classpath (as it can be problematic for security, side effect, etc).

jbonofre (approved):

PR #1428, opened 2025-05-12
AMQ-9697: Removed inline JS and CSS from the Web Console. Added CSP header to jetty.xml

sergio-d-lemos:
[AMQ-9697]: Moves inline CSS and Javascript scripts into dedicated files in the Web Console. The changes allow to enable the Content Security Policy header in jetty.xml such as inline scripts are blocked, offering another layer of protection agains script injections.

I tested the changes by accessing all the impacted pages. Also tested the copy, move and delete messages functionalities.


jbonofre:
I did additional tests and a new pass, it's all good. Thanks !

PR #1441, opened 2025-05-26
AMQ-9716: Fix `maxMessageSize=-1` to correctly disable message size limit

therepanic:
This PR fixes a regression introduced in 5.19.0 where setting `maxMessageSize=-1` no longer disables message size validation, despite being [documented](https://activemq.apache.org/components/classic/documentation/rest) to do so.

The fix ensures that checks are only enforced if `maxMessageSize >= 0`.


mattrpav (approved):
LGTM

PR #1444, opened 2025-06-01
AMQ-9718 Removed collections which are never queried

dk2k:
I believe, if the changed code is buildable and existing tests pass, then it's enough for this PR

### DIFF
--- a/activemq-broker/src/test/java/org/apache/activemq/store/PListTestSupport.java
+++ b/activemq-broker/src/test/java/org/apache/activemq/store/PListTestSupport.java
@@ -128,11 +128,9 @@ public abstract class PListTestSupport {
     }
 
     protected void doTestRemove(final int COUNT) throws IOException {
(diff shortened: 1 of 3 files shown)


jbonofre:
Thanks for the PR. Let me take a look.

dk2k:
@cshannon please review

cshannon:
This needs to be looked at more to see why the collection is not used on browse, i see in the one it's setting a selector but not queried. I have very little experience using the console so i'm not familiar why it's not being used and if we should fix the command to use it again so i'll see what @jbonofre thinks.

jbonofre:
Yeah I will take a deeper look today (sorry I was travelling).

jbonofre (approved):
I did a more complete review and it looks to me.
1. in `PListTestSupport#doTestRemove()` is only used in `testRemove()` and `testDestroy()` and do not actually use the "inner" map.
2. `BrowseCommand` and `PurgeCommand` are similar. They could use sub objects with the `--xmsgsel` argument, but the list is not actually used in this context.

PR #1445, opened 2025-06-02
AMQ-9719 Removed 2 unused stringbuffers

dk2k:


activemq-client/src/main/java/org/apache/activemq/filter/XPathExpression.java
```diff
@@ -120,17 +120,6 @@ protected static void setupFeatures(DocumentBuilderFactory factory) {
                 }
             }
         }
-        if (features.size() > 0) {
```
cshannon, line 123: Instead of removing this, I think it makes more sense to add an actual log statement (probably debug level) as based on the comment that was the intent of the loop. If we did that we should probably do a check to see if the log level is enabled before doing the actual iteration and building of the buffer to avoid the performance hit if not logging of building the string. I would also switch to using StringBuilder as well.

activemq-client/src/main/java/org/apache/activemq/filter/XPathExpression.java
```diff
@@ -120,15 +121,13 @@ protected static void setupFeatures(DocumentBuilderFactory factory) {
...
-            StringBuilder featureString = new StringBuilder();
+        StringJoiner featureString = new StringJoiner(", ");
```
cshannon: You could move this inside the if statement so we are not constructing a new StringJoiner unless we need it

cshannon (approved):

cshannon:
I pushed a follow on commit as I think it would be better to use trace logging as it would be super verbose otherwise as it would log on every dispatch and previously it was commented out anyways. See https://github.com/apache/activemq/commit/932ac01c0a82cf7c58e5461b50a19f4d60b6f69f

PR #1446, opened 2025-06-02
Code cleanup: use StringBuilder instead of StringBuffer

grigoni:
cleanup: use StringBuilder instead of StringBuffer
StringBuilder is not synchronized, which makes it faster than StringBuffer
in some cases where there is no loop involved replace StringBuffer with a pure String concatenation (leaving java compiler doing the optimizations)


cshannon (approved):
LGTM, thanks for the PR, these changes make sense to switch to StringBuilder when we need it and as you pointed out the compiler can optimize string concatenation when not inside a loop, so it makes a lot of sense to simplify and remove the usage of StringBuffer in those spots as you did.

Do you have permission or the ability to create a new Jira? We should be tracking any changes like this against a [Jira](https://issues.apache.org/jira/projects/AMQ/issues) before we merge.

grigoni:
@cshannon thx a lot for reviewing the pr
I've already created the jira but only after this pr:

https://issues.apache.org/jira/browse/AMQ-9720

PR #1447, opened 2025-06-03
AMQ-9721 - Fix performance issues during non-persistent cursor removal

cshannon:
This fixes the broker so multiple removals are no longer done for the same message leading to having to search the entire non persistent pending list. Durable subscriptions now check the persistence type of the message so the cursor will no longer search everything in a non-persistent pending list when the message is persistent.

Since the fixes are primarily performance related it's a little bit tricky to test but  I added one new test for the durable sub cursor removal and the existing tests should cover everything else including memory usage being correct (such as the test added recently in AMQ-9698)


cshannon:
The build failures look like flaky tests (i ran locally and they passed)

PR #1449, opened 2025-06-04
AMQ-9723 implemented try with resources for socket

dk2k:
Not sure what to return in the end of method - used value -1 for now.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/transport/tcp/QualityOfServiceUtils.java
+++ b/activemq-client/src/main/java/org/apache/activemq/transport/tcp/QualityOfServiceUtils.java
@@ -17,6 +17,7 @@
 package org.apache.activemq.transport.tcp;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.Socket;
 import java.net.SocketException;
 import java.util.HashMap;
@@ -135,7 +136,7 @@ public class QualityOfServiceUtils {
             throw new IllegalArgumentException("Setting Differentiated Services"
                 + " not supported: " + e);
         } catch (IOException e) {
-            throw new IllegalArgumentException("Failed to close the socket: " + e);
+            throw new UncheckedIOException(e);
         }
     }
 }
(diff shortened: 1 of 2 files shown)


activemq-client/src/main/java/org/apache/activemq/transport/tcp/QualityOfServiceUtils.java
```diff
@@ -125,15 +126,17 @@ private static int adjustDSCPForECN(int dscp)
...
-        try {
+        try (Socket socket = new Socket()) {
             int systemTrafficClass = socket.getTrafficClass();
             // The 1st and 2nd bits of the system traffic class are the ECN
             // bits.
             return (dscp << 2) | (systemTrafficClass & 3);
         } catch (SocketException e) {
             throw new IllegalArgumentException("Setting Differentiated Services"
                 + " not supported: " + e);
+        } catch (IOException e) {
```
jbonofre, line 138: Why not returning IAE here ?

jbonofre (changes requested):
Thanks again. Generally speaking, for this kind of "code style update", can we group in one ? I will include only on `main` for now. Thoughts ?

dk2k:
@jbonofre  It's not code style, it's static code analysis.
From my experience I prefer separate PRs for different issues. If they are atomic, it's easier to get approvals and have them merge. I can use jira ticket with title "Fix issues found by static code analysis" if you don't mind (it's not created yet)

jbonofre:
Yes that would be better.

activemq-client/src/main/java/org/apache/activemq/transport/tcp/QualityOfServiceUtils.java
dk2k: done

cshannon: I don't think IAE makes sense, there should be a different exception as it's not an illegal argument so it's a misleading exception to throw.

If the goal is to simply make the exception unchecked then the best thing to do is just wrap it in [UncheckedIOException](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/io/UncheckedIOException.html)

dk2k: done

PR #1452, opened 2025-06-04
AMQ-9726 - Fix FilePendingMessageCursor clear() method

cshannon:
This fixes the clear() method so that when clearing the memory map it will decrement memory usage, and when clearing the disk list it will destroy and reset the list for future writes.


cshannon:
Test failure is flaky test that has been fixed in: https://github.com/apache/activemq/commit/6581ed7e74cf1ad9ca95aa192a2f6cd05a464a5d

PR #1455, opened 2025-06-13
AMQ-9727 string should be compared with equals()

dk2k:


jbonofre (approved):

jbonofre:
It looks good to me but I want to check no side effect in the generator.

PR #1461, opened 2025-06-14
More Code cleanup: use StringBuilder instead of StringBuffer

grigoni:
second round about StringBuffer usage
use StringBuilder instead of StringBuffer: this avoids unnecessary sync
use StringBuilder best practice when concatenating
avoid concatenation in logging
use `java.nio.charset.StandardCharsets` for getBytes
use isEmpty() instead of `...length() > 0 `
some typos fixing

see https://issues.apache.org/jira/browse/AMQ-9732

### DIFF
--- a/activemq-amqp/src/test/java/org/apache/activemq/transport/amqp/client/sasl/CramMD5Mechanism.java
+++ b/activemq-amqp/src/test/java/org/apache/activemq/transport/amqp/client/sasl/CramMD5Mechanism.java
@@ -16,13 +16,14 @@
  */
 package org.apache.activemq.transport.amqp.client.sasl;
 
-import java.io.UnsupportedEncodingException;
 import java.security.InvalidKeyException;
 import java.security.NoSuchAlgorithmException;
 
 import javax.crypto.Mac;
 import javax.crypto.spec.SecretKeySpec;
 import javax.security.sasl.SaslException;
+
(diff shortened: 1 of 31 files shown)


grigoni:
hello @jbonofre , @cshannon 
my intention with this PR was to complete StringBuffer cleanup started with AMQ-9720
do you have a chance to review it?
thx

jbonofre:
@grigoni yes I started to review it. Generally speaking, I'm not a big fan of large PR for code cleanup (when generated by tool or not). It's long to review (it touches different part of the code) and side effects can happen (for instance, using a not expected locale, ...).

So, instead of a super large PR changing everything at once, I would split into smaller chuncks.

That said for this PR, at first glance, it looks good, but I still have to do a new pass.

grigoni:
@jbonofre 
fully agree about large PR, if you prefer I'll split/remove some parts

jbonofre:
@grigoni it's ok. Don't worry. I will do a pass.

mattrpav:
Breaking up the commits by type of fix and maven module is helpful in the event something need to be reverted.

grigoni:
@mattrpav splitting makes definetely sense 
I've rolled back the commit in my branch and **with same code** created a new one
which has commits by module and/or by fix type where changes were more specific
see: https://github.com/apache/activemq/pull/1471

grigoni:
closing this PR, current code base does not need this changes

PR #1464, opened 2025-07-01
[AMQ-9737] activemq-spring: Remove references to schema namespaces pre-6.0.0

mattrpav:


jbonofre (changes requested):
The same should be done in activemq-osgi module.

mattrpav:
> The same should be done in activemq-osgi module.

Good catch. I modified the activemq-osgi module to pull the spring.schemas file from the activemq-spring module (as we do for the schemas) so we only need to manage it in one place going forward.

jbonofre (approved):

PR #1465, opened 2025-07-02
Update yield to use Thread.yield

harmstyler:
Unqualified call to 'yield' method is not supported in releases since Java 14, update it to use `Thread.yield`.


mattrpav:
Addressed in [#1172] as part of Tech Preview support for Virtual Threads

harmstyler:
@mattrpav will that change make it to the 6.x branch? I'm working on writing up a bug I've found in the network connectors runtime config changing. I would like to provide a failing test. Should I be working in the main branch, then?

mattrpav:
@harmstyler Please make a patch for _main_. Depending on the nature of the bug and the fix, we may want to backport it from _main_ to the other active branches (activemq-6.1.x and activemq-5.19.x).

jbonofre:
I second @mattrpav here: the PR should open based on the `main` branch (and the team does the cherry pick on the branches where it makes sense).

harmstyler:
got it, thank you!

PR #1466, opened 2025-07-02
Remove requirement for network connector order in runtime configuration

harmstyler:
This change updates the runtime configuration processing for network connectors to remove the dependency on the order of connectors in the configuration.
Network connectors are now matched by object instead of relying on their position in the list.
This improves reliability when adding, modifying, or removing network connectors via configuration updates.
Includes test coverage for scenarios with reordered and removed connectors.


jbonofre (approved):
I reviewed and tested this change: it looks good to me.
Thanks also to have updated the tests.
Great work !

PR #1469, opened 2025-07-04
AMQ-9740: Fixes "org.apache.commons.lang3.StringEscapeUtils cannot be resolved to a type" in the  Web Console.

sergio-d-lemos:
The Web Console depends on `commons-lang3`, this commit adds back the Maven dependency and updates the commons-lang3 to 3.17.0. It also updates it to version 3.17.0 (latest)

The dependency on `commons-lang3` was removed in https://github.com/apache/activemq/commit/f5014221d027b71e842988cf6a3b23e2feed4d18. However it seems to cause issues in the Web Console, were some pages cause the error:

```
2025-07-03 19:43:34,007 | WARN  | /admin/send.jsp | org.eclipse.jetty.server.HttpChannel | qtp1910813448-74
jakarta.servlet.ServletException: jakarta.servlet.ServletException: org.apache.jasper.JasperException: Unable to compile class for JSP:An error occurred at line: [27] in the jsp file: [/WEB-INF/tags/form/text.tag]
org.apache.commons.lang3.StringEscapeUtils cannot be resolved to a type
24:     if (value == null) {
25:         value = "";
26:     }
27:     value = org.apache.commons.lang3.StringEscapeUtils.escapeHtml4(value);
28:
29: %>
30: <input type="text" name="${name}" value="<%= value %>"/>
```

`commons-lang3` is still used in a couple of tags: https://github.com/search?q=repo%3Aapache%2Factivemq%20%20org.apache.commons.lang3&type=code

Please let me know if there is a more appropriate way to add the dependency.

### DIFF
--- a/activemq-web-console/src/main/webapp/WEB-INF/tags/form/short.tag
+++ b/activemq-web-console/src/main/webapp/WEB-INF/tags/form/short.tag
@@ -14,16 +14,15 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 --%>
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
 <%@ attribute name="text" type="java.lang.String" required="true"  %>
 <%@ attribute name="length" type="java.lang.Integer" required="false" %>
 <%
- text = org.apache.commons.lang3.StringEscapeUtils.escapeHtml4(text);
- text = org.apache.commons.lang3.StringEscapeUtils.escapeEcmaScript(text);
- if (length == null || length < 20)
-    length = 20;
- if (text.length() <= length) {
-     out.print(text);
- } else {
-     out.println(text.substring(0, (length-10)) + "..." + text.substring(text.length() - 5));
+ if (length == null || length < 20) {
+     length = 20;
+ }
+ if (text.length() > length) {
+     text = text.substring(0, (length-10)) + "..." + text.substring(text.length() - 5);
  }
 %>
+<c:out value="<%= text %>" />
--- a/activemq-web-console/src/main/webapp/WEB-INF/tags/form/text.tag
+++ b/activemq-web-console/src/main/webapp/WEB-INF/tags/form/text.tag
@@ -14,6 +14,7 @@
     See the License for the specific language governing permissions and
     limitations under the License.
 --%>
+<%@ taglib prefix="c" uri="http://java.sun.com/jsp/jstl/core" %>
 <%@ attribute name="name" type="java.lang.String" required="true"  %>
 <%@ attribute name="defaultValue" type="java.lang.String" required="false"  %>
 <%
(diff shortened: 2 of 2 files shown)


mattrpav:
I’d prefer not to add the entire dependency for one or two methods.

sergio-d-lemos:
> I’d prefer not to add the entire dependency for one or two methods.

@mattrpav Sorry, I'm not sure if I updated my pull request correctly. I now replaced the code using `StringEscapeUtils` with `c:out` tags which already do the sanitization. This way we don't need to depend on commons-lang3.

I tested this change by:

1.  Tested the `form:text` tag by trying to inject javascript code into the "send" page via URL parameters: `http://localhost:8161/admin/send.jsp?JMSCorrelationID=a%22%3C\input%3E%3Cscript%3Ealert(%22test%22);%3C/script%3E` 
1. Tested the `form:short` tag by using the send page to publish messages to queues called `<script>alert('a');</script>` and `<script>alert('b');</script>`, then opening one of the messages.

jbonofre:
Agree with @mattrpav, regarding the simple code, we can implement the same logic easily with a homemade method. No need to have the whole commons-lang3 dep just for this.

sergio-d-lemos:
@jbonofre what do you think of my suggestion? Escaping HTML is a common task, I'd suggest reusing any implementation we already have if it is possible.

mattrpav:
@sergio-d-lemos please update title and JIRA title to reflect the change

mattrpav (approved):
I like the approach of using the a native feature of the web framework. I'm not an expert on it, and would like to see a couple extra set of eyes on it.

jbonofre (approved):
The approach makes sense to me and cleaner than adding the commons-lang deps. Using `<c:out/>` should do the equivalent escape.

coheigea:
@jbonofre @mattrpav We are still shipping ./lib/optional/commons-lang3-3.14.0.jar in the distribution, as it's a dependency in the console. Should it be updated or removed as it has an existing CVE against it?

jbonofre:
@coheigea let me take a look. You should at least update.

coheigea:
> @coheigea let me take a look. You should at least update.

@jbonofre I did submit this PR then closed it https://github.com/apache/activemq/pull/1483 - should I re-open?

mattrpav:
@coheigea yes, go ahead and reopen that PR against 6.1.x. 

6.2.x - commons-lang3 removed
6.1.x - upgrade for security fix
5.19.x - upgrade for security fix

jbonofre:
Yeah, agree with @mattrpav it's a good plan.

coheigea:
The JIRA has a 6.1.8 as the affects version and 6.1.9/6.2.0 as the fix version, but the PR to remove commons-lang and velocity was only done on main right? I don't see any problems with the web console on the current activemq-6.1.x branch.

mattrpav:
@coheigea the issue only impacted 'main' branch. At the time of JIRA reporting, the main may have been versioned 6.1.8-SNAPSHOT. I cleaned-up the _affected_ and _fixed_ fields in the JIRA.

Managing issue-to-PR alignment would be more straight-forward if we used GH issues ;-)

PR #1474, opened 2025-07-23
AMQ-9747 - Handle IOExceptionHandler thrown exceptions in KahaDB

cshannon:
Update the KahaDB scheduled tasks to catch any runtime exceptions thrown by the configured IOExceptionHandler. This will prevent the tasks from being killed and no longer running if the IOExceptionHandler does not shut down the broker.

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
@@ -448,10 +448,10 @@ public abstract class MessageDatabase extends ServiceSupport implements BrokerSe
                 }
             } catch (IOException ioe) {
                 LOG.error("Checkpoint failed", ioe);
-                brokerService.handleIOException(ioe);
+                handleIOException("CheckpointRunner", ioe);
             } catch (Throwable e) {
                 LOG.error("Checkpoint failed", e);
-                brokerService.handleIOException(IOExceptionSupport.create(e));
+                handleIOException("CheckpointRunner", IOExceptionSupport.create(e));
             }
         }
     }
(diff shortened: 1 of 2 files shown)


mattrpav:
This looks like a good set of changes to add robustness and the ability for the broker to recover from some IO errors without restart.

I see the store() method has the handler-- any reason to not also add it to the load(Location )?

ref: https://github.com/apache/activemq/blob/18e9b34dfe7e3227baf0be122a8db54020f48014/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java#L1214

Seems like a _read_ is as (or more) recoverable than the store -- especially for network / cloud filesystems, etc.

mattrpav:
> This looks like a good set of changes to add robustness and the ability for the broker to recover from some IO errors without restart.
> 
> I see the store() method has the handler-- any reason to not also add it to the load(Location )?
> 
> ref:
> 
> https://github.com/apache/activemq/blob/18e9b34dfe7e3227baf0be122a8db54020f48014/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java#L1214
> 
> Seems like a _read_ is as (or more) recoverable than the store -- especially for network / cloud filesystems, etc.

Noting that no changes are needed -- any change to make more recoverable without restart should be a follow-on task / feature change.

This change is about not terminating kahadb sub-tasks only

mattrpav (approved):
LGTM. Approve for merge when tests go 'green'

cshannon:
> This looks like a good set of changes to add robustness and the ability for the broker to recover from some IO errors without restart.
> 
> I see the store() method has the handler-- any reason to not also add it to the load(Location )?
> 
> ref:
> 
> https://github.com/apache/activemq/blob/18e9b34dfe7e3227baf0be122a8db54020f48014/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java#L1214
> 
> Seems like a _read_ is as (or more) recoverable than the store -- especially for network / cloud filesystems, etc.

The goal of this PR is only to handle the runtime exception that migh be killing the tasks in KahaDB, not to try and change how IOExceptions are propagated or handled in general. So any other changes in terms of when we throw IO errors and what is and isn't recoverable would be a separate issue and outside the scope of this change

cshannon:
> The goal of this PR is only to handle the runtime exception that migh be killing the tasks in KahaDB, not to try and change how IOExceptions are propagated or handled in general. So any other changes in terms of when we throw IO errors and what is and isn't recoverable would be a separate issue and outside the scope of this change

Oops I just saw you already responded with the same thing :)

PR #1477, opened 2025-07-29
WIP: [AMQ-9750] Update JaasDualAuthenticationBroker to support mixed-mode SSL

mattrpav:


cshannon:
Instead of having to add a new flag, couldn't you just check the ssl config for the transport and see if it's set to require a cert or if the cert is optional?

mattrpav:
Great idea. I'll update if 'wantAuth' or 'needAuth' AND then if there is a certificate present, return true-- otherwise it will fallback to the non-SSL certificate JAAS domain.

cshannon:
Yeah that should be fine, since it will fallback to the other option. If the cert exists and the context is configured to want/need the cert then that should be good enough. This way you don't need to add another config option because it's essentially already being configured in the context by setting need/want to be true. Prevents mistakes and having to remember to change the config in the same spot twice.

PR #1478, opened 2025-08-06
AMQ-9691: Add `createdTimestamp` to `MessageStoreStatistics`

therepanic:


activemq-broker/src/main/java/org/apache/activemq/store/MessageStoreStatistics.java
```diff
@@ -28,24 +28,23 @@ public class MessageStoreStatistics extends StatsImpl {
...
 
     public MessageStoreStatistics() {
         this(true);
     }
 
     public MessageStoreStatistics(boolean enabled) {
-
         messageCount = new CountStatisticImpl("messageCount", "The number of messages in the store passing through the destination");
         messageSize = new SizeStatisticImpl("messageSize","Size of messages in the store passing through the destination");
+        createdTimestamp = System.currentTimeMillis();
```
therepanic, line 40: Should we create another constructor where the second argument will be `createdTimestamp` or the current version is OK?

mattrpav:
This timestamp will always be the same as the broker startup time as implemented in this PR. 

The intent of the JIRA is to have the timestamp be stored across reboots. There is more to change and most likely a KahaDB version bump.

therepanic:
Thanks for the quick reply, @mattrpav. I realize now that I misunderstood the scope of the task. Saving the timestamp persistently involves more complexity than I initially expected, so I think it’s best to leave this to the core team. I’ll close this PR for now. Thanks anyway!

PR #1480, opened 2025-08-15
[AMQ-9244] Expanded proposal for supporting OAuth 2.0 Authentication

sergio-d-lemos:
This PR expands the design initially proposed by @kenliao94 and @LinC0416 for supporting OAuth 2.0 authentication and @jbonofre initial work on https://github.com/apache/activemq/pull/1035 back in 2023.

### DIFF
--- a/assembly/src/release/conf/jetty.xml
+++ b/assembly/src/release/conf/jetty.xml
@@ -82,13 +82,13 @@
                 <bean id="header" class="org.eclipse.jetty.rewrite.handler.HeaderPatternRule">
                     <property name="pattern" value="*"/>
                     <property name="name" value="Content-Security-Policy"/>
-                    <property name="value" value="upgrade-insecure-requests; style-src-elem 'self'; style-src 'self'; img-src 'self'; script-src-elem 'self'; default-src 'none'; object-src 'none'; frame-ancestors 'none'; base-uri 'none';" />
+                    <property name="value" value="style-src-elem 'self'; style-src 'self'; img-src 'self'; script-src-elem 'self'; default-src 'none'; object-src 'none'; frame-ancestors 'none'; base-uri 'none';" />
                 </bean>
                 <!-- More relaxed rules to allow browsers to properly render XML -->
                 <bean id="header" class="org.eclipse.jetty.rewrite.handler.HeaderPatternRule">
                     <property name="pattern" value="/admin/xml/*"/>
                     <property name="name" value="Content-Security-Policy"/>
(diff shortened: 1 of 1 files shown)


mattrpav:
I like the idea of a standard oauth2 plugin for authn, but it could be tricky without a bunch of provider specific plugins to re-work request and response handler payloads b/c there are subtle differences. Might need that layer to be pluggable/highly configurable. 

Using scope claims for groups (ie authz) is tricky, because some providers limit the number of available groups provided in the token—- which means users can’t rely on those tokens for group membership and need to make secondary lookups for group/role info to a non-standard API. 

The flow is simple enough that it would be great if we could use the JDK http client.

mattrpav:
@sergio-d-lemos I think we can move forward with the JAAS LoginModule. Do you want to take a pass at it or would it be helpful if I drafted one out?

sergio-d-lemos:
I can publish a draft (likely over the weekend), won't be "production ready" but it's a start.

sergio-d-lemos:
Included a basic implementation. The code is still WIP with a bunch of TODOs on things which still need more work (and there are no tests), but I think it demonstrates the ideas I had in mind.

mattrpav:
@sergio-d-lemos Good start. 

I think having an embedded 'mock' OAuth2 server would be helpful. I think using the JDK http server would be pretty simple to spin up and then allow us to mock payloads and status codes.

Thoughts?

assembly/src/release/conf/login.config
```diff
@@ -18,4 +18,12 @@ activemq {
...
         org.apache.activemq.jaas.properties.group="groups.properties";
-};
\ No newline at end of file
+};
+
+OAuthConfiguration {
+    org.apache.activemq.jaas.oauth.OAuthLoginModule required
+    issuer="https://cognito-idp.us-west-2.amazonaws.com/us-west-2_iz1VrIxct"
+    jwks_uri="https://cognito-idp.us-west-2.amazonaws.com/us-west-2_iz1VrIxct/.well-known/jwks.json"
+    group_resolver_class="org.apache.activemq.jaas.oauth.NoOpGroupResolver"
```
mattrpav: Using a class for group resolution gets tricky across runtimes that have varying classloader behaviors (specifically Apache Karaf and OSGi-based runtimes.)

Since there is only one method, I think having an AbstractOAuthLoginModule with shared methods and an abstract method for groupLookup() is the way to go.

The default implementation can simply load a groups.properties file with user, group and role mappings similar to how pure local accounts work.

sergio-d-lemos:
> @sergio-d-lemos Good start.
> 
> I think having an embedded 'mock' OAuth2 server would be helpful. I think using the JDK http server would be pretty simple to spin up and then allow us to mock payloads and status codes.
> 
> Thoughts?

Yes, I think somehow mocking the HTTP server will be necessary to test. I will look into it.

PR #1484, opened 2025-08-26
[AMQ-9692] Support destination gc sweep of destinations with only wildcard consumers

mattrpav:
## Background

Garbage collecting destinations allows for a better alternative from temp-queue and temp-topic usage, since those rely on randomly generated destination names and are not subject to fine-grained permissions. In that, TempQueues and TempTopics are best to be used by applications that have dedicated brokers.

A one-time use _named_ queue can be created in a namespace for a shared broker cluster and provide for fine-grained security and self-service.

### Scope

| Destination Type | In scope |
|--------|--------|
| Queue | Yes |
| Topic (non-durable) | Yes |
| Topic (durable | No |

### Current

Current destination gc sees a wildcard-only consumer as a consumer and will not reap that destination. Adding support for gc'n wildcard-only consumers means a pattern where producers send to one-time use queues and consumers use wildcard consumers may still have the one-time queues deleted.

### Benefits

One-time use queues paired with wildcard consumers are a better pattern over using priority or message groups as there is no longer a front-of-line problem or the need to have permanent queues or additional indexing added.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
@@ -260,16 +260,17 @@ public abstract class AbstractRegion implements Region {
     }
 
     @Override
-    public void removeDestination(ConnectionContext context, ActiveMQDestination destination, long timeout)
-            throws Exception {
-
+    public void removeDestination(ConnectionContext context, ActiveMQDestination destination, long timeout) throws Exception {
         // No timeout.. then try to shut down right way, fails if there are
         // current subscribers.
         if (timeout == 0) {
             for (Iterator<Subscription> iter = subscriptions.values().iterator(); iter.hasNext();) {
                 Subscription sub = iter.next();
                 if (sub.matches(destination) ) {
-                    throw new JMSException("Destination: " + destination + " still has an active subscription: " + sub);
+                    var dest = destinations.get(destination);
+                    if(dest == null || dest.isActive()) {
+                        throw new JMSException("Destination: " + destination + " still has an active subscription: " + sub);
+                    }
                 }
             }
         }
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
@@ -260,16 +260,16 @@ public abstract class AbstractRegion implements Region {
     }
 
     @Override
-    public void removeDestination(ConnectionContext context, ActiveMQDestination destination, long timeout)
-            throws Exception {
-
+    public void removeDestination(ConnectionContext context, ActiveMQDestination destination, long timeout) throws Exception {
         // No timeout.. then try to shut down right way, fails if there are
         // current subscribers.
         if (timeout == 0) {
-            for (Iterator<Subscription> iter = subscriptions.values().iterator(); iter.hasNext();) {
-                Subscription sub = iter.next();
-                if (sub.matches(destination) ) {
-                    throw new JMSException("Destination: " + destination + " still has an active subscription: " + sub);
+            final var dest = destinations.get(destination);
+            if (dest != null) {
+                for (var sub : subscriptions.values()) {
+                    if (sub.matches(destination) && dest.isActive()) {
+                        throw new JMSException("Destination: " + destination + " still has an active subscription: " + sub);
+                    }
                 }
             }
         }
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
@@ -19,6 +19,7 @@ package org.apache.activemq.broker.region;
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
 
 import jakarta.jms.ResourceAllocationException;
 
@@ -104,7 +105,8 @@ public abstract class BaseDestination implements Destination {
     private long inactiveTimeoutBeforeGC = DEFAULT_INACTIVE_TIMEOUT_BEFORE_GC;
     private boolean gcIfInactive;
     private boolean gcWithNetworkConsumers;
-    private long lastActiveTime=0l;
+    private boolean gcWithOnlyWildcardConsumers;
+    private long lastActiveTime = 0L;
     private boolean reduceMemoryFootprint = false;
     protected final Scheduler scheduler;
     private boolean disposed = false;
@@ -310,12 +312,37 @@ public abstract class BaseDestination implements Destination {
 
     @Override
     public boolean isActive() {
-        boolean isActive = destinationStatistics.getConsumers().getCount() > 0 ||
-                           destinationStatistics.getProducers().getCount() > 0;
-        if (isActive && isGcWithNetworkConsumers() && destinationStatistics.getConsumers().getCount() > 0) {
-            isActive = hasRegularConsumers(getConsumers());
+        // if we have producers then we are active
+        if (destinationStatistics.getProducers().getCount() > 0) {
+            return true;
         }
-        return isActive;
+
+        // Check if we have active consumers that should prevent GC
+        if (destinationStatistics.getConsumers().getCount() > 0) {
+            // if we have consumers and both gcWithNetwork and gcOnlyWildcard consumers
+            // are false we can just return true, otherwise we need to check each consumer
+            return (!isGcWithNetworkConsumers() && !isGcWithOnlyWildcardConsumers()) ||
+                    hasActiveConsumers();
+        }
+
+        return false;
+    }
+
+    protected Predicate<Subscription> canGcConsumer = subscription -> {
+        // if isGcWithNetworkConsumers() is true and this is a network subscription then we can GC
+        boolean canGcNetwork = isGcWithNetworkConsumers() && subscription.getConsumerInfo().isNetworkSubscription();
+        // if isGcWithOnlyWildcardConsumers() is true and this is a wildcard then we can GC
+        return canGcNetwork || (isGcWithOnlyWildcardConsumers() && subscription.isWildcard());
+    };
+
+    protected boolean hasActiveConsumers() {
+        final List<Subscription> consumers = getConsumers();
+        for (Subscription subscription: consumers) {
+            if (!canGcConsumer.test(subscription)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     @Override
@@ -784,19 +811,37 @@ public abstract class BaseDestination implements Destination {
         return gcWithNetworkConsumers;
     }
 
+    /**
+     * Indicate if it is ok to gc destinations that have only wildcard consumers
+     * @param gcWithOnlyWildcardConsumers
+     */
+    public void setGcWithOnlyWildcardConsumers(boolean gcWithOnlyWildcardConsumers) {
+        this.gcWithOnlyWildcardConsumers = gcWithOnlyWildcardConsumers;
+    }
+
+    public boolean isGcWithOnlyWildcardConsumers() {
+        return gcWithOnlyWildcardConsumers;
+    }
+
     @Override
     public void markForGC(long timeStamp) {
-        if (isGcIfInactive() && this.lastActiveTime == 0 && isActive() == false
-                && destinationStatistics.getMessages().getCount() == 0 && getInactiveTimeoutBeforeGC() > 0l) {
+        if (isGcIfInactive()
+            && this.lastActiveTime == 0
+            && destinationStatistics.getMessages().getCount() == 0
+            && getInactiveTimeoutBeforeGC() > 0L
+            && !isActive()) {
             this.lastActiveTime = timeStamp;
         }
     }
 
     @Override
     public boolean canGC() {
-        boolean result = false;
-        final long currentLastActiveTime = this.lastActiveTime;
-        if (isGcIfInactive() && currentLastActiveTime != 0l && destinationStatistics.getMessages().getCount() == 0L ) {
+        var result = false;
+        final var currentLastActiveTime = this.lastActiveTime;
+        if (isGcIfInactive()
+            && currentLastActiveTime != 0L
+            && destinationStatistics.getMessages().getCount() == 0L
+            && !isActive()) {
             if ((System.currentTimeMillis() - currentLastActiveTime) >= getInactiveTimeoutBeforeGC()) {
                 result = true;
             }
@@ -852,17 +897,6 @@ public abstract class BaseDestination implements Destination {
 
     @Override
     public abstract List<Subscription> getConsumers();
-
-    protected boolean hasRegularConsumers(List<Subscription> consumers) {
-        boolean hasRegularConsumers = false;
-        for (Subscription subscription: consumers) {
-            if (!subscription.getConsumerInfo().isNetworkSubscription()) {
-                hasRegularConsumers = true;
-                break;
-            }
-        }
-        return hasRegularConsumers;
-    }
 
     public ConnectionContext createConnectionContext() {
         ConnectionContext answer = new ConnectionContext();
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
@@ -19,6 +19,7 @@ package org.apache.activemq.broker.region;
 import java.io.IOException;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
 
 import jakarta.jms.ResourceAllocationException;
 
@@ -105,7 +106,8 @@ public abstract class BaseDestination implements Destination {
     private long inactiveTimeoutBeforeGC = DEFAULT_INACTIVE_TIMEOUT_BEFORE_GC;
     private boolean gcIfInactive;
     private boolean gcWithNetworkConsumers;
-    private long lastActiveTime=0l;
+    private boolean gcWithOnlyWildcardConsumers;
+    private long lastActiveTime = 0L;
     private boolean reduceMemoryFootprint = false;
     protected final Scheduler scheduler;
     private boolean disposed = false;
@@ -311,12 +313,42 @@ public abstract class BaseDestination implements Destination {
 
     @Override
     public boolean isActive() {
-        boolean isActive = destinationStatistics.getConsumers().getCount() > 0 ||
-                           destinationStatistics.getProducers().getCount() > 0;
-        if (isActive && isGcWithNetworkConsumers() && destinationStatistics.getConsumers().getCount() > 0) {
-            isActive = hasRegularConsumers(getConsumers());
+        // if we have producers then we are active
+        if (destinationStatistics.getProducers().getCount() > 0) {
+            return true;
         }
-        return isActive;
+
+        // Check if we have active consumers that should prevent GC
+        if (destinationStatistics.getConsumers().getCount() > 0) {
+            // if we have consumers and both gcWithNetwork and gcOnlyWildcard consumers
+            // are false we can just return true, otherwise we need to check each consumer
+            return (!isGcWithNetworkConsumers() && !isGcWithOnlyWildcardConsumers()) ||
+                    hasActiveConsumers();
+        }
+
+        return false;
+    }
+
+    Predicate<Subscription> canGcConsumer = subscription -> {
+        // if isGcWithNetworkConsumers() is true and this is a network subscription then we can GC
+        boolean canGcNetwork = isGcWithNetworkConsumers() && subscription.getConsumerInfo().isNetworkSubscription();
+        // if isGcWithOnlyWildcardConsumers() is true and this is a non-durable wildcard then we can GC.
+        // An attached durable subscription never permits gc - its registration and pending messages
+        // live in the destination's store, which gc destroys. Note offline durable subscriptions
+        // stay attached only with keepDurableSubsActive=true (the default); brokers running with
+        // keepDurableSubsActive=false forfeit this protection while the subscriber is offline.
+        return canGcNetwork || (isGcWithOnlyWildcardConsumers() && subscription.isWildcard()
+                && !subscription.getConsumerInfo().isDurable());
+    };
+
+    protected boolean hasActiveConsumers() {
+        final List<Subscription> consumers = getConsumers();
+        for (Subscription subscription: consumers) {
+            if (!canGcConsumer.test(subscription)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
```diff
@@ -311,12 +312,24 @@ public final MessageStore getMessageStore() {
...
-        boolean isActive = destinationStatistics.getConsumers().getCount() > 0 ||
-                           destinationStatistics.getProducers().getCount() > 0;
-        if (isActive && isGcWithNetworkConsumers() && destinationStatistics.getConsumers().getCount() > 0) {
-            isActive = hasRegularConsumers(getConsumers());
+        if (destinationStatistics.getProducers().getCount() > 0) {
+            return true;
         }
-        return isActive;
+
+        var destinationActive = true;
```
cshannon: This is not very efficient because it may need to iterate through the consumers more than once. The way this is written, it may iterate through the consumers twice if both isGcWithNetworkConsumers() and isGcWithOnlyWildcardConsumers() is true. 

What would make more sense is to only iterate through the consumers one time and use the predicate API to specify what you are looking for. You could build a predicate that look for either a network consumer, or not a wildcard consumer or some combination quite easily using https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/function/Predicate.html . This would also make it easy to expand the criteria in the future for other conditions. 

If all the flags are off then we wouldn't need to iterate through the consumers at all.

cshannon (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
mattrpav: Resolved with latest change

mattrpav:
[x] GC of destinations with wildcard only works
[x] Tested lifecycle of wildcard consumer working when matching queue is added, gc'd and re-added. Messages are consumed properly.

activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
```diff
@@ -260,16 +260,21 @@ protected List<Subscription> addSubscriptionsForDestination(ConnectionContext co
     }
 
     @Override
-    public void removeDestination(ConnectionContext context, ActiveMQDestination destination, long timeout)
-            throws Exception {
-
+    public void removeDestination(ConnectionContext context, ActiveMQDestination destination, long timeout) throws Exception {
```
jbonofre, line 263: nit: it would be good to avoid reformating, it simplified the review.

jbonofre (dismissed):
LGTM, thanks !

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
```diff
@@ -311,12 +313,27 @@ public final MessageStore getMessageStore() {
...
+        if (destinationStatistics.getConsumers().getCount() > 0) {
+            Predicate<Subscription> isActiveConsumer = subscription -> {
```
cshannon: Previously we didn't have to filter if the isGcWithNetworkConsumers() flag was false. We could simply just look for the count of consumers and if greater than 0 return. There was no iterating.

Now we always have to iterate, and on top of that java streams is slow and this is somewhat on a hot path because it's going to be called frequently depending on the setup. I really like streams but not if there's potential performance issues if someone is tuning frequent checks for active consumers and has a lot of consumers.

I would prefer if this:
1. Preserved avoiding iterating if the two flags are both false
2. Just used a for each loop and avoid streams for performance.

cshannon (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
cshannon: Also, if a predicate is used, it should be created once and re-used and not every single method call.

cshannon:
```java
@Override
public boolean isActive() {
    // if we have producers then we are active
    if (destinationStatistics.getProducers().getCount() > 0) {
        return true;
    }

    // Check if we have active consumers that should prevent GC
    if (destinationStatistics.getConsumers().getCount() > 0) {
        // if we have consumers and both gcWithNetwork and gcOnlyWildcard consumers
        // are false we can just return true, otherwise we need to check each consumer
        return (!isGcWithNetworkConsumers() && !isGcWithOnlyWildcardConsumers()) ||
                hasActiveConsumers();
    }
    
    return false;
}

protected Predicate<Subscription> canGcConsumer = subscription -> {
    // if isGcWithNetworkConsumers() is true and this is a network subscription then we can GC
    boolean canGcNetwork = isGcWithNetworkConsumers() && subscription.getConsumerInfo().isNetworkSubscription();
    // if isGcWithOnlyWildcardConsumers() is true and this is a wildcard then we can GC
    return canGcNetwork || (isGcWithOnlyWildcardConsumers() && subscription.isWildcard());
};

protected boolean hasActiveConsumers() {
    final List<Subscription> consumers = getConsumers();
    for (Subscription subscription: consumers) {
        if (!canGcConsumer.test(subscription)) {
            return true;
        }
    }
    return false;
}
```

cshannon: That is a suggestion on how we could write it, I played around with the logic a bit and i found it less confusing to write a predicate that checked if the consumer could be GC'd vs is active . The predicate returns true if either can gc network flag is true and it's a network consumer, or if the gc with only wildcard flag is true and the subscription is a wildcard. 

It also optimizes by doing an initial check to see if either flag is true and if both are false we can skip iteration entirely. I used a for each loop vs streams just for the slight performance boost as isActive() will probably be called a lot on each destination (although streams would be a lot cleaner)

mattrpav: Changes applied

cshannon:
@mattrpav - The biggest thing now is I think the test coverage is lacking.

1. Some of the new methods could be unit tested .  The new hasActiveConsumers() and canGcConsumer predicate can probably be tested in isolation as they are protected so maybe with mocks or something. 
2. I think we need to double check there is sufficient test cover for the different combinations of the gcWithOnlyWildcardConsumers and gcWithOnlyWildcardConsumers flags being set. You added a test if gcWithOnlyWildcardConsumers is true but we should also add a quick test where both flags are true (easy enough to change in the policy and can probably just reuse the same test). We should also verify there's a test somewhere that still works for gc network consumers flag (i would hope so). 
3. Verify that there's good GC test coverage with both flags false as the default, i would expect there should already be some GC tests but I haven't looked in a while.

I think better test coverage is important because it would be bad to start deleting destinations that should not be deleted as that leads to permanent data loss.

mattrpav:
@cshannon new DestinationIsActiveTest adds parameterized tests for logical combinations of:

PolicyEntry gc flag combinations vs status of:

1. Normal consumer
2. Network consumer
3. Wildcard consumer

There is also a fail-safe check to ensure parameter config is correct by verifying that anytime the normal app consumer is active, the queue is flagging as active.

Intellij confirmed 100% branch coverage on those methods (aside from the producer count test)

activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
```diff
@@ -260,16 +260,21 @@ protected List<Subscription> addSubscriptionsForDestination(ConnectionContext co
...
         // No timeout.. then try to shut down right way, fails if there are
         // current subscribers.
         if (timeout == 0) {
             for (Iterator<Subscription> iter = subscriptions.values().iterator(); iter.hasNext();) {
                 Subscription sub = iter.next();
                 if (sub.matches(destination) ) {
-                    throw new JMSException("Destination: " + destination + " still has an active subscription: " + sub);
+                    if(sub.isWildcard()) {
+                        var dest = destinations.get(destination);
+                        if(dest != null && dest.isGcWithOnlyWildcardConsumers()) {
```
cshannon: There still seems to be a missing test case because this logic is broken, Intellij flagged it. This continue does nothing as it's in a loop and this will not throw the exception if isGcWithOnlyWildcardConsumers() is false.

This should likely just be rewritten as something like and include a test that would catch a similar mistake in  the future.

```java
if (sub.matches(destination) ) {
    var dest = destinations.get(destination);
    if(!sub.isWildcard() || dest == null || !dest.isGcWithOnlyWildcardConsumers()) {
        throw new JMSException("Destination: " + destination + " still has an active subscription: " + sub);
    } 
}
```

cshannon (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
mattrpav: Would this also need to look for the same for network-only subscribers for consistency? Or better yet perhaps re-use the isActive() or canGC() methods?

Note: This method is invoked by the RegionBroker gc check _after_ confirming the expiration rules apply for the given destination.

In the case of GC, this is a repeat of the check. 

In the case of an admin operation (ie via CLI or JMX) this is the only check to prevent deleting a destination w/ active subscribers.

cshannon: isActive() and canGC() are a bit different because they also check for active producers. I'm not entirely sure why the method to remove was written not to care about producers, maybe an oversight, but changing that now would be a behavior change that could cause some side effects that are unexpected. Re-using logic is good but maybe it just needs to be subdivided to only check for active consumers (that method already exists and isActive() calls it now).

As you pointed out, the RegionBroker calls this to delete the destination but already after the gc check so we know there are no active consumer/producers. So the real question is whether or not we care about blocking a manual deletion through JMX if there are active producers. It seems odd to not check that but changing that now as I said is a behavior change that could break someone's use case. (I suppose i could see someone wanting to block deletes with active consumers but not producers). 

Your point about it not checking the gcWithNetworkConsumers flag is valid and probably a bug. In fact it proves the point that we should be sharing the logic because the GC automated check will check it but not the manual deletion.

cshannon: There's a long list of things that are like this where we may want to revisit for AMQ 7 where we want to change default behavior or settings but it's technically a breaking change so a major version is more appropriate. I guess whether or not we change this to check active producers depends on if we consider it a bug or not.

If we change it I would think by default this would share the same logic as the isActive and GC check and that should block removal by default even by an admin through JMX/console etc to prevent mistakes. But, if we do that we could add a force flag that would allow an administrator to force delete, but all of this is kind of out scope of the work for this.

jbonofre:
I guess we would like this fix included in 5.19.2 right ?

@cshannon do we make a new pass on this PR or is it good to go ?

cshannon:
> I guess we would like this fix included in 5.19.2 right ?
> 
> @cshannon do we make a new pass on this PR or is it good to go ?

This still needs a little more work so maybe we can hold off if we are ready to do releases now

jbonofre:
@mattrpav if you want me to include this PR in the 5.19.2 release, can you please update ?
Else, I will move forward with the release and we can include this PR in the next one.

mattrpav:
Updated tests include non-durable topics. Durable topics are not included by design. Also, added a 'last minute' check to make sure destinations don't get busy from the time of being marked and removed.

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
```diff
@@ -311,12 +313,46 @@ public final MessageStore getMessageStore() {
...
+        return false;
+    }
+
+    protected Predicate<Subscription> canGcConsumer = subscription -> {
+        // if isGcWithNetworkConsumers() is true and this is a network subscription then we can GC
+        boolean canGcNetwork = isGcWithNetworkConsumers() && subscription.getConsumerInfo().isNetworkSubscription();
+        // if isGcWithOnlyWildcardConsumers() is true and this is a non-durable wildcard then we can GC.
+        // Durable subscriptions never permit gc - their registration and pending messages live in the
+        // destination's store, which gc destroys, so collecting would break the durability guarantee.
+        return canGcNetwork || (isGcWithOnlyWildcardConsumers() && subscription.isWildcard()
```
cshannon, line 340: Why did you make this specific change with durables? I don't think this change is necessary and it also isn't described anywhere and is outside of the scope of the original PR.

Destinations get marked for GC only if the pending size is 0 and there are no producers/consumers. If keepDurableSubsActive = false and there are no pending messages anyways, why would that block GC? If you set keepDurableSubsActive = true then it would count as a consumer and stay active.

I could see a use case for the existence of any durable blocking the removal of a destination, but i don't think that check would fall under the "is active" check. While a bit annoying, this feels like yet another policy setting that is something like "preventDestinationGCWithDurables" or something like that and should likely be an entirely separate PR.

cshannon (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
cshannon: Thinking about this more I think we just need to do some more testing/investigation for durables and the wildcard case because as you noted durables get created on real destinations when it's a wildcard. (including if the wildcard destination itself exists). maybe some improved tests would help clarify.

I think we just want to make sure of 2 things:

1. If the new flag is false (the default) the existing behavior with cleanup shouldn't change. That includes whether the sub is a wildcard or not and whether keepDurableSubsActive is true or false, etc.
2. if the new flag is set to cleanup when wildcard, we just want to make sure it makes sense for durables too. I'm not sure we need to do anything special for durables because if keepDurableSubsActive = true of course will just keep the destinations active as they count as a consumer. For the case where keepDurableSubsActive = false, a durable gets registered on the specific topic (even if a wildcard initially). I'm not sure we want to block GC in that case just f it's offline and no messages it seems like it might be ok to remove...but only if the current behavior today also would remove it in that case (needs testing)

mattrpav: 1. I have an updated unit test that validates this.
2. I think supported durable subscriptions would be expensive. We'd have to check each durable subscription's size on each topic to ensure it is empty every sweep. 

Options:

a. Keep this change non-durable-topics and queues-only (preserves pre-change behavior)
b. Bite off durable topic subscription support.

cshannon: @mattrpav - I think we just remove all the special durable handling entirely as it's really out of the scope of this issue and PR.

I don't think we need to actually do anything special for durables because there's already a mechanism for it with the `keepDurableSubsActive` flag that can be used to prevent GC'ing if desired. The current check for active destinations doesn't care specifically about durables, just consumers. If that flag is true then a durable is considered an online consumer and we don't GC, if false it's treated just like any other offline consumer and we can GC if no pending messages exist.

So, I would remove all the durable stuff and if you have a use case to change the behavior around how durable subscriptions work in conjunction with the GC active check, I would make that an entirely separate issue/PR and also it would need to take into account all use cases, not just wild cards. (but like i said it seems entirely uneeded)

activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
```diff
@@ -260,16 +260,21 @@ protected List<Subscription> addSubscriptionsForDestination(ConnectionContext co
...
         // current subscribers.
         if (timeout == 0) {
+            final Destination dest = destinations.get(destination);
+            final boolean destActive = dest != null && dest.isActive();
             for (Iterator<Subscription> iter = subscriptions.values().iterator(); iter.hasNext();) {
                 Subscription sub = iter.next();
-                if (sub.matches(destination) ) {
-                    throw new JMSException("Destination: " + destination + " still has an active subscription: " + sub);
+                if (sub.matches(destination)) {
+                    if (dest == null) {
```
cshannon: This null check doesn't make sense, sub.matches(null) will throw an NPE and I would get rid of it entirely. There could be race conditions with destination removal where it doesn't exist in the map anymore because you are not under lock. If you look further down the code will just only process if not null. 

I would simply this entire block of code and do the following:

```java
if (timeout == 0) {
    final Destination dest = destinations.get(destination);
    if (dest != null) {
        for (Subscription sub : subscriptions.values()) {
            if (sub.matches(destination) && dest.isActive()) {
                throw new JMSException("Destination: " + destination + " still has an active subscription: " + sub);
            }
        }
    }
}
```

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
```diff
@@ -311,12 +313,42 @@ public final MessageStore getMessageStore() {
...
+            // if we have consumers and both gcWithNetwork and gcOnlyWildcard consumers
+            // are false we can just return true, otherwise we need to check each consumer
+            return (!isGcWithNetworkConsumers() && !isGcWithOnlyWildcardConsumers()) ||
+                    hasActiveConsumers();
+        }
+
+        return false;
+    }
+
+    protected Predicate<Subscription> canGcConsumer = subscription -> {
```
cshannon: this should likely be private and final as there is no need to override it for now.

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
```diff
@@ -311,12 +313,42 @@ public final MessageStore getMessageStore() {
...
+        // if isGcWithNetworkConsumers() is true and this is a network subscription then we can GC
+        boolean canGcNetwork = isGcWithNetworkConsumers() && subscription.getConsumerInfo().isNetworkSubscription();
+        // if isGcWithOnlyWildcardConsumers() is true and this is a non-durable wildcard then we can GC.
+        // An attached durable subscription never permits gc - its registration and pending messages
+        // live in the destination's store, which gc destroys. Note offline durable subscriptions
+        // stay attached only with keepDurableSubsActive=true (the default); brokers running with
+        // keepDurableSubsActive=false forfeit this protection while the subscriber is offline.
+        return canGcNetwork || (isGcWithOnlyWildcardConsumers() && subscription.isWildcard()
+                && !subscription.getConsumerInfo().isDurable());
```
cshannon, line 341: I thought you were getting rid of this durable check here as it isn't necessary if relying on `keepDurableSubsActive` flag?

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
```diff
@@ -796,19 +828,37 @@ public boolean isGcWithNetworkConsumers() {
...
     public boolean canGC() {
-        boolean result = false;
-        final long currentLastActiveTime = this.lastActiveTime;
-        if (isGcIfInactive() && currentLastActiveTime != 0l && destinationStatistics.getMessages().getCount() == 0L ) {
+        var result = false;
+        final var currentLastActiveTime = this.lastActiveTime;
+        if (isGcIfInactive()
+            && currentLastActiveTime != 0L
+            && destinationStatistics.getMessages().getCount() == 0L
+            && !isActive()) {
```
cshannon: I don't think this change is needed, you don't need to change canGC() at all or add the isActive() check.

The way this is supposed to work is `markForGC() `does all that work and then sets `lastActiveTime`, so this method only needs to check that. I would not change canGC() method at all

activemq-broker/src/main/java/org/apache/activemq/broker/region/RegionBroker.java
```diff
@@ -936,6 +936,13 @@ protected void purgeInactiveDestinations() {
                 context.setBroker(this);
 
                 for (Destination dest : list) {
+                    // Re-verify eligibility right before removal - a message can arrive
+                    // (e.g. from an anonymous producer, which does not hold the purge
+                    // lock) between the mark loop above and this removal loop.
+                    if (!dest.canGC()) {
```
cshannon: This is also unnecessary and should be removed, it breaks the `maxPurgedDests` check as we may now not purge the max that we want. 

When this loop executes it's executing immediately after the loop runs to check and mark for GC and also count the number of destinations for maxPurgeDests. No matter what there is a race condition here, this isn't going to fix that and it doesn't make sense to check like 2 milliseconds later the same condition.

cshannon (changes requested):
I found several things to change, comments in line

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
cshannon: Actually the other option is to make it package scope and then you could write unit tests directly for the predicate, which are missing. It would probably be a good idea to have test that just runs checks on the predicate to verify it returns correctly based on all the different conditions. The existing tests could even call it directly as another check, etc

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
mattrpav: I was able to remove the counter comparison check; however, if this .isDurable() check is removed a topic with only one wildcard durable subscription would be deleted (b/c of         subscription.isWildcard() evaluating true) and lead to unintended message loss.

activemq-broker/src/main/java/org/apache/activemq/broker/region/RegionBroker.java
mattrpav: Agreed. Removed in second commit with review feedback changes.

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
cshannon: @mattrpav - Part of the GC check is that the message count must be 0, the destination won't be GC'd if there are pending messages.Therefore this check should be removed so it is consistent with the other isActive() check as the other checks for non wildcard do not check for durables. 

As I stated, nothing about durables should be part of this PR, the behavior should be the same as it is today.

Are there race conditions around the GC sweep? Yep, but nothing here fixes that. If you think durables should be part of the check that needs to be a separate discussion.

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
mattrpav: This was added originally to gate the network and wildcard only config flags behind the isActive().  Since markForGc calls isActive, this is now redundant. Removing.  Good catch

activemq-broker/src/main/java/org/apache/activemq/broker/region/AbstractRegion.java
mattrpav: Resolved in commit with review feedback changes

activemq-broker/src/main/java/org/apache/activemq/broker/region/BaseDestination.java
cshannon: @mattrpav - what about this, instead of checking if any subs are durables, we could do the following and check sub pending size for all subs, which is what we really care about (are there pending messages):

```java
protected Predicate<Subscription> canGcConsumer = subscription -> {
    // Check 1: if isGcWithNetworkConsumers() is true and this is a network subscription then we can GC
    boolean canGcNetwork = isGcWithNetworkConsumers() && subscription.getConsumerInfo().isNetworkSubscription();
    // Check 2: if isGcWithOnlyWildcardConsumers() is true and this is a wildcard then we can GC.
    //
    // Check 3: if the pending size is > 0, then we can't' GC because there are pending messages
    // This check is necessary for subs on a Topic in particular because the messages count metric
    // used inside of canGC() won't track the messages after they are passed to the subs.
    //
    // Note: getPendingQueueSize() will always return 0 for on offline durable sub unless
    // the keepDurableSubsActive flag is set to true. This means that offline durables that 
    // have keepDurableSubsActive=false will not block GC even if there are pending messages.
    return canGcNetwork || (isGcWithOnlyWildcardConsumers() && subscription.isWildcard())
            || subscription.getPendingQueueSize() == 0;
};
```

The obvious glaring thing is how to handle the case of offline durables where keepDurableSubsActive=false. The above keeps the current behavior and makes them eligible for GC. I am hesitant to make yet another flag, but I suppose we could as a separate task to allow the existence of offline durables, even if  keepDurableSubsActive=false, to prevent GC.

cshannon: @mattrpav -

I did some testing and I realized my assumption was wrong, the destination consumer count metric **will track any** durable subscription that exists, regardless of if it is active or if the keepDurableSubsActive is true or false. The metric for consumers is **only** decremented on subscription [deletion](https://github.com/apache/activemq/blob/c1bb9d6613b7713449e588056cf6bfbbfa92988c/activemq-broker/src/main/java/org/apache/activemq/broker/region/Topic.java#L228).

So going back and forth I think what you have is probably best. My original pushback was not wanting to treat wildcards as special but I agree there isn't any other good mechanism to prevent data loss for durables. The current code only considers network subs, so it just won't delete a destination if there are durables otherwise. Now, we could delete if there are durables but the consumer is a wildcard so that is probably bad.

The  subscription.getPendingQueueSize() == 0 isn't good enough because offline durables that are inactive will return 0 if the keep active flag is false, leading to message loss so that isn't helpful.

So in the end, after like 2 days of back and forth, I think what you have makes sense after all so i'll approve it as it's the safest option with the new wildcard check.

cshannon (dismissed):

cshannon (approved):

PR #1487, opened 2025-09-12
[AMQ-9768] Fix JMX value escaping in web console.

thezbyg:


jbonofre:
If I understand the need here, it looks like an improvement more than a bug fix as it changes some "interfaces" (introducing name field).

thezbyg:
The main issue is missing JMX value escaping. It could be fixed without any broker interface changes by adding JMXSupport.encodeObjectNamePart() calls in BrokerFacadeSupport.java file, but there is the following code that attempts to transform from JMX value to connection name that is displayed to user:
https://github.com/apache/activemq/blob/fb781880e65a7bd4c6159567fc41edeb4d01cd63/activemq-web/src/main/java/org/apache/activemq/web/BrokerFacadeSupport.java#L178
JMX value escaping performed by JMXSupport.encodeObjectNamePart() is not reversible and it is impossible to get the original value from escaped value. There is no way to get original connection name as ConnectionViewMBean does not have a name property. This is the reason for interface changes.

I can remove all other changes and leave only BrokerFacadeSupport.java changes if that is preferred.

jbonofre:
I would suggest to split into two PRs:
1. One just fixing the formatting that we can also apply on 6.1.x
2. One changing the interface (adding name) that will be apply just on `main`

thezbyg:
> One just fixing the formatting that we can also apply on 6.1.x

Done.

jbonofre:
@thezbyg thanks ! Much appreciated !

PR #1493, opened 2025-09-14
 [AMQ-9769] Add name property to connector interface and MBean.

thezbyg:


activemq-web-console/src/main/webapp/WEB-INF/tags/jms/forEachConnection.tag
```diff
@@ -23,9 +23,8 @@
 <%
 	Iterator it = broker.getConnections(connectorName).iterator();
 	while (it.hasNext()) {
-		String conName = (String) it.next();
-		ConnectionViewMBean con = broker.getConnection(conName);
-		request.setAttribute(connectionName, conName);
+		ConnectionViewMBean con = (ConnectionViewMBean) it.next();
+		request.setAttribute(connectionName, con.getClientId());
```
jbonofre, line 27: I'm a bit confused here. Is it `con.getCliientId()` or connection name MBean constructed.
Can you please just add a comment here to explain the use of `clientId` (instead of constructing with the name, and more) ?

mattrpav: Heads up-- iirc, there are inconsistencies in connection name/id/clientId. 

Some protocols only allow a single clientId-per-broker (mqtt) and some can in certain scenarios (jms). We need to be super sure about which is used to avoid a scenario where the web ui can have two entries for the same id in a list view.

jbonofre: That's my point here. I'm a bit concerned by this connection name here.

thezbyg: @jbonofre 
Connection MBeans with `connectionViewType=clientId` type use connection `clientId` value for `connectionName` parameter, so `connectionName` MBean parameter value and `clientId` connection property value should be equivalent.

This is the code responsible for `connectionName` value in connection MBean:
https://github.com/apache/activemq/blob/51553ddc5f4fac3cd8d89dc5cf485d5d527d60d1/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagedTransportConnection.java#L84
https://github.com/apache/activemq/blob/51553ddc5f4fac3cd8d89dc5cf485d5d527d60d1/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagedTransportConnection.java#L117
https://github.com/apache/activemq/blob/51553ddc5f4fac3cd8d89dc5cf485d5d527d60d1/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerMBeanSupport.java#L181

@mattrpav 
If JMS allows multiple connections with same `clientId` value, then it should have already caused problems by attempting to create multiple MBeans with the same name. If this was not an issue before my changes, then I will have to investigate more, because I thought that `clientId` must always be unique and can not be shared between connections.

jbonofre: @thezbyg thanks for your feedback. It looks good. Let me do a new pass, but I think it's good to merge. @mattrpav thoughts ?

mattrpav:
Requested change:

1. Please add getConnectionId() to ConnectionViewMBean

The _connectionId_ is the unique value within the broker for the connection. 

Side note: We should be registering JMX using that unique value instead of the _clientId_ since _clientId_ can be reused across connections and quickly swapped out if link stealing is enabled, but we won't dial up that change now.

thezbyg:
Done, but now ConnectionView returns the same value for both `getClientId` and `getConnectionId`:
```
public String getClientId() {
    return connection.getConnectionId();
}
public String getConnectionId() {
    return connection.getConnectionId();
}
```
And `connection.getConnectionId()` executes the following code in TransportConnection class:
```
    public String getConnectionId() {
        List<TransportConnectionState> connectionStates = listConnectionStates();
        for (TransportConnectionState cs : connectionStates) {
            if (cs.getInfo().getClientId() != null) {
                return cs.getInfo().getClientId();
            }
            return cs.getInfo().getConnectionId().toString();
        }
        return null;
    }
```
So it may return either `clientId` or `connectionId`.

I propose to remove connection handling changes from this pull request and keep only changes to connector MBean and connector output in web UI.

jbonofre:
@thezbyg can you please rebase and fix conflict ?

thezbyg:
Rebased.

Connection `clientId` issue could be fixed by saving current `clientId` value when registering connection MBean. Like this: https://github.com/thezbyg/activemq/commit/53d7e53366b88ceb00d6f5a08aad5bd83c618891
This would ensure that `clientId` returned by `connection.getClientId()` could not change after MBean is registered and would always match `connectionName` property value in connection MBean object name.

jeanouii:
@thezbyg I'm sorry to bug you again. Would you mind rebase this PR?
It looks good and we'd like to merge it, but we started a bit stabilization task and we touched many files. So this PR is not mergeable as is.

PR #1495, opened 2025-09-17
NO-JIRA: Enable security features when building XML Schema in activemq-runtime-config.

sergio-d-lemos:
Enables XML security features in the runtime-config plugin similar to how we do in other modules.

I tested by running an ActiveMQ broker with the `runtimeConfigurationPlugin` enabled, the plugin started and picked up changes in activemq.xml.


mattrpav:
This is a good enhancement. Please make a JIRA, so it’ll be in release notes

sergio-d-lemos:
Update PR in https://github.com/apache/activemq/pull/1498

PR #1496, opened 2025-09-17
Weekly s390x Job

SanyogDeshmukh:
Added weekly s390x job which triggers on every Sunday.

PR #1497, opened 2025-09-17
Use of GitHub Actions, and enable dependabot

jbonofre:
Keep a backup of Jenkinsfile for now


jbonofre:
I also integrated GitHub Action in the PRs.

cshannon (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+jobs:
+  build:
+
+    permissions:
+      contents: read
+
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
+        java-version: [ 17, 21, 24, 25 ]
```
mattrpav: JDK 25 can't be used until we make a JDK 21 min branch (removed JAAS APIs, removed ThreadLocal)

Jenkinsfile.bak
mattrpav, line 1: Do we have to disable Jenkins completely?

It would be good to keep Jenkins as a schedule and manual-only to do sonar and s390x

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+
+name: Deploy
```
mattrpav, line 20: What is the use case for doing a deploy?

.github/workflows/build.yml
jbonofre: Good point, I was a bit too "aggressive" 😄 

Let me remove 25 here. Thanks !

.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+# under the License.
+#
+
+name: Build
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    branches: [ "main" ]
```
mattrpav: If we get more builds for free should we add the active backport branches here?

activemq-6.1.x
activemq-5.19.x

mattrpav (changes requested):

Jenkinsfile.bak
jbonofre: I guess we can do Sonar via GH Actions.

For s390x, yeah, GitHub Action doesn't have hosted runner for s390x.

It's a good idea to use Jenkins on schedule (once a day) but GH Actions for PRs verify.

Let me do that.

.github/workflows/deploy.yml
jbonofre: The purpose here is to deploy SNAPSHOTs on repository.apache.org.

If it's useless, happy to remove the deploy workflow.

.github/workflows/build.yml
jbonofre: Oh sure, I can definitely do that.

jbonofre: Including for PRs, we should create PRs for cherry-pick, just to be sure CI is happy.

.github/workflows/deploy.yml
mattrpav: Deploying to Apache SNAPSHOTs makes sense I'm good with keeping it. Thanks.

.github/workflows/build.yml
```diff
@@ -0,0 +1,52 @@
...
+
+jobs:
+  build:
+
+    permissions:
+      contents: read
+
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, windows-latest ]
```
mattrpav: I think we should specify target LTS platforms where possible, so we can include that info in the matrix of supported things -- like JDK, EE version, etc.

ubuntu-24.04
ubuntu-24.04-arm
ubuntu-22.04
ubuntu-22.04-arm
macos-26
macos-15
windows-2025
windows-2022
windows-11-arm

jbonofre: The available images are listed here: https://github.com/actions/runner-images?tab=readme-ov-file#available-images

I will extend the list, but:
1. We don't have the host runner for all architecture you said
2. I'm not sure it makes sense to have all this

jbonofre: I updated the list of runners to use.

mattrpav:
I don’t think we should remove the unit test retry.  @cshannon pointed that out as a good solution to get past some flaky tests and tests that fail when concurrent builds are running

jbonofre:
@mattrpav ok I readd it. I will also use a different strategy for tests to use different jdk.

Jenkinsfile.bak
jbonofre: I re-added Jenkinsfile with s390x build only and Sonar phase. I added a trigger to run daily. I think we should not run Jenkins by PR (to speed up 😄 ).

jbonofre:
FYI, I enabled again Jenkins on PR. I will disable it only when GitHub Action will be happy.

jbonofre:
Resuming this one 😄

jbonofre:
Regarding the size of the matrix (2 ubuntu, 4 macos, 2 windows, with 3 different JDK versions), I propose:

1. In automatic mode (for each PR), I propose to do only Ubuntu 24.04 with just JDK 17
2. In nightly and/or on demand, I propose to do all the matrix (all OS, all JDK)

The purpose is to speed up build/test on PRs, and keep the "full pass" on nightly (on demand). Imho, 80% of the matrix is useless (macos, all windows version, ...).

Thoughts ? @cshannon @mattrpav @jeanouii

jeanouii:
Personnally, I think it's a great PR and it demonstrates what we can do and the benefits.
That being said, I can see 2 cases (similar to you)

- Dev/pr: I'd only go for Ubuntu (1 version only) and JDK 17 (or 24). That's it.
If we need to work with JDK 24 or 25, we can manually trigger it.

- nightly: we can go wider and test the 3 JDK, and Ubuntu / Windows
I'm fine adding MacOS, but 4 versions (for 3 JDK) sounds like a big cardinality for a very small coverage of the real life. I think probably less than 1% of production deployments are on MacOS. 
We can keep 1 or 2 versions of Windows, if we want, but multiplied by 3 JVM, that's already at least 5*3 = 15 builds+tests

So overall, fully inline with you.

activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/JournalCorruptionEofIndexRecoveryTest.java
```diff
@@ -63,12 +63,13 @@
...
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.slf4j.event.LoggingEvent;
 
-
+@Ignore("Temporary to test GitHub Action")
```
jeanouii: This one should now be stable - whenever the other PRs are merged, we can rebase and put it back

activemq-kahadb-store/pom.xml
```diff
@@ -30,6 +30,10 @@
   <name>ActiveMQ :: KahaDB Store</name>
   <description>The ActiveMQ KahaDB Store Implementation</description>
 
+  <properties>
+    <surefire.argLine>-Xmx512M</surefire.argLine>
```
jeanouii: This should not be required when we merge all other PRs. The OOME should not trigger anymore.

pom.xml
```diff
@@ -968,13 +968,12 @@
...
             <forkCount>1</forkCount>
             <reuseForks>true</reuseForks>
-            <argLine>-enableassertions</argLine>
             <failIfNoTests>false</failIfNoTests>
             <systemPropertyVariables>
                 <java.awt.headless>true</java.awt.headless>
                 <org.apache.activemq.kahaDB.files.skipMetadataUpdate>true</org.apache.activemq.kahaDB.files.skipMetadataUpdate>
             </systemPropertyVariables>
-            <argLine>-Xmx512m</argLine>
+            <argLine>-enableassertions -Xmx1G</argLine>
```
jeanouii: Same here for the memory increase.

activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/JournalCorruptionEofIndexRecoveryTest.java
jbonofre: I removed the `@Ignore`.

pom.xml
jbonofre: Ack, remove the `Xmx` configuration.

activemq-kahadb-store/pom.xml
jbonofre: Ack. I removed the `Xmx`.

jeanouii:
So close 3h28 as far as the HTTP module.

```
[INFO] Running org.apache.activemq.transport.amqp.interop.AmqpReceiverTest
Error:  Tests run: 25, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 3.187 s <<< FAILURE! -- in org.apache.activemq.transport.amqp.interop.AmqpReceiverTest
Error:  org.apache.activemq.transport.amqp.interop.AmqpReceiverTest.testTwoQueueReceiversOnSameConnectionReadMessagesAcceptOnEach -- Time elapsed: 0.074 s <<< FAILURE!
java.lang.AssertionError: expected:<4> but was:<3>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.junit.Assert.assertEquals(Assert.java:633)
	at org.apache.activemq.transport.amqp.interop.AmqpReceiverTest.testTwoQueueReceiversOnSameConnectionReadMessagesAcceptOnEach(AmqpReceiverTest.java:490)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
	at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:299)
	at org.junit.internal.runners.statements.FailOnTimeout$CallableStatement.call(FailOnTimeout.java:293)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.lang.Thread.run(Thread.java:840)
```

jbonofre:
GitHub Action is all green.

Jenkins failed but it's unrelated. Let me check the last issue on Jenkins to see if I can do a quick fix and also double check we build only for s390x.

jeanouii:
I've "replay" the last build because a STOMP test failed looks like.

jbonofre:
@jeanouii I restarted as it failed again on STOMP.

jbonofre:
As Jenkins often fails and super long, I propose to keep only GitHub Action as PR check, and keep Jenkins on demand and cron for s390x check.

jbonofre:
@mattrpav GitHub Action CI is now ok (and way faster than Jenkins). I propose:
1. To keep Jenkins for s390x only, on-demand and build on main after a commit, but disable on PRs (in favor of GitHub Action)
2. Merge this PR
OK for you ?

Jenkinsfile
```diff
@@ -42,7 +46,7 @@ pipeline {
     }
 
     parameters {
-        choice(name: 'nodeLabel', choices: ['ubuntu', 's390x', 'arm', 'Windows']) 
+        choice(name: 'nodeLabel', choices: [ 's390x'])
```
mattrpav: Can we keep the other architectures for manual runs (specifically for sonar on ubuntu)?

This contribution auto-detects and select s390x based on cron trigger:
ref: https://github.com/apache/activemq/pull/1496

jbonofre: Yes, sure. If it's executed on demand (not on PR), I'm totally fine with that. The purpose is really to have stable and fast checks on PR.

PR #1499, opened 2025-09-24
[AMQ-9773] Fix for only one message being recovered from backup

mattrpav:


mattrpav:
JIRA notes updated, comment added with this text as well

```
The order index sequence key value increases for every message since the beginning of the creation of the index, so the first message available won't be at sequence key value:0 in the index when there were older messages acknowledged.

The MessageOrderCursor _position_ value is relative to the index, so there is a disconnect between queue _position_ and index sequence value over time.

The MessageRecoveryContext determines the recovery start position based off the provided offset, or the position of the requested startMessageId. If a startMessageId is specified, but not found in the index, then the value of 0 is used as a fallback.

The MessageRecoveryContext determines the recovery end position based off of the provided endMessageId (if provided), or the maximum recovered message count, or if the MessageRecoveryListener signals that no more messages should be recovered (ie memory is full).
```

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java
```diff
@@ -796,7 +803,11 @@ public void execute(Transaction tx) throws Exception {
                                 break;
                             }
                         }
-                        sd.orderIndex.stoppedIterating();
+
+                        // The sd.orderIndex uses the destination's cursor
+                        if(!messageRecoveryContext.isUseDedicatedCursor()) {
```
jbonofre, line 808: What happen when it's dedicated cursor ? Don't we need to flag the context and stop iterating ?

jbonofre (approved):
LGTM, I just wonder if we should not stop iterating when using a dedicated cursor.

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java
mattrpav: @jbonofre the sd.orderIndex.stopIterating() operates on the destination's cursor by default. If there is a dedicated cursor, there is no need to stop iterating, b/c it is a one-time use cursor.

cshannon: @mattrpav is right, the other cursor that is part of the StoredDestination is stateful so that we can page off messages as they come in. The dedicated is meant to be stateless so there's no need to call that

jbonofre: Understood, just wanted to be sure we are not missing something here. Thanks !

PR #1500, opened 2025-09-25
AMQ-9775: Add the maxBrowsePageSize attribute to the JMX destination MBean

jbonofre:
This PR adds the `maxBrowsePageSize` attribute to the JMX destination MBean.

As described in the Jira (AMQ-9775), the `maxBrowsePageSize` property can be set in the destination policies, in the broker.
I have several use cases where client applications are interested to get this metric (via JMX or REST with Jolokia):
* know the browse paging limit to "optimize" paging/browsing on client side (faster)
* get detail about the limit to "recommend" update on the paging limit

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/DestinationView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/DestinationView.java
@@ -31,7 +31,6 @@ import jakarta.jms.InvalidSelectorException;
 import jakarta.jms.MessageProducer;
 import jakarta.jms.Session;
 import java.util.function.Function;
-import java.util.function.Supplier;
 import javax.management.MalformedObjectNameException;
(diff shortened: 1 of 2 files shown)


jbonofre:
I can add a test about this new JMX attribute.

mattrpav:
@jbonofre please add a test for queue AND topic subscription recovery policy to catch regression

jeanouii:
The ticket refers to being able to provide a configuration option (maxBrowsePageSize), but the PR also includes the ability to update it. If this is intended, then we probably need to update the ticket and check the impact of updating the property at runtime. If it's not intended, we can I think remove the setter and not necessarly have tons of tests, it's read-only.

jbonofre:
My bad: the intent is really to just read (not update). I'm updating the PR accordingly.

jbonofre:
@jeanouii updated to just have read attribute.

PR #1503, opened 2025-10-02
[AMQ-9777] Migrate to using a multi-release jar for activemq-client 

mattrpav:
Needed for Virtual Thread support


mattrpav:
bad commit, creating new branch and pr

PR #1508, opened 2025-10-10
[AMQ-9783] Centralize XML safe parsing settings in a single class.

sergio-d-lemos:
Centralize settings for DocumentBuilderFactory and TransformerFactory in a single class, so we have a way to apply always the same security settings.

Unfortunately there is no single shared module, so I can't use  XmlFactories class in https://github.com/apache/activemq/blob/main/activemq-client/src/main/java/org/apache/activemq/filter/XPathExpression.java

Sanity tested by using the `activemq create` and by running a broker with the `runtimeConfigurationPlugin` plugin enabled.


jbonofre (approved):
That's a good idea to "centralize" the configuration of the XML parsing.
Thanks !

PR #1522, opened 2025-10-20
NO-REF: Update to spring-webmvc:5.3.41 to address CVE-2024-38819

nshestiuk:
Fix for https://spring.io/blog/2024/09/12/spring-framework-5-3-40-and-6-0-24-available-now

**Impact:** An attacker can craft malicious HTTP requests and obtain any file on the file system that is also accessible to the process in which the Spring application is running.

> **Affected Spring Products and Versions**
> Spring Framework:
>
>    5.3.0 - 5.3.40
>    6.0.0 - 6.0.24
>    6.1.0 - 6.1.13
>    Older, unsupported versions are also affected

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -103,7 +103,7 @@
     <shiro-version>1.13.0</shiro-version>
     <slf4j-version>2.0.16</slf4j-version>
     <snappy-version>1.1.2</snappy-version>
-    <spring-version>5.3.39</spring-version>
+    <spring-version>5.3.41</spring-version>
     <taglibs-version>1.2.5</taglibs-version>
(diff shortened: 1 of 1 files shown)


mattrpav:
Has Broadcom agreed to make Spring 5.3.41 available to everyone? I suspect not, so this change won’t be able to go into the main source tree. 

The better fix would be to remove reliance on spring-mvc altogether. iirc, only a few classes are used and we should be able to use other in-place frameworks or write our own code to remove it.

mattrpav:
Spring v5.3.41 is not open source, so this can't merge :(

PR #1523, opened 2025-10-27
WIP: [AMQ-9798] Refactored Web Console and removed dependency on Spring-MVC.

sergio-d-lemos:
This PR removes the dependency on Spring MVC and does some clean ups in the Web Console. Apologize for the long PR.

* spring-context is used to inject beams into the context
* Created the ActionServlet which plays the role of the DispatcherServlet (from spring-webmvc)
* Removed the applicationContext because I could not find any JSP page using it
* Moved CSRF validation to its own Filter
* Minor clean ups here and there

I tested each page using Chrome and did some sanity tests using Firefox. I was unable to test running the console on an OSGI container, but I can test if we think is necessary.


kenliao94:
Can we also write some tests for the change?

Great initiative!

sergio-d-lemos:
> Can we also write some tests for the change?
> 
> Great initiative!

Yes, if we don't need any major change to the PR I will include some unit tests. At least for the classes I've added

jbonofre:
That's a nice PR, thanks !

At first glance, it looks good to me. I don't think we would need additional tests compared to what we have today.

However, I think the Karaf features have to be updated to cleanup the dependency there as well. I can help on this front.

Let me do a complete review.

sergio-d-lemos:
> That's a nice PR, thanks !
> 
> At first glance, it looks good to me. I don't think we would need additional tests compared to what we have today.
> 
> However, I think the Karaf features have to be updated to cleanup the dependency there as well. I can help on this front.
> 
> Let me do a complete review.

Thanks JB, I'm not very familiar with Karaf, appreciate any help on this front, I can also take a look.

cshannon:
I didn't notice this until now, I think it would be nice to replace Spring entirely (not just MVC) and use Jax-RS and Jersey. That might be a longer term goal for ActiveMQ 7 though

sergio-d-lemos:
> I didn't notice this until now, I think it would be nice to replace Spring entirely (not just MVC) and use Jax-RS and Jersey. That might be a longer term goal for ActiveMQ 7 though

I think this is the medium/long term goal, we started by removing Spring MVC since its a heavy dependency and we don't really get a lot of benefit from it. The Web Console needs some love definitly.

sergio-d-lemos:
Rebased with the latest changes, also added unit tests. 

Moved the logic to redirect to/from `slave.jsp` to its own Servlet Filter.

PR #1524, opened 2025-10-31
AMQ-9801: Support for Dark Mode in web-console

ggrandes:
Minimal change needed to support dark mode when the browser prefers it.

Reference: [AMQ-9801](https://issues.apache.org/jira/browse/AMQ-9801)

darkmode_small

### DIFF
--- a/activemq-web-console/src/main/webapp/styles/dark-mode.css
+++ b/activemq-web-console/src/main/webapp/styles/dark-mode.css
@@ -15,13 +15,9 @@
  * limitations under the License.
  */
 
-/* ===================================================== */
-/* Single place to import all css files used in head.jsp */
-/* ===================================================== */
-
-@import url('sorttable.css');
-@import url('type-settings.css');
-@import url('site.css');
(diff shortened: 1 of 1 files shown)


sergio-d-lemos:
Is it possible to toggle back-and-forth from dark mode and "light" mode? Just as a curiosity, does this improve the web console accessibility?

ggrandes:
@sergio-d-lemos : This CSS is automatic/dynamic and respects the system/browser configuration/preference; if the user does not require dark mode, the inner style/filter is not activated.

```css
@media (prefers-color-scheme: dark)
```
Reference [doc](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)

ggrandes:
Oops, this is new, I could swear that before you could delete the fork and the PR would remain 😓

jbonofre:
@ggrandes I'm not against, but I would suggest to add a button to toggle between mode. And the PR is closed now 😄

PR #1529, opened 2025-11-05
AMQ-9392 - Rework fix for read check Timer leak 

cshannon:
This commit updates the previous fix for preventing a read check Timer
leak by the inactivity monitor when TCP connections fail
This is a follow on to https://github.com/apache/activemq/pull/1119

This updated version has the following improvements:

  1) The logic to cancel timers and shutdown the executor has been
consolidated into one location and only done during stopping of the
transport. Previously, the update could stop the task/executor during
normal broker operation (see next point)
  2) stopConnectCheckTask() is called under normal circumstances during
initialization of a connection and the previous change caused the read
task to be cancelled and the executor to be stopped, only to be
immediately restarted again during the first connection creation.
  3) Null checks were added to guarantee no null pointer issues when
referencing tasks/timers during stop
  4) Helper methods and some comments were added to simplify the code
and make it easier to understand what is going on.


jbonofre (approved):
Much cleaner approach. Thanks !

jbonofre:
I'm just waiting Jenkins before merging (it worked locally on my machine).

mattrpav (commented):
LGTM

PR #1532, opened 2025-11-11
[NO-JIRA] Update enforcer-plugin to releasing with JDK 21

mattrpav:
I tested building with JDK 21 and then running with JDK 17 without a problem.

The docs for enforcer plugin make it read that this setting impacts _runtime_, but it does _**not**_ in fact change the compiler-plugin values used for source/target/release.

NOTE: This needs to be merged into main after main is updated to 6.3.0-SNAPSHOT

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1133,7 +1133,7 @@
             <configuration>
               <rules>
                 <requireMavenVersion>
-                  <version>3.0.5</version>
+                  <version>3.6.3</version>
                 </requireMavenVersion>
               </rules>
             </configuration>
@@ -1337,6 +1337,39 @@
       </activation>
       <build>
         <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-enforcer-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>enforce-maven</id>
+                <goals>
+                  <goal>enforce</goal>
+                </goals>
+                <configuration>
+                  <rules>
+                    <requireMavenVersion>
+                      <version>3.6.3</version>
+                    </requireMavenVersion>
+                  </rules>
+                </configuration>
+              </execution>
+              <execution>
+                <id>enforce-java-version</id>
+                <goals>
+                  <goal>enforce</goal>
+                </goals>
+                <configuration>
+                  <rules>
+                    <requireJavaVersion>
+                      <version>[21,)</version>
+                      <message>You must use Java 21+ during release to support multi-release jars</message>
+                    </requireJavaVersion>
+                  </rules>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>


gemmellr:
The java version rule is just about the JVM running the enforcement, i.e the current runtime. It has nothing to do with the compiler settings or what the build output compiles for / runs on, which is all independent.

Wasnt the idea to only require _release builds_ are done with JDK 21+? This change also prevents building testing on 17+, which would ideally still be a thing if supporting/targeting it?

mattrpav:
@gemmellr good catch. I'll move this change inside the apache-release profile (edited for exact profile name)

mattrpav:
Output when building with JDK 17 and apache-release profile

```
% mvn -Papache-release clean install -DskipTests=true
...
[ERROR] Rule 0: org.apache.maven.enforcer.rules.version.RequireJavaVersion failed with message:
[ERROR] You must use Java 21+ to release.
```

pom.xml
```diff
@@ -1340,6 +1340,39 @@
...
+            <executions>
+              <execution>
+                <id>enforce-maven</id>
+                <goals>
+                  <goal>enforce</goal>
+                </goals>
+                <configuration>
+                  <rules>
+                    <requireMavenVersion>
+                      <version>3.0.5</version>
```
jbonofre: Maven 3.0.x doesn't support JDK21.
I suggest to set Maven to 3.9.11.

gemmellr:
> This change looks good to me for the enforcer.
> 
> However, if we merge it, it means Jenkins will fail. I suggest to update `Jenkinsfile` as part of this PR.

I dont see mention of it enabling the apache-release profile, so I dont think it will fail unless thats being introduced elsewhere.

pom.xml
gemmellr: I'd definitely use something much newer than 3.0.5 since most current plugins dont work with versions that old (baseline is 3.6.3 currently I think), though I probably wouldnt require 3.9.11 either (even though it is what I am now using) but perhaps some older 3.9.x.

mattrpav: Updated to 3.6.3

pom.xml
```diff
@@ -1133,7 +1133,7 @@
             <configuration>
               <rules>
                 <requireMavenVersion>
-                  <version>3.0.5</version>
+                  <version>3.6.3</version>
```
gemmellr, line 1136: The parent pom 37 now requires Maven 3.9 to build. Given 3.8.x is now entirely out of support and new plugin releases are aimed at 3.9+, I'd say its time to do the same (and have already done so in other projects). I'd also defer to its enforcer definition (by removing this one entirely) at this point to simplify the pom.

You could also get rid of the java version check below this one (but not the other one in the release profile) and defer to the parents one. They have also defined a new property in v37 to link all the java verison enforcement and compiler settings together, though personally I prefer having the compiler properties set in the components own pom and so am still setting them regardless.

pom.xml
```diff
@@ -1337,6 +1337,39 @@
...
+                  <goal>enforce</goal>
+                </goals>
+                <configuration>
+                  <rules>
+                    <requireMavenVersion>
+                      <version>3.6.3</version>
+                    </requireMavenVersion>
+                  </rules>
+                </configuration>
+              </execution>
```
gemmellr, line 1356: Could similarly remove this maven version enforcement.

jeanouii:
@mattrpav I'm closing this one because with the Java 25 work, we are now enforcing Maven 3.9 and Java 24 for the release. So I don't think this one is necessary anymore. If I'm wrong, my apologies and please reopen.

PR #1533, opened 2025-11-13
[#] JDK 25 support (multiple JIRA)

mattrpav:
Apache Jenkins JDK 25 test passed except for two tests

TODO:
 - [ ] JmsSchedulerTest.testUpdatesAppliedToIndexBeforeJournalShouldBeDiscarded failing -- log4j2 test appender is not picking up events, but everything is working correctly
 - [ ] JmsSchedulerTest.testCron is hanging indefinitely

ref: https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/PR-1533/6/

PR #1534, opened 2025-11-14
[#] NO-JIRA: SecurityManager Shim and MRJAR

jeanouii:
This PR leverages Matt's PR [#] NO-JIRA: Update Jenkinsfile to add JDK 25](https://github.com/apache/activemq/pull/1533)

It is meant to check CI status on tests and gather community feedback. We might prefer to create a PR against Matt's branch to join efforts.


activemq-compat/src/main/java/org/apache/activemq/compat/sm/SecurityManagerShim.java
```diff
@@ -0,0 +1,199 @@
...
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionException;
+
+/*
+ * SecurityManager related shim.
+ * This specific class uses legacy methods toward usage on Java 17 - 23.
+ *
+ * The API of this class must be kept the same as the Java 24+ implementation
+ * variant of this class which can be found at:
+ * src/main/java24/org/apache/activemq/artemis/utils/sm/SecurityManagerShim.java
```
gemmellr, line 35: All of the Javadoc in this class is inaccurate now as you havent updated it to reflect how/where it is being used.

activemq-compat/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+  <name>ActiveMQ :: Compat</name>
+  <description>The ActiveMQ Compatibility (MRJAR)</description>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <extensions>true</extensions>
+        <inherited>true</inherited>
```
gemmellr: Inherited being true seems to make little sense here, no children and its true by default

activemq-compat/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+              !com.google.errorprone.annotations,
+              !com.google.errorprone.annotations.concurrent,
+              com.thoughtworks.xstream.*;resolution:="optional",
+              javax.jmdns.*;resolution:="optional",
+              *
+            </Import-Package>
+            <Private-Package>
+              com.google.errorprone.annotations,
+              com.google.errorprone.annotations.concurrent
+            </Private-Package>
```
gemmellr: At least some of these seem like they should be superfluous.

activemq-compat/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+            <artifactId>maven-compiler-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>java21-compile</id>
+                <phase>compile</phase>
+                <goals>
+                  <goal>compile</goal>
+                </goals>
+                <configuration>
+                  <release>21</release> <!-- Specific Java version for alternative classes -->
```
gemmellr: I probably wouldnt change anything for 21 users personally...but if going with 21 for this, then you may want to avoid, or split out, the rest of the shim beyond the Subject bit since if the methods are ever used the current approach would break anyone actually using the SecurityManager on 21-23 currently.

gemmellr (commented):
This should clearly be going against a Jira.

activemq-compat/src/main/java/org/apache/activemq/compat/sm/SecurityManagerShim.java
jeanouii: Yes, JIRA was out yesterday, but I can create the issue. Need to check with @jbonofre if there is a parent issue I need to create a sub-task.
And also yes, I need to update the entire javadoc. That was a plain copy to follow up on the email thread.

gemmellr: Yeah the ASF has been under DDoS  attack for the last couple days, but Jira would usually load after another attempt or two if it failed initially.

mattrpav: @jeanouii no need to do a JIRA. Keep the convo here.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/AnnotatedMBean.java
```diff
@@ -205,8 +204,7 @@ public Object invoke(String s, Object[] objects, String[] strings) throws MBeanE
         objects = (objects == null) ? new Object[]{} : objects;
         JMXAuditLogEntry entry = null;
         if (audit != OFF) {
-            // [AMQ-9563] TODO: JDK 21 use Subject.current() instead
-            Subject subject = Subject.getSubject(AccessController.getContext());
+            Subject subject = SecurityManagerShim.currentSubject();
```
mattrpav, line 207: Do we really need the shim? I'm wondering if we just ship two AnnotatedBean classes using MR jar and switch the JAAS API implementation when getting the Subject.

activemq-compat/src/main/java/org/apache/activemq/compat/sm/SecurityManagerShim.java
gemmellr: The Jira is for issue and commit tracking, not discussion of the code. Changes to make things work on Java 24+ when they don't currently, seems pretty clearly deserving of being committed against a Jira. Almost everything should be.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/AnnotatedMBean.java
gemmellr: If there are no other SM / related API usage to handle then you wouldn't need it all, just the Subject method.

However I'm not seeing how duplicating a sizable complex class that may need to be updated, just to change a couple lines, would really be better than having 2 new tiny classes , that likely won't ever change, to do the switch out and avoid the need to duplicate the complex class? 

For sure the classes don't need to be in their own module if that's what you dislike about it.

activemq-compat/src/main/java/org/apache/activemq/compat/sm/SecurityManagerShim.java
cshannon: This PR is definitely going to need a Jira before it is merged but that can be done before it is ready to merge

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/AnnotatedMBean.java
cshannon: If we only need a small part of the Shim around handling Subject then we should just use that piece for now to keep the changes minimal.

mattrpav:
Updated approach w/ all three changes here: https://github.com/apache/activemq/pull/1533

 - [x] Jenkinsfile updated for JDK 25 option
 - [x] KahaDBPersistenceAdapter adding DataFileFactory (to support testing w/o SecurityManager)
 - [x] Add small SecurityShim for simply resolving Subject to activemq-broker as a Multi-Release jar  

Two Jenkins jobs running on Apache CI, one with JDK 17 and another kicked off using JDK 25

jeanouii:
@mattrpav We can probably close this PR then and keep going with what you pulled in your own branch/PR

PR #1535, opened 2025-11-16
[AMQ-9806] Add H2 database support

mattrpav:


activemq-jdbc-store/src/main/java/org/apache/activemq/store/jdbc/adapter/H2JDBCAdapter.java
```diff
@@ -0,0 +1,38 @@
...
+public class H2JDBCAdapter extends BytesJDBCAdapter {
+
+    @Override
+    public void setStatements(Statements statements) {
+        statements.setBinaryDataType("BLOB");
+        super.setStatements(statements);
+    }
+
+    @Override
+    public String limitQuery(String query) {
```
jeanouii, line 34: I had a PR open to use the limitQuery everywhere because it's not in the default adapter (super class). Just an FYI

activemq-unit-tests/src/test/java/org/apache/activemq/store/jdbc/h2/H2DB.java
```diff
@@ -0,0 +1,45 @@
...
+    public static DataSource createDataSource(String db) throws IOException {
+        var ds = new JdbcDataSource();
+        ds.setURL(createURL(db));
+        ds.setUser(getUser());
+        ds.setPassword(getPassword());
+        return ds;
+    }
+
+    public static String createURL(String db) {
+        return "jdbc:h2:./target/h2-db/" + db + "/" + db + "-h2.db;DB_CLOSE_DELAY=-1";
```
jeanouii, line 35: Good to have a database per test so we can later use parallelism if needed.

PR #1536, opened 2025-11-17
AMQ-9808: Dead lock when using CRON schedule with AMQ_SCHEDULED_REPEAT

jeanouii:
This is not related to the Java 25 work because the bug can be reproduced in 5.18.x branch for instance. It has been broken for a long time I believe.

When using repeat with a cron scheduler, we need to deffer the actual scheduling when process is done to avoid a dead lock to happen because we want to write a command while already writting a command.

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/scheduler/JobSchedulerImpl.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/scheduler/JobSchedulerImpl.java
@@ -268,6 +268,16 @@ public class JobSchedulerImpl extends ServiceSupport implements Runnable, JobSch
         this.store.store(update);
     }
 
+    private void doSchedule(final List<Closure> toSchedule) {
+        for (Closure closure : toSchedule) {
+            try {
+                closure.run();
+            } catch (final Exception e) {
+                LOG.warn("Failed to schedule job", e);
+            }
+        }
+    }
+
     private void doRemove(final List<Closure> toRemove) throws IOException {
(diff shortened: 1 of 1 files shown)


activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/scheduler/JobSchedulerImpl.java
```diff
@@ -797,6 +804,12 @@ protected void mainLoop() {
                 } finally {
                     this.store.readUnlockIndex();
 
+                    // deferred execution of all jobs to be scheduled to avoid deadlock with indexLock
+                    for (Runnable command : toSchedule) {
+                        command.run();
```
mattrpav: I think we may need some sort of catch-and-continue pattern here, since if scheduling one fails due to IOException, the remaining would all fail to _try_ to schedule.

jeanouii: Thanks @mattrpav 
I'll give it a try and push a better exception handling

jeanouii:
@jbonofre I don't know why Jenkins is complaining. I don't have enough permissions to trigger a new build, so I'm not sure what else to do.

This is clearly a bug that should be fixed. Any comments?

jbonofre:
@jeanouii I'm triggering a new run on Jenkins.

jbonofre:
https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/PR-1536/3/

PR #1546, opened 2025-11-21
AMQ-9810 - Add additional validation for MQTT control packets

cshannon:
Validate that the remaining length field is the correct number of bytes


tabish121:
LGTM

PR #1547, opened 2025-11-21
AMQ-9810 - Add additional validation for MQTT wireformat

cshannon:
Validate that the remaining length field is the correct number of bytes in MQTTWireFormat


tabish121:
Having a feeling of déjà vu but looks good the second time also.

PR #1551, opened 2025-11-26
AMQ-9813 - fix wrong QueueSize for non-persistent message with TTL

lenoch7:
See full description in https://issues.apache.org/jira/browse/AMQ-9813

Simply:
It seems like [8b23e07](https://github.com/apache/activemq/commit/8b23e07) breaks previous contract of `expireOldMessages()` methods, which can caused that expired messages are not "processed", when expiration is triggered by `tryAddMessageLast(MessageReference, long)`, `addMessageFirst(MessageReference)` methods.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
@@ -206,7 +206,19 @@ public class FilePendingMessageCursor extends AbstractPendingMessageCursor imple
      * @throws Exception
      */
     @Override
-    public synchronized boolean tryAddMessageLast(MessageReference node, long maxWaitTime) throws Exception {
+    public boolean tryAddMessageLast(MessageReference node, long maxWaitTime) throws Exception {
+        // Discarding expired message should be done outside of synchronized section (deadlock, see AMQ-5785)
+        List<MessageReference> expiredMessages = new ArrayList<>();
+        boolean isExpiration = tryAddMessageLastInternal(node, maxWaitTime, expiredMessages);
+        for (MessageReference expiredMessage : expiredMessages) {
+            discardExpiredMessage(expiredMessage);
+        }
+        return isExpiration;
+    }
+
+    private synchronized boolean tryAddMessageLastInternal(
+      MessageReference node, long maxWaitTime, List<MessageReference> expiredMessages
+    ) {
         if (!node.isExpired()) {
             try {
                 regionDestination = (Destination) node.getMessage().getRegionDestination();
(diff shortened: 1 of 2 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
```diff
@@ -263,16 +284,16 @@ public synchronized void addMessageFirst(MessageReference node) {
...
                         node.incrementReferenceCount();
                         setCacheEnabled(true);
-                        return;
+                        return expiredMessages;
                     }
                 }
                 if (!hasSpace()) {
                     if (isDiskListEmpty()) {
-                        expireOldMessages();
+                        expiredMessages = expireOldMessages();
```
cshannon, line 292: So I just realized that this is actually could be issue because the expireOldMessages() method is going to decrement the reference on each message, but the messages get copied to that temporary list. The problem is that now that we are hanging onto the list reference (to process later outside of the lock) we are more likely to blow memory because the hasSpace() method will think there is free space but those messages are still held in the heap until later. We are only adding one message but if that message happens to be very large it could be an issue.

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
```diff
@@ -220,7 +232,7 @@ public synchronized boolean tryAddMessageLast(MessageReference node, long maxWai
                 }
                 if (!hasSpace()) {
                     if (isDiskListEmpty()) {
-                        expireOldMessages();
+                        expiredMessages.addAll(expireOldMessages());
```
cshannon, line 235: Same comment as in addMessageFirst

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
lenoch7: Thank you for very quick feedback and analyze. I think, I got the point, which you mean. But unfortunately I have no idea, how to change/improve it, except the way you already mentioned - put new message to disk (temp storage) immediately without attempt to "dispose" expired messages before switch to disk  (if I really got the point).

Maybe it will be better to close this PR without merge to give you more comfortable way to make fix/changes/improvements. I am happy, you got the problem, which I tried to report.

lenoch7:
I am sorry for of topic, but...
I noticed, that one of build - Build 2 (https://ci-builds.apache.org/job/ActiveMQ/job/ActiveMQ/job/PR-1551/2/cloudbees-pipeline-explorer/) failed (`org.apache.activemq.web.RestTest.testConsumeAsyncTimeout`). I don't believe that it is caused by this PR (change). I guess (maybe), the `errorResult` async request was received as first request (before `result` request), regardless on "position" in the test code (because they are async request). "Stupid" sleep between these two async requests in the test? But it is different story, I am sorry again.

cshannon:
After looking more into this I think it's fine to merge. It's only a single message being added and the memory pressure should not be so high that one message causes an OOM. There are plenty of spots in the broker where we are incrementing/decrementing memory counters and it isn't exact because messages are still in memory and being processed. In fact, queues allow paging in up to 110% of the memory setting to allow dispatch and unblocking. The memory usage tracking is always just an estimate in general and I think the fix makes sense and unlikely to cause any issues.

PR #1553, opened 2025-12-01
AMQ-9813 - Minor updates

cshannon:
Rework handling of expiration array list to avoid allocation when not needed


cshannon:
follow on to #1551

jeanouii:
Looks good @cshannon 

Outside of this PR, I'm wondering if we could leverage ReentrantReadWrite lock to avoid this many synchronized methods?

cshannon:
> Looks good @cshannon
> 
> Outside of this PR, I'm wondering if we could leverage ReentrantReadWrite lock to avoid this many synchronized methods?

Replacing with a readwrite lock would only make sense if we could identify that it would actually help enough to justify the complexity. In a busy broker with messages being both written/read at the same time I'm not sure that is the case, it just kind of depends I guess on how it's being used and where lock contention is.

PR #1555, opened 2025-12-02
[AMQ-9817] Fix long running KahaDBOffsetRecoveryListenerTest on CI

jeanouii:
This is taking about an hour to run.
The @Before code is running every single time for all the test scenario of the class. 

I'm questioning what the benefit of the @Before peace of code. If it's an actual test, I'd pull it into its own @Test method.
But for now, it's burning a lot of time because it does a lot of IO and the CI is slow for IO. Moreover, the directory used in the sendMessage call of the @Before portion is the default one, because the broker is lazily created after using the VM Transport, whereas all the other tests are actually using a different directory. 

Locally on my Mac M1 it runs on 16 min and after the removal, it runs in 2'30. 

If everyone is happy with it, I can create a JIRA and rename this PR and update the commit message

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/store/kahadb/KahaDBOffsetRecoveryListenerTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/store/kahadb/KahaDBOffsetRecoveryListenerTest.java
@@ -17,11 +17,12 @@
 
 package org.apache.activemq.store.kahadb;
 
-import java.io.File;
-import java.io.IOException;
-import java.util.LinkedList;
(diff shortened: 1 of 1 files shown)


jeanouii:
[INFO] Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 522.1 s -- in org.apache.activemq.store.kahadb.KahaDBOffsetRecoveryListenerTest

Build is green and the test went from an hour, to less than 10 minutes. 
@jbonofre Should I create a JIRA for it and update the PR + commit message?

jbonofre (approved):

PR #1556, opened 2025-12-08
[AMQ-9815] Add additional attributes to ConnectorView

mattrpav:


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectorView.java
```diff
@@ -151,4 +153,82 @@ public boolean isAutoStart() {
     public boolean isStarted() {
         return this.connector.isStarted();
     }
+
+    @Override
+    public String getConnectURI() {
+        URI tmpConnectUri = null;
+        try {
+            tmpConnectUri = this.connector.getConnectUri();
```
jeanouii, line 162: I find it more readable to do a return this.connector.getConnectUri() and return null in the catch block.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectorView.java
```diff
@@ -151,4 +153,82 @@ public boolean isAutoStart() {
...
+            publishableConnectURI = this.connector.getPublishableConnectURI();
+        } catch (Exception e) {}
+
+        return (publishableConnectURI != null ? publishableConnectURI.toString() : null);
+    }
+
+    @Override
+    public String getBrokerInfoString() {
+        var tmpBrokerInfo =  this.connector.getBrokerInfo();
+        return (tmpBrokerInfo != null ? tmpBrokerInfo.toString() : null);
```
jeanouii: if getBrokerInfo() can really return null, you probably could add a null check above in other methods.

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectorView.java
```diff
@@ -151,4 +153,82 @@ public boolean isAutoStart() {
...
+            tmpConnectUri = this.connector.getConnectUri();
+        } catch (Exception e) {}
+        return (tmpConnectUri != null ? tmpConnectUri.toString() : null);
+    }
+
+    @Override
+    public String getPublishableConnectURI() {
+        URI publishableConnectURI = null;
+        try {
+            publishableConnectURI = this.connector.getPublishableConnectURI();
```
jeanouii, line 171: Same as previous

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectorView.java
jeanouii: There is a getBrokerInfo() method already available

PR #1563, opened 2025-12-16
ActiveMQ on Java 25

jeanouii:
No need to review, this is a temp PR I'll kill whenever I have more visibility on what now fails on Java 25

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1122,39 +1122,6 @@
 
     <plugins>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-enforcer-plugin</artifactId>
-        <executions>
-          <execution>
-            <id>enforce-maven</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <configuration>
-              <rules>
-                <requireMavenVersion>
-                  <version>3.0.5</version>
-                </requireMavenVersion>
-              </rules>
-            </configuration>
-          </execution>
-          <execution>
-            <id>enforce-java-version</id>
-            <goals>
-              <goal>enforce</goal>
-            </goals>
-            <configuration>
-              <rules>
-                <requireJavaVersion>
-                  <version>[17,)</version>
-                  <message>You must use Java 17+ to build.</message>
-                </requireJavaVersion>
-              </rules>
-            </configuration>
-          </execution>
-        </executions>
-      </plugin>
-      <plugin>
         <groupId>org.apache.geronimo.genesis.plugins</groupId>
         <artifactId>tools-maven-plugin</artifactId>
         <executions>
@@ -1271,29 +1238,6 @@
   </build>
   <reporting>
     <plugins>
-      <!--
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-jxr-plugin</artifactId>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-surefire-report-plugin</artifactId>
-        <version>${maven-surefire-plugin-version}</version>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-checkstyle-plugin</artifactId>
-      </plugin>
-      <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
-        <artifactId>maven-pmd-plugin</artifactId>
-      </plugin>
-      <plugin>
-        <groupId>org.codehaus.mojo</groupId>
-        <artifactId>cobertura-maven-plugin</artifactId>
-      </plugin>
-      -->
       <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>taglist-maven-plugin</artifactId>
@@ -1337,6 +1281,26 @@
       </activation>
       <build>
         <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-enforcer-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>enforce-java-version</id>
+                <goals>
+                  <goal>enforce</goal>
+                </goals>
+                <configuration>
+                  <rules>
+                    <requireJavaVersion>
+                      <version>[24,)</version>
+                      <message>You must use Java 24+ to build (we leverage MRJAR).</message>
+                    </requireJavaVersion>
+                  </rules>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
           <plugin>
             <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-surefire-plugin</artifactId>


activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
```diff
@@ -89,11 +89,8 @@
 import org.apache.activemq.store.kahadb.disk.index.BTreeIndex;
 import org.apache.activemq.store.kahadb.disk.index.BTreeVisitor;
 import org.apache.activemq.store.kahadb.disk.index.ListIndex;
-import org.apache.activemq.store.kahadb.disk.journal.DataFile;
-import org.apache.activemq.store.kahadb.disk.journal.Journal;
+import org.apache.activemq.store.kahadb.disk.journal.*;
```
mattrpav: I know Intellij defaults to use wildcard imports. However, I'm not a fan b/c it makes discovery via searching less deterministic.

Not a blocker, more of a mention to get other's feedback

jeanouii: Yes, you are correct IntelliJ...
I agree with the feedback because I share it and I usually set an high value to make sure Intellij does not abbreviate using *

I'll fix it after the end of the build.

activemq-broker/src/main/java24/org/apache/activemq/broker/jmx/SubjectShim.java
```diff
@@ -0,0 +1,35 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.broker.jmx;
+
+import javax.security.auth.Subject;
+import java.security.AccessController;
```
gemmellr: Looks like you have some stale commits from another PR.

Same months-old feedback [as on the other PR](https://github.com/apache/activemq/pull/1533#discussion_r2533507877), this is unused and shouldnt be here (and will break things when the SecurityManager APIs are eventually removed)

activemq-broker/pom.xml
```diff
@@ -263,5 +263,32 @@
...
+                </goals>
+                <configuration>
+                  <release>24</release> <!-- Specific Java version for alternative classes -->
+                  <compileSourceRoots>${project.basedir}/src/main/java24</compileSourceRoots>
+                  <multiReleaseOutput>true</multiReleaseOutput>
+                </configuration>
+              </execution>
+            </executions>
+          </plugin>
+        </plugins>
```
gemmellr, line 300: Elsewhere after the compile defintion I also just brought forward the jar preparation in the multi-release module, to before the test phase rather than after, thus making the jar and multi-release classes in it available to the other modules tests.

```
<plugin>
  <artifactId>maven-jar-plugin</artifactId>
  <!-- Do jar creation earlier, before the "test" phase, so that the
       multi-release-jar content is visible to any dependent modules,
       enabling them to use the relevant classes during their tests. -->
  <executions>
    <execution>
      <id>default-jar</id>
      <phase>process-test-classes</phase>
      <goals>
        <goal>jar</goal>
      </goals>
    </execution>
  </executions>
</plugin>
```

activemq-client/src/main/java/org/apache/activemq/thread/TaskRunnerFactory.java
```diff
@@ -273,13 +267,20 @@ protected TaskRunner createVirtualThreadTaskRunner(Executor executor, Task task,
...
         }
     }
 
+    private void assertJDK21VirtualThreadSupport() {
+        if(!(Runtime.version().feature() >= 21)) {
+            LOG.error("Virtual Thread support requires JDK 21 or higher");
+            throw new IllegalStateException("Virtual Thread support requires JDK 21 or higher");
+        }
+    }
+
```
gemmellr, line 283: The cleanup seems fine, but a thought for later would be....is there any need for all the reflection above and this runtime version checking, when the client jar is already a multi-release jar requiring Java 21 (and soon 24[/25]) to release it? Could maybe have a [couple variants of] trivial delegate class.

activemq-unit-tests/pom.xml
```diff
@@ -465,6 +465,18 @@
...
             <exclude>**/StartAndConcurrentStopBrokerTest.java</exclude>
             <exclude>**/BlobTransferPolicyUriTest.java</exclude>
+
+            <!--
+              added when running under Java 21+, but it can't run as part of normal tests
+              because Maven Surefire will use activemq-client/target/classes instead of the MRJAR
+
+              This test needs to run as part of the integration test phase to ensure it's using the packaged version
+              We use failsafe instead which is a better choice for integration tests. See bellow the configuration.
+            -->
```
gemmellr: See earlier comment for potential route to avoid such config as this and the later inversion of it.

activemq-unit-tests/pom.xml
```diff
@@ -507,6 +519,37 @@
           </dependency>
         </dependencies>
       </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-failsafe-plugin</artifactId>
+        <version>3.5.2</version>
```
gemmellr: If still adding it (rather than omitting from previous suggestion), the plugin version should be managed rather than specified here, and match the surefire version (it definitely already is managed by the apache parent, though I believe activemq might be overriding the surefire version prop somewhat superflously)

gemmellr (commented):
This should really be enforcing that Java 24+ is used for the release process since its required for this to be of any worth to end users.

An earlier PR is still open to require 21+ since Java 21+ wasnt used for the previous release when it should have been, though that PR (#1532) requires changes as its somewhat [out of date now](https://github.com/apache/activemq/pull/1532#discussion_r2727819960)

(I required 25 to release elsewhere, saves changing it later and noone should be really be using 24 now anyway)

jeanouii:
@gemmellr Thanks a lot for the detailed review. I grab back the branch and address the feedback.

activemq-client/src/main/java/org/apache/activemq/thread/TaskRunnerFactory.java
jeanouii: This was a refactoring of @mattrpav 's code. So I'd let him reply.
I agree that having the assertion to fail would be weird because that would mean the JVM could not read the MRJAR but still run in some way this class.

activemq-unit-tests/pom.xml
jeanouii: I tried pushing this and I'll monitor the results to see if everything is ok

activemq-unit-tests/pom.xml
jeanouii: Same as above

jeanouii:
All feedback addressed
Ok to merge @gemmellr  and @mattrpav ?

gemmellr:
I'll leave the test config changes to Matt, not sure whats wanted or has been discussed there.

It doesnt look like you covered this bit of my earlier review. Either the original PR needs updated and merged before this one so this can update things, or this one should just do the necessary, to avoid a repeat of the previous releasing issue.

> This should really be enforcing that Java 24+ is used for the release process since its required for this to be of any worth to end users.
> 
> An earlier PR is still open to require 21+ since Java 21+ wasnt used for the previous release when it should have been, though that PR (#1532) requires changes as its somewhat [out of date now](https://github.com/apache/activemq/pull/1532#discussion_r2727819960)
> 
> (I required 25 to release elsewhere, saves changing it later and noone should be really be using 24 now anyway)

https://github.com/apache/activemq/pull/1563#pullrequestreview-3755585650

jeanouii:
@gemmellr I was waiting for Matt to speak, but you are correct, I'll add that in the same PR so it's autonomous. I think it need a JIRA as well.

jeanouii:
@gemmellr I've now added the enforcer plugin in the apache-release profile to make sure we release the MRJAR supporting Java 21 and Java 24.
I've removed the rule for the Maven version even though to be honest, I prefer it to be explicit like plugin versions or java compiler version. It's a personal preference and I'm fine if community prefers to yank it.
Let me know if that's ok, so I can create the issue in JIRA and then, squash update the commit messages to clean everything up

gemmellr:
> @gemmellr I've now added the enforcer plugin in the apache-release profile to make sure we release the MRJAR supporting Java 21 and Java 24. I've removed the rule for the Maven version even though to be honest, I prefer it to be explicit like plugin versions or java compiler version. It's a personal preference and I'm fine if community prefers to yank it. Let me know if that's ok, so I can create the issue in JIRA and then, squash update the commit messages to clean everything up

There is a property used in the parent for the maven version enforcement, you could set that if you wanted it to be specified locally more obviously.

It is personal preference as you say. I dont know what others here might prefer. I definitely prefer to use the parent pom for a lot that is possible (but not everything; I set/override the compiler properties locally) in order to make our poms more concise. Most especially all the plugin versions possible. Users dont really [need to] care about _our_ plugin versions (they have their own plugin versions) so long as things still work, and the [constant plugin version upgrade churn](https://github.com/apache/activemq/pulls?q=is%3Apr+in%3Atitle+%22plugin%22) otherwise mostly just distracts from other far more meaningful dependency etc changes that users actually are more affected by and likely to be more interested in. I find using the parent versions removes a bunch of that noise, and using the same parent across different components can also mean more consistency among which plugins are being used (albeit, in general all the plugins do then use wildly inconsistent deps, so you still end up needing to grab lots of different versions of the same plugin deps regardless). Its still trivial to see what is actually being used with e.g help:effective-pom, or just running the build.

jeanouii:
Yeah, I have similar preferences. Sometimes I prefer to just be explicit as opposed to think other will go a research. I have bad experience on this. Also sometimes updating just one thing (parent pom for example) will break everything. So parent pom is just an organizational (group) thing for me. I tend myself to put as minimum as possible inside. Add checkstyle rules, deployments rules, rat, and plugins required for that. 
Then everything is explicit in project parent pom. That's my preference and I understand we don't have all the same experience and preferences. 
If current version (in PR) looks like a good compromise, I'm fine with it.
Like we say in other projects, as soon as it moves the project forward and it does not break things, we can adjust if needed in the future. It does not need to be pixel perfect.

.github/workflows/ci-nightly.yml
```diff
@@ -69,7 +69,7 @@ jobs:
           java-version: ${{ matrix.java-version }}
           distribution: temurin
       - name: Test
-        run: mvn -B -e -fae test
+        run: mvn -B -e -fae verify
```
jbonofre, line 72: Any specific reason to use `verify` here more than `test` ? I'm just curious.

.github/workflows/ci-quick.yml
```diff
@@ -64,10 +64,11 @@ jobs:
       - name: Set up JDK
         uses: actions/setup-java@v4
         with:
-          java-version: 17
+          # temporary using 25 until all tests pass on Java 25 - to be changed back to 17 for quick. 25 will run nightly only
+          java-version: 25
           distribution: temurin
       - name: Test
-        run: mvn -B -e -fae test
+        run: mvn -B -e -fae verify -Pactivemq.tests-quick
```
jbonofre, line 71: I agree with using `activemq.tests-quick` profile. Just curious about `verify` vs `test`.

.github/workflows/ci-nightly.yml
jeanouii: Yes it was mostly due to the introduction of failsafe to delay the execution of tests using the MRJAR because the jar is created only after the test phase. So tests using MRJAR were postponed to run only during the integration test phase. So if you use test and not verify you won't run them.
In the end, I could have reverted this because Robbie requested instead to change the phase of the JAR plugin so the MRJAR is built in the process classes phase and therefor it's visible to the test plugin. Does not hurt. Could have been removed

PR #1566, opened 2025-12-18
AMQ-9823 - properly clear ack set from ackAndPreparedMap

cshannon:
Fix the logic that removes prepared acks from the ackAndPreparedMap that were loaded during XA recovery on commit/rollback so that if the set of acks for a key is empty the set it is removed from the map

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java
@@ -465,7 +465,7 @@ public class KahaDBStore extends MessageDatabase implements PersistenceAdapter, 
                         Set<String> ackedAndPrepared = ackedAndPreparedMap.get(key);
                         if (ackedAndPrepared != null) {
                             ackedAndPrepared.remove(id);
-                            if (ackedAndPreparedMap.isEmpty()) {
+                            if (ackedAndPrepared.isEmpty()) {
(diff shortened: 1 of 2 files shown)


cshannon:
The org.apache.activemq.bugs.AMQ6815Test.testHeapUsage test failed but that is a flaky test that I have seen fail before, I will rerun again but we probably need to fix up that test

cshannon:
The failure was caused by a port issue which happens. It looks like we are no longer using the re-run failed tests flag which I feel like is kind of required for projects like this just because it's hard to avoid the following types of issues (ie ports in use) 

```
[ERROR]   JMSConsumerTest>CombinationTestSupport.runBare:107->testPrefetch1MessageNotDispatched:803
[ERROR] Errors: 
[ERROR]   ActiveMQXASslConnectionFactoryTest>CombinationTestSupport.runBare:113->CombinationTestSupport.runBare:107->ActiveMQSslConnectionFactoryTest.testCreateFailoverTcpConnectionUsingKnownPort:106->ActiveMQSslConnectionFactoryTest.createBroker:339 » IO Failed to bind to server socket: tcp://localhost:61610?wireFormat.tcpNoDelayEnabled=true due to: java.net.BindException: Address already in use
```

@mattrpav - The other interesting thing is both my recent PRs jenkins is not finding the test output so it's not showing the failed tests. I have to manually dig through the output to find the failed test. I'm wondering if the parallel test stuff is causing issues with jenkins finding the file.

The output shows: `No test report files were found. Configuration error?`

PR #1567, opened 2025-12-19
AMQ-9824 - Cleanup code in KahaDB classes

cshannon:
This cleanups the MessageDatabase and KahaDBStore classes.

This commit includes the following:

* Fixes the scope of several methods and types. For example, there were many cases where protected methods were referencing types that were package scope.
* Simplified the code by replacing anonymous methods with lambdas
* removed unused methods and parameters
* removed unnecessary casts
* cleaned up the use of generics where types could be inferred
* Replaced the ReentrantReadWriteLock that was used for indexLock with ReentrantLock becuase only the write lock was ever being used (the page file doesn't support concurrent reads right now). This should provide a small performance/memory improvement and simplifies the code a bit.
* removed unnecessary null initializations
* cleaned up logging to remove string concatenation and instead use parameters
* removed method overrides that are the same as the parent or just call the super method
* removed unused checked exceptions from method's throws
* marked inner classes as static when possible

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java
@@ -76,17 +76,14 @@ import org.apache.activemq.store.TransactionStore;
 import org.apache.activemq.store.kahadb.data.KahaAddMessageCommand;
 import org.apache.activemq.store.kahadb.data.KahaDestination;
 import org.apache.activemq.store.kahadb.data.KahaDestination.DestinationType;
-import org.apache.activemq.store.kahadb.data.KahaLocation;
 import org.apache.activemq.store.kahadb.data.KahaRemoveDestinationCommand;
 import org.apache.activemq.store.kahadb.data.KahaRemoveMessageCommand;
 import org.apache.activemq.store.kahadb.data.KahaSubscriptionCommand;
 import org.apache.activemq.store.kahadb.data.KahaUpdateMessageCommand;
 import org.apache.activemq.store.kahadb.disk.journal.Location;
 import org.apache.activemq.store.kahadb.disk.page.Transaction;
-import org.apache.activemq.store.kahadb.disk.page.Transaction.CallableClosure;
 import org.apache.activemq.store.kahadb.disk.util.SequenceSet;
 import org.apache.activemq.store.kahadb.scheduler.JobSchedulerStoreImpl;
-import org.apache.activemq.usage.MemoryUsage;
 import org.apache.activemq.usage.SystemUsage;
 import org.apache.activemq.util.IOExceptionSupport;
 import org.apache.activemq.util.ServiceStopper;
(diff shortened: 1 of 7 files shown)


cshannon:
Note that this PR includes the change from https://github.com/apache/activemq/pull/1566 , once that is merged i'll merge main back into this branch (or force push) so it is gone from the diff

**Edit:** That PR was merged and I merged main into this branch now.

cshannon:
What is everyone's thoughts on back porting to older branches?

The diff is large because of the cleanup but there should not be breaking changes, it's all just clean up and fixing issues. No public API or anything is effected so in terms of Semver it is fine. Back porting would be nice because if we don't then if we have to fix any future KahaDB bugs we would be more likely to run into some merge conflicts if trying to backport those and these changes are not there as well. There were some JDK 17 stuff that I didn't do such as switch to more modern switch statements and instanceof matching because i wasn't sure if we'd backport to 5.x or not. That stuff could be done if we are not.

I think either way is probably fine, we probably wouldn't be doing a lot with KahaDB in older branches so may not be much of an issue I just figured I'd bring it up.

mattrpav:
I think back-porting is a good idea, just to have continuity a little longer if there are any security backport items. Perhaps we make 5.20.x that is JDK 17-based w/ javax.jms.*.

Interesting about the indexLock was only ever using the write lock.

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java
```diff
@@ -109,8 +106,8 @@ public class KahaDBStore extends MessageDatabase implements PersistenceAdapter,
 
     protected ExecutorService queueExecutor;
     protected ExecutorService topicExecutor;
-    protected final List<Map<AsyncJobKey, StoreTask>> asyncQueueMaps = new LinkedList<>();
-    protected final List<Map<AsyncJobKey, StoreTask>> asyncTopicMaps = new LinkedList<>();
+    final List<Map<AsyncJobKey, StoreTask>> asyncQueueMaps = new LinkedList<>();
```
mattrpav, line 109: What is the thought on narrowing the scope here? What if someone extended the class but it is not in the same package?

cshannon: All the scope changes were done because they were previously incorrect and my IDE flagged it. In this case, [AsyncJobKey](https://github.com/apache/activemq/blob/ef789aad2659334777657b779a6e00fe025937f0/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java#L1674) is package scope and not protected.

It would is impossible for someone to extend the class and use AsyncJobKey anyways, etc so changing the scope didn't cause problems. The two choices were to narrow that line or to expand the scope and my preference is always to narrow vs expand unless we need to. Generally speaking with KahaDB most things should likely just be package scope anyways.

jeanouii:
+1 to backport so that future maintenance is easier.

cshannon:
The tests seem quite flaky at the moment (not sure if it is related to the recent test changes with parallel tests) but every test run seems to produce another unrelated failure

cshannon:
@mattrpav and @jbonofre - Any issues if I merge this and backport? The test failures seem like flaky tests and unrelated and I think the updates here are a nice cleanup

jbonofre:
@cshannon agree to backport.

jeanouii:
This is does not fail frequently as far as I recall. I'll take a look to see if you can improve it.
Meanwhile, It's ok to merge I think

```
[ERROR] Failures: 
[ERROR]   NIOSSLConcurrencyTest.testLoad:133 expected:<100000> but was:<84647>
```

PR #1571, opened 2026-01-08
AMQ-9826 Fix assert according to test purpose

jeanouii:
limit to 1 the Remove…Info message on exception which was meant to be the fix in https://github.com/apache/activemq/commit/6d4459a00c2441822297b06221bb107c5843c390


jeanouii:
Closing for now and using the fix in another PR grouped with other fixes to get a chance to see a green

PR #1607, opened 2026-01-12
Until Java 25 work is finished

jeanouii:
because SecurityManager is also not available in 24


jeanouii:
Build isn't green because of some flacky tests, but changes are quite straight forward. We still have the security manager so there is zero chance until Java 25 work is finish to pass a build under Java 24

@jbonofre maybe we can still merge it?

PR #1609, opened 2026-01-12
Harden test to prevent race condition changing the exceptions.

jeanouii:
No JIRA because I'm not sure whereas we should introduce a delay or add a configuration to make the exception propagation synchronous or a listen


jeanouii:
Closing for now and using the fix in another PR grouped with other fixes to get a chance to see a green

PR #1610, opened 2026-01-12
Facky tests revealed mainly with faster CI (aka Github Actions but also with Java 25)

jeanouii:
Includes various fixes related to race conditions or bugs

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/jms2/ActiveMQJMS2MessageListenerTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/jms2/ActiveMQJMS2MessageListenerTest.java
@@ -77,85 +77,108 @@ public class ActiveMQJMS2MessageListenerTest extends ActiveMQJMS2TestBase {
     @Test
     public void testMessageListener() {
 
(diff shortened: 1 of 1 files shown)


cshannon:
I don't understand why we are combining multiple unrelated changes in one PR? This doesn't make sense at all to combine several Jiras of changes into one commit. We need to be able to isolate issues by jira to rollback or fix

cshannon:
The consumer changes to the client especially need to be reviewed more they are higher risk changes and should be by itself

jbonofre:
OK, the change doesn't look problematic to me.

I propose to keep the test updates and revert the client change for broader discussion then. 

@jeanouii do you mind to create a PR to revert client change ?

cshannon:
@jbonofre - Those changes themselves may or may not be fine, but they need to be done in isolation so that they can be noticed. I didn't even notice it was part of this because the title of the PR only mentions flaky tests and faster CI.

Anything with a bug fix, especially I think needs to be separated out. Also, if it really is a bug we should backport the fix but we may not necessarily want to backport the rest of the commit.

jbonofre:
@cshannon i agree. Sorry about that. My intent was to unblock all the freaking test failures thanks to @jeanouii . I under-estimated the change on the client.

PR #1616, opened 2026-01-16
AMQ-9829 Track prefetched messages for duplicate suppression during failover

jeanouii:
During failover in transacted sessions with async dispatch (MessageListener),
  prefetched messages sitting in the unconsumedMessages buffer were not being
  tracked in previouslyDeliveredMessages. This caused them to be incorrectly
  identified as duplicates on redelivery and poison-acked to the DLQ.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
@@ -113,6 +113,7 @@ public class ActiveMQMessageConsumer implements MessageAvailableConsumer, StatsC
     class PreviouslyDelivered {
         org.apache.activemq.command.Message message;
         boolean redelivered;
+        boolean prefetchedOnly;  // true if message was only prefetched, not delivered to application
 
         PreviouslyDelivered(MessageDispatch messageDispatch) {
             message = messageDispatch.getMessage();
@@ -121,6 +122,12 @@ public class ActiveMQMessageConsumer implements MessageAvailableConsumer, StatsC
         PreviouslyDelivered(MessageDispatch messageDispatch, boolean redelivered) {
             message = messageDispatch.getMessage();
             this.redelivered = redelivered;
+        }
+
+        PreviouslyDelivered(MessageDispatch messageDispatch, boolean redelivered, boolean prefetchedOnly) {
+            message = messageDispatch.getMessage();
+            this.redelivered = redelivered;
+            this.prefetchedOnly = prefetchedOnly;
         }
     }
 
@@ -770,8 +777,12 @@ public class ActiveMQMessageConsumer implements MessageAvailableConsumer, StatsC
                     LOG.debug("{} clearing unconsumed list ({}) on transport interrupt", getConsumerId(), unconsumedMessages.size());
                     // ensure unconsumed are rolledback up front as they may get redelivered to another consumer
                     List<MessageDispatch> list = unconsumedMessages.removeAll();
+                    final boolean isTransacted = session.isTransacted();
                     if (!this.info.isBrowser()) {
                         for (MessageDispatch old : list) {
+                            if (isTransacted) {
+                                capturePrefetchedMessagesForDuplicateSuppression(old);
+                            }
                             session.connection.rollbackDuplicate(this, old.getMessage());
                         }
                     }
@@ -933,6 +944,16 @@ public class ActiveMQMessageConsumer implements MessageAvailableConsumer, StatsC
         if (!isAutoAcknowledgeBatch()) {
             synchronized(deliveredMessages) {
                 deliveredMessages.addFirst(md);
+                if (session.isTransacted()) {
+                    PreviouslyDelivered entry = null;
+                    if (previouslyDeliveredMessages != null) {
+                        entry = previouslyDeliveredMessages.get(md.getMessage().getMessageId());
+                    }
+                    if (entry != null && entry.prefetchedOnly) {
+                        entry.prefetchedOnly = false;
+                        entry.redelivered = true;
+                    }
+                }
             }
             if (session.getTransacted()) {
                 if (transactedIndividualAck) {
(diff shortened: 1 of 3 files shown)


mattrpav:
The change to ActiveMQMessageConsumer needs some discussion, add'l unit tests and perhaps some changes.

A couple things:

1. The map isn't cleared on close()
2. Consumer's are cached a lot so we need to make sure there aren't memory leaks or issues with re-using consumers -- at the minimum we need some unit testing to check for leaks of that new map.
3. If the map is keyed off of transactionId and only one transaction can be opened per-consumer at a time -- why do we need a new class?
 
ref: https://github.com/apache/activemq/pull/1610/commits/354e764ef9429e86cab4bda9625c1eef1ad81aee

jeanouii:
@mattrpav I did not change the logic for the map, but I did make sure it was cleared out. So I'm not sure what you mean. Can you elaborate?

As far as I can see, the map gets cleared on every transaction boundary (commit/rollback) and on consumer close, so there's no memory leak possible, right?

What do you mean by why do we need a new class?

mattrpav:
@jeanouii My initial git commit made it appear that the PreviouslyDeliveredMap was newly added. After digging deeper, I see that your change makes modifications and the map is in fact cleared via doClose() call chain. 

Pulling this over into a separate PR will help make the review easier, and if it is a bug fix we will for sure want to back-port it.

Looks like a good find!

jeanouii:
Of course, I'll cherry pick the other commit that went to another PR and create a PR to revert the commit on main

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
```diff
@@ -933,6 +943,16 @@ private void beforeMessageIsConsumed(MessageDispatch md) throws JMSException {
         if (!isAutoAcknowledgeBatch()) {
             synchronized(deliveredMessages) {
                 deliveredMessages.addFirst(md);
+                if (session.isTransacted()) {
```
mattrpav, line 947: Can you expand on how this is only impacted by transacted sessions?

cshannon:
I plan to take a look at this tomorrow

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
jeanouii: The bug (prefetched messages not redelivered after transaction failure) was related to transactions only. So I did not wanted to impact other scenarios. But there is outside of this a rational. 

The previouslyDeliveredMessages map (and prefetchedOnly) is only used to track rollback/redelivery behavior, which only exists for transacted sessions. In a non‑transacted session there is no rollback; delivery state is driven by the ack mode (AUTO/CLIENT/DUPS_OK).

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
```diff
@@ -771,6 +778,9 @@ void clearMessagesInProgress() {
                     // ensure unconsumed are rolledback up front as they may get redelivered to another consumer
                     List<MessageDispatch> list = unconsumedMessages.removeAll();
                     if (!this.info.isBrowser()) {
+                        if (session.isTransacted()) {
+                            capturePrefetchedMessagesForDuplicateSuppression(list);
```
cshannon: This method iterates over the list and then right below we iterate again to rollback duplicates. Im curious if we can somehow refactor to just iterate once. Maybe something like:

```java
for (MessageDispatch old : list) {
    if (session.isTransacted()) {
        capturePrefetchedMessagesForDuplicateSuppression(old);
    }
    session.connection.rollbackDuplicate(this, old.getMessage());
}
```

```java
private void capturePrefetchedMessagesForDuplicateSuppression(final MessageDispatch pending) {
    if (previouslyDeliveredMessages == null) {
        previouslyDeliveredMessages = new PreviouslyDeliveredMap<>(session.getTransactionContext().getTransactionId());
    }
    if (pending.getMessage() != null) {
        previouslyDeliveredMessages.put(pending.getMessage().getMessageId(), new PreviouslyDelivered(pending, false, true));
    }
    LOG.trace("{} tracking existing transacted {} prefetched ({})", getConsumerId(), previouslyDeliveredMessages.transactionId, pending);
}
```

cshannon (changes requested):
@jeanouii - This looks really good, thanks for the fix! The tests are really well done as well.

The only comment I had I made inline, I noticed we were iterating over the list twice so I made a suggestion to be more efficient to just iterate once. Otherwise this looks pretty good to me to merge. See what you think of my suggestion.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
jeanouii: @cshannon you are totally correct. There is no need to iterate twice. Good catch!

Your proposal works fully. 
We could probably even simply the method like this 

```
private void capturePrefetchedMessagesForDuplicateSuppression(final MessageDispatch pending) {
    if (pending.getMessage() == null) return; // nothing to do
    if (previouslyDeliveredMessages == null) {
        previouslyDeliveredMessages = new PreviouslyDeliveredMap<>(session.getTransactionContext().getTransactionId());
    }
    previouslyDeliveredMessages.put(pending.getMessage().getMessageId(), new PreviouslyDelivered(pending, false, true));
    LOG.trace("{} tracking existing transacted {} prefetched ({})", getConsumerId(), previouslyDeliveredMessages.transactionId, pending);
}
```

jbonofre: That's a good catch from @cshannon . I think we can avoid dual looping here.

jeanouii: All pushed, you can double check @cshannon and @jbonofre if you are happy with it. I squashed the commits so if you are happy, it can be merged straight

cshannon (approved):
LGTM

cshannon:
@jbonofre - i think this is good to merge unless you have anything else or any other comments? i suppose we could try running the tests again after rebasing against the latest test fixes but it's probably also just fine to merge it to main and let the tests run on main

jbonofre (approved):
This change looks good to me. I think the CI "fixes" are incremental, so let me merge this one and move forward to identify new flaky tests.

PR #1621, opened 2026-01-19
Flaky tests GitHub actions

jeanouii:


activemq-mqtt/src/test/java/org/apache/activemq/transport/mqtt/MQTTProtocolConverterTest.java
```diff
@@ -129,7 +129,7 @@ public void run() {
         executorService.awaitTermination(10, TimeUnit.SECONDS);
 
         ArgumentCaptor<RemoveInfo> removeInfo = ArgumentCaptor.forClass(RemoveInfo.class);
-        Mockito.verify(transport, times(4)).sendToActiveMQ(removeInfo.capture());
+        Mockito.verify(transport, times(1)).sendToActiveMQ(removeInfo.capture());
```
jbonofre, line 132: I'm a bit confused by the change from 4 to 1. One time is probably enough (to speed up the tests). I wonder why we put 4 times initially.

activemq-ra/src/test/java/org/apache/activemq/ra/ActiveMQConnectionFactoryTest.java
```diff
@@ -136,7 +136,7 @@ public void testXAResourceReconnect() throws Exception {
         try {
             final TransportConnector transportConnector = brokerService.getTransportConnectors().get(0);
 
-            String failoverUrl = String.format("failover:(%s)?maxReconnectAttempts=1", transportConnector.getConnectUri());
+            String failoverUrl = String.format("failover:(%s)?maxReconnectAttempts=10&initialReconnectDelay=100", transportConnector.getConnectUri());
```
jbonofre, line 140: For the record, this initial reconnect delay could have been set at code level (in the `waitFor()` method). As we are already using `maxReconnectAttempts`, it's fine to use `initialReconnectDelay`.

activemq-ra/src/test/java/org/apache/activemq/ra/ActiveMQConnectionFactoryTest.java
```diff
@@ -165,6 +165,22 @@ public boolean isSatisified() throws Exception {
 
             transportConnector.start();
 
+            // Wait for failover to reconnect and recover() to succeed
+            // The ReconnectingXAResource should handle reconnection transparently
```
jbonofre, line 170: nit: maybe worth to mention that it's related to `maxReconnectDelay` on the transport.

activemq-ra/src/test/java/org/apache/activemq/ra/ActiveMQConnectionFactoryTest.java
```diff
@@ -165,6 +165,22 @@ public boolean isSatisified() throws Exception {
...
+                public boolean isSatisified() throws Exception {
+                    try {
+                        resource.recover(100);
+                        return true;
+                    } catch (Exception e) {
+                        // Still reconnecting
+                        return false;
+                    }
+                }
+            }, 30000, 500));
```
jbonofre: nit: The values here are arbitrary right ? The only risk is that it could be flaky on "slow" machines. It's reasonable though.

I would prefer to use `TimeUnit` here for clarify.

activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/Stomp11Test.java
```diff
@@ -565,13 +565,26 @@ public void testAckMessageWithNoId() throws Exception {
...
+        // could be redelivered before ERROR is received.
         String unsub = "UNSUBSCRIBE\n" + "destination:/queue/" + getQueueName() + "\n" +
                        "id:12345\n\n" + Stomp.NULL;
         stompConnection.sendFrame(unsub);
+
+        // Receive frames until we get the ERROR frame, ignoring any MESSAGE frames
+        // that arrive due to redelivery (especially relevant for SSL transport)
+        StompFrame error = null;
+        for (int i = 0; i < 5; i++) {
+            error = stompConnection.receive();
```
jbonofre: nit: `stompConnection.receive()` is a blocking method afair. We can pass a timeout to the `receive()` method. I would define a timeout on `@Test` or define a timeout here else we can have a hanging test if something's wrong.

activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/Stomp12Test.java
```diff
@@ -158,9 +159,32 @@ public void testClientAckWithoutAckId() throws Exception {
...
 
+        // Unsubscribe immediately to prevent message redelivery while waiting for ERROR
+        String unsubscribe = "UNSUBSCRIBE\n" + "id:1\n\n" + Stomp.NULL;
+        stompConnection.sendFrame(unsubscribe);
+
+        // Receive frames until we get the ERROR frame, ignoring any MESSAGE frames
+        // that arrive due to redelivery (especially relevant for SSL transport)
+        StompFrame error = null;
+        for (int i = 0; i < 5; i++) {
+            error = stompConnection.receive();
```
jbonofre: Same comment here about the `receive()` method.

activemq-unit-tests/src/test/java/org/apache/activemq/EmbeddedBrokerTestSupport.java
```diff
@@ -63,7 +63,10 @@ protected void tearDown() throws Exception {
         if (broker != null) {
             try {
                 broker.stop();
+                broker.waitUntilStopped();
```
jbonofre, line 66: I guess the goal here is to avoid "conflict" between embedded brokers between tests. just curious.

jbonofre (approved):

activemq-mqtt/src/test/java/org/apache/activemq/transport/mqtt/MQTTProtocolConverterTest.java
jeanouii: I have no idea. The goal of the test was to ensure only one call to sendToActiveMQ was made, which was a fix committed at the same time and using an AtomicBoolean to enforce it. So I can tell you.

jeanouii: It's not about speeding the test, it's about the converter implementation which is cristal clear. I think it's more Mockito or the attachment of the agent failed sometimes. 
I updated Mockito and used a regular JavaAgent as opposed to attaching the agent to self. Might have helped revealing the issue especially it was returning all the calls as opposed to the calls of RemoveInfo.

activemq-unit-tests/src/test/java/org/apache/activemq/EmbeddedBrokerTestSupport.java
jeanouii: Exactly. I Github nodes where it's faster, it was causing issues

activemq-ra/src/test/java/org/apache/activemq/ra/ActiveMQConnectionFactoryTest.java
jeanouii: It's not the same level I think.

Doing it in the WaitFor in my opinion does not control the transport but how the test needs to wait to verify the reconnection succeeded.

I've put it in the transport URL because I want to control how quickly the ActiveMQ client retries connection at the transport level.

The second one is the one I wanted in this test.

activemq-ra/src/test/java/org/apache/activemq/ra/ActiveMQConnectionFactoryTest.java
jeanouii: Adding a comment for it. Thank you

activemq-ra/src/test/java/org/apache/activemq/ra/ActiveMQConnectionFactoryTest.java
jeanouii: You're absolutely right about using TimeUnit - I've updated the code for clarity.
I've calculated the values actually :-)
maxReconnectAttempts=10 and exponential backoff up to 30s for the maxReconnectDelay
In practice it's a few seconds, so 30s is more than enough even on a slow machine. 
I've added a timeout to the test itself in case.

activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/Stomp11Test.java
jeanouii: correct. Applied!

activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/Stomp12Test.java
jeanouii: applied

jbonofre:
There is a test timeout on XA but I propose to merge and keep improving.

PR #1622, opened 2026-01-22
Batch of flaky tests fixes

jeanouii:


activemq-ra/src/test/java/org/apache/activemq/ra/ActiveMQConnectionFactoryTest.java
```diff
@@ -137,7 +137,7 @@ public void testXAResourceReconnect() throws Exception {
         try {
             final TransportConnector transportConnector = brokerService.getTransportConnectors().get(0);
 
-            String failoverUrl = String.format("failover:(%s)?maxReconnectAttempts=10&initialReconnectDelay=100", transportConnector.getConnectUri());
+            String failoverUrl = String.format("failover:(%s)?startupMaxReconnectAttempts=10&maxReconnectAttempts=2&initialReconnectDelay=100&timeout=2000", transportConnector.getConnectUri());
```
jbonofre, line 140: I saw this one. Also, something that I observed is that the 60s timeout on the test could be not enough. It seems that increasing the timeout (120s) makes the change.
I guess that's the purpose of adding `timeout=2000` on the connector here.

jeanouii: Yes, 2 things here
- maxReconnectAttempts isn't used the first time, so we may well timeout even without starting
- Adding startupMaxReconnectAttempt make it fails faster it has to
- timeout=2000 is another way to approach it yes. Increasing to 120s would work, but does not go in the direction we are trying to go (decrease build time if possible)

jbonofre: Totally agree, that's also what I have observed.

jbonofre:
Still failed but after a run of 2:43 😀

PR #1626, opened 2026-01-22
Configure broker to check for expired messages every second in tests

jeanouii:
Very simple test to valid expiration is serialized with HTTP protocol.
The test simply sends a message with TTL 100ms and waits 30s for the message to show up in the DLQ.
Unfortunately the broker side check happens every 30s, so on fast nodes, the receive returns before the broker actually sends the message to the DLQ.
Adding a policy to speed up the move to the DLQ. The test should also run way faster.

### DIFF
--- a/activemq-http/src/test/java/org/apache/activemq/bugs/AMQ9255Test.java
+++ b/activemq-http/src/test/java/org/apache/activemq/bugs/AMQ9255Test.java
@@ -25,6 +25,8 @@ import jakarta.jms.TextMessage;
 
 import org.apache.activemq.ActiveMQConnectionFactory;
 import org.apache.activemq.broker.BrokerService;
+import org.apache.activemq.broker.region.policy.PolicyEntry;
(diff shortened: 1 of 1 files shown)


jbonofre:
Yes, that's clearly planned for ActiveMQ 7. We should use a configuration aggregator that clearly list and document all configurations.
Also the default configuration should be probably reviewed (again for AMQ 7 as it's potentially breaking change for users).

jbonofre (approved):

jeanouii:
I agree that Spring does not bring much to us. But we have to deal with dependency updates, cve, etc. 

Having a broker that we can build with a builder pattern, like it's been done on other projects I've been involved in, would be neat and helpful.

jbonofre:
@jeanouii that's the PoC I started 😄

jeanouii:
Loolks like it's green!

jbonofre:
That's super nice !

PR #1628, opened 2026-01-22
Refactor JMSConsumerTest and ZeroPrefetchConsumerTest

jeanouii:
Eliminate Thread.sleep and improve message assertion logic

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/bugs/MKahaDBTxRecoveryTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/bugs/MKahaDBTxRecoveryTest.java
@@ -52,6 +52,7 @@ import jakarta.jms.MessageProducer;
 import jakarta.jms.Session;
 import javax.management.ObjectName;
 import java.io.IOException;
+import java.io.File;
 import java.io.RandomAccessFile;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -435,7 +436,23 @@ public class MKahaDBTxRecoveryTest {
             assertTrue("broker/store found corruption", foundSomeCorruption.get());
             assertTrue("broker/store ignored corruption", ignoringCorruption.get());
 
+            // effectively wait for the async process to clean up after corrupt detection
+            final File txStoreDir = new File(pathToDataDir, "mKahaDB/txStore");
+            assertTrue("txStore cleanup", Wait.waitFor(() -> {
+                File[] files = txStoreDir.listFiles((dir, name) -> name.endsWith(".log"));
+                if (files == null || files.length == 0) {
+                    return false;
+                }
+                for (File file : files) {
+                    if ("db-1.log".equals(file.getName())) {
+                        return false;
+                    }
+                }
+                return true;
+            }, TimeUnit.SECONDS.toMillis(5), 100));
+
             broker.stop();
+            broker.waitUntilStopped();
 
             foundSomeCorruption.set(false);
             ignoringCorruption.set(false);


activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
```diff
@@ -1019,15 +1016,22 @@ public void testAckOfExpired() throws Exception {
...
         final int count = 4;
         for (int i = 0; i < count; i++) {
             TextMessage message = sendSession.createTextMessage("" + i);
             producer.send(message);
         }
 
         // let first bunch in queue expire
-        Thread.sleep(1000);
+        final long expiryTime = System.currentTimeMillis() + TimeUnit.MILLISECONDS.toMillis(ttl * 2L);
+        assertTrue("ttl elapsed", Wait.waitFor(new Wait.Condition() {
```
cshannon: I'm not sure I see the point of using the Wat.waitFor() here. Isn't this just doing the same thing as if you called Thread.sleep(expiryTime)?

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
```diff
@@ -1066,4 +1070,13 @@ protected DestinationViewMBean createView(ActiveMQDestination destination) throw
         }
         return (DestinationViewMBean) broker.getManagementContext().newProxyInstance(name, DestinationViewMBean.class, true);
     }
+
+    private void assertNoAdditionalMessages(final AtomicInteger counter, final int expected) throws Exception {
```
cshannon, line 1057: Nit: I think it's a little bit nicer to just use a lambda instead of an anonymous method (this would apply the other uses of Wait as well). It's not strictly necessary of course but just a style thing. For example:

```java
private void assertNoAdditionalMessages(final AtomicInteger counter, final int expected) throws Exception {
    assertFalse("unexpected additional messages received", Wait.waitFor(
        () -> counter.get() > expected, TimeUnit.SECONDS.toMillis(2), 50));
}
```

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
jeanouii: Hum, I orgot something looks like.
I think the goal of the WaitFor was to actually check the value of expired messages for the destination. In essence, instead of relying on elapsed time from the client perspective, actually wait for the broker to process the expired messages.

Condition would be like 

`...getDestinationStatistics().getExpired().getCount() >= count`

Does that work for you?

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
jeanouii: From a code style perspective, I prefer because it's more readable. I'm happy to use modern additions of the language.

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
jbonofre: I think `Wait.waitFor()` with a condition (and timeout) is more efficient/secure compared to `Thread.sleep()`: an arbitrary `Thread.sleep()` depends of the performance of the machine running the test, whereas `Wait.waitFor()` is condition "adapting" to the executor performance.
I think we should use a "strong" condition here (like the expired count).

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
jbonofre: I don't have a strong opinion here, both works for me.

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
jeanouii: Actually this is not even needed. The waitfor is only needed at the statistics level bellow (asserts) because before that, the sleep is perfectly valid; the expiration is only time based.
Thanks for catching it up and the review @cshannon

jbonofre: Yes, I just see that with a closer review. My comment was more "generally speaking".

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
jeanouii: I prefer new code style and lambda for this usage. It's a bit more concise and readable. But the code base does not massively use new language features, so I wasn't sure it was good or not to start introducing myself.

jeanouii:
Still green and looks better now.

jbonofre:
It looks good to me.

@cshannon thoughts ?

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
cshannon: I just noticed that something seemed off because the wait for was essentially just looping until the sleep time finished and wasn't waiting for a real condition to change (just time elapsing) so i am glad it is sorted

activemq-unit-tests/src/test/java/org/apache/activemq/JMSConsumerTest.java
cshannon: That's fine, you don't need to change it. It's just a stylistic thing and doesn't matter much. Intellij just likes to point out when you can collapse those methods which is why I noticed.

cshannon: > I prefer new code style and lambda for this usage. It's a bit more concise and readable. But the code base does not massively use new language features, so I wasn't sure it was good or not to start introducing myself.

Ah oops I missed this and saw you changed it so that is good.

Yeah much of the code is quite old so there is not a lot of modern stuff but I think it's fine to introduce it. I figure it's not like we are going to go through and do it all at once so it's fine to mix.

jbonofre:
@cshannon great. Thanks !

PR #1629, opened 2026-01-22
[AMQ-9830] Replaces commons-io usage in activemq-jaas with Java nio

mattrpav:


activemq-jaas/src/test/java/org/apache/activemq/jaas/PropertiesLoginModuleTest.java
```diff
@@ -92,10 +94,10 @@ public void testLoginReload() throws Exception {
 
         //Modify the file and test that the properties are reloaded
         Thread.sleep(1000);
-        FileUtils.copyFile(new File(sourcePropDir, "usersReload.properties"), usersFile);
-        FileUtils.copyFile(new File(sourcePropDir, "groupsReload.properties"), groupsFile);
-        FileUtils.touch(usersFile);
-        FileUtils.touch(groupsFile);
+        Files.copy(new File(sourcePropDir, "usersReload.properties").toPath(), usersFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+        Files.copy(new File(sourcePropDir, "groupsReload.properties").toPath(), groupsFile.toPath(), StandardCopyOption.REPLACE_EXISTING);
+        Files.setLastModifiedTime(usersFile.toPath(), FileTime.from(Instant.now()));
```
cshannon, line 100: This doesn't seem necessary...wouldn't the Files.copy() calls above update the last modified time already? I realize this is replacing the FileUtils.touch() call but that seems unnecessary as well unless I am missing something.

jeanouii: Attributes are not guaranteed to be copied actually, unless you add COPY_ATTRIBUTES option at the end. This is JVM specific so you might get the source file's timestamp or the current time of the copy. 
So strictly speaking, to mimic FileUtiles.touch(), setting last modified date makes sense here to trigger reloading for instance.

jeanouii:
@mattrpav This was the only use of Commons IO?
There are regular CVE affecting it, so if we can get rid of it, even in tests, that'd be great!

mattrpav:
@jeanouii Agreed =) commons-io was already removed from the runtime.

I've been regularly removing dependencies that provide small value. Most of the commons* can be replaced by JDK native functions at this point.

jbonofre:
That's a good one, aligned with similar change we did in the past (in the webconsole). I'm gonna take a look.

PR #1631, opened 2026-01-26
Add VM transport broker restart test

pradeep85841:
This PR adds a unit test to validate VM transport behaviour when the broker
is stopped and restarted.

The test ensures:
- Messages can be sent and received before broker shutdown
- A restarted broker correctly accepts new VM transport connections
- No stale broker registry state affects message delivery

Test passes when run in IDE and follows existing ActiveMQ unit test patterns.

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/VmTransportBrokerRestartTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/VmTransportBrokerRestartTest.java
@@ -1,0 +1,82 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq;
+
+import jakarta.jms.TextMessage;
+import org.apache.activemq.broker.BrokerService;
+import org.apache.activemq.test.annotations.ParallelTest;
+import org.junit.After;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+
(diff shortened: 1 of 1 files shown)


activemq-unit-tests/src/test/java/org/apache/activemq/VmTransportBrokerRestartTest.java
```diff
@@ -0,0 +1,77 @@
...
+                (TextMessage) template.receive(destination);
+
+        assertNotNull(receivedBefore);
+        assertEquals(firstMessage, receivedBefore.getText());
+
+        broker.stop();
+        broker.waitUntilStopped();
+
+        broker = createBroker();
+        startBroker();
```
jeanouii, line 57: Shouldn't you do a waitUntilStarted() after startBroker()?
I've seen a few flaky tests falling on CI because if race condition issues.

activemq-unit-tests/src/test/java/org/apache/activemq/VmTransportBrokerRestartTest.java
```diff
@@ -0,0 +1,77 @@
...
+ * limitations under the License.
+ */
+package org.apache.activemq;
+
+import jakarta.jms.TextMessage;
+import org.apache.activemq.broker.BrokerService;
+import org.junit.After;
+
+
+public class VmTransportBrokerRestartTest extends EmbeddedBrokerTestSupport {
```
jeanouii, line 27: I think this test is safe to run in parallel. 
Can you add `@Category(ParallelTest.class)` to it please?

pradeep85841: Thanks for the feedback!

I’ve added broker.waitUntilStarted() after restarting the broker to avoid any race conditions.

The test was added based on issues observed around VM transport behavior after broker restarts, and is intended to guard against regressions in that area.

I’ve also added @category(ParallelTest.class) since the test uses an isolated embedded broker and should be safe to run in parallel.

I kept JmsTemplate for consistency with existing tests, but agree that moving to plain JMS would make sense in a follow-up if needed.

activemq-unit-tests/src/test/java/org/apache/activemq/VmTransportBrokerRestartTest.java
jbonofre: `waitUntilStarted()` is there now.

activemq-unit-tests/src/test/java/org/apache/activemq/VmTransportBrokerRestartTest.java
jbonofre: Thanks ! It looks good now.

jbonofre:
I just ran the CI. I will merge as soon as green.

PR #1635, opened 2026-01-28
AMQ-9843 Handle InstanceNotFoundException during MBean unregistration…

jeanouii:
… to avoid errors on shutdown


jbonofre:
I checked and we have a test for the ManagementContext checking the MBean unregistration flow. It looks good to me. Thanks.

jbonofre (approved):

PR #1637, opened 2026-01-28
Enable publishing of test results in CI workflows

jeanouii:
I followed the documentation at https://github.com/marketplace/actions/publish-test-results but I'm not sure how it will work on our repo. Let's see

Here is the result https://github.com/apache/activemq/actions/runs/21436334968
Not too happy about it. I wanted to see individual test failures with output, but probably some tuning to do.


jeanouii:
Unfortunate that of course this PR passed so we can't see what would be the result after the small tuning ...

jbonofre:
I was about to do that. This approach works for me (an alternative approach would be using dony test reporter).

PR #1641, opened 2026-01-30
AMQ-9852 Ensure advisory consumer receives producer advisory before creation to prevent race conditions

jeanouii:
This may have performance impact because the firing is not negligable. So I need to make sure it's important

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/advisory/AdvisoryBroker.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/advisory/AdvisoryBroker.java
@@ -367,6 +367,8 @@ public class AdvisoryBroker extends BrokerFilter {
             } finally {
                 consumersLock.writeLock().unlock();
             }
+            // Fire advisory outside the lock to avoid potential deadlocks
+            // (fireConsumerAdvisory calls next.send() which can acquire other locks)
             if (!dest.isTemporary() || destinations.containsKey(dest)) {
                 fireConsumerAdvisory(context, dest, topic, info.createRemoveCommand());
             }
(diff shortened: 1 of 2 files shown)


cshannon:
I looked at the test and I think I know what is going on but still would need to do some more work to prove it.

The test sends the consumer info for the second consumer (consumer on real topic) to generate the advisory, and then immediately sends the consumer info for the first consumer (consumer on the advisory topic) so it can receive the advisory and get a replay. I believe the issue is that the advisory consumer may receive a duplicate of the consumer coming online because it is getting both the "live" advisory and also a "replay" of the advisories because the second consumer is still in flight and being processed as coming online at the same time the advisor consumer is coming online. I'm not quite sure of the order but it seems like that is what is going on where it's just getting 2 copies as the logic for processing the new advisory consumer hasn't quite finished.

When I run the test in a loop in Intellij it usually fails after a few thousand times. If I put in a 10 millisecond sleep immediately after [line 155](https://github.com/apache/activemq/blob/61a0dd846eba9eb2c922fd50c75a9aee6737503a/activemq-unit-tests/src/test/java/org/apache/activemq/broker/advisory/AdvisoryBrokerTest.java#L155) to give the real consumer time to fully come online (which is the goal, so we can test replay) before bringing the advisory consumer online then the test no longer fails.

I'm not sure there's much to do here to fix it other than fix the test, it would be hard to prevent issues in all cases without deadlocks. A duplicate consumer advisory isn't really a big deal as the consumer id is the same so the advisory just gets notified twice but it's the same id so I would expect processes listening to the advisories (like network bridges) to be able to handle it without issue.

jeanouii:
@cshannon Yeah, I don't want to introduce a deadlock. 
I ended up fixing the test and waiting for the producerInfo to be fully processed by the broker.

Does it sound good to you?

cshannon:
@jeanouii - can you update the jira description and the title to be accurate with this now just fixing the test vs changing anything with the AdvisoryBroker and locking?

jeanouii:
Should be done @cshannon

PR #1643, opened 2026-01-30
AMQ-9853 Ensure consumer is closed after all acknowledgments are processed

jeanouii:
ACKs are Async and Closing Consumer Too Early Skips Them. The broker's queue statistics (dequeueCount) are the only reliable synchronization point to confirm ACKs were processed


cshannon (approved):

cshannon:
Is there an associated test Jira with this? The test failures seem unrelated of course so I think it can be merged

jeanouii:
@cshannon Issue created [AMQ-9853](https://issues.apache.org/jira/browse/AMQ-9853)
Commit updated and PR title

gemmellr:
This seems more like a potential client bug the test was catching but no longer will. Calling close from another thread is meant to wait for an in progress onMessage delivery to complete, so there doesnt appear to be a legitimate race in the original test code calling close after the latch was tripped due to receiving the final message.

PR #1645, opened 2026-01-31
Add Jolokia-based healthcheck for ActiveMQ containerFeature/activemq jolokia healthcheck

pradeep85841:
This PR adds a Docker HEALTHCHECK for the ActiveMQ container using Jolokia.

Key points:
- Uses authenticated Jolokia search to verify Broker MBean is available.
- Ensures broker is fully initialized before container is marked healthy.
- Compatible with any ActiveMQ 6.x brokerName.
- Avoids fragile attribute checks like 'Started'.
- Origin header included to satisfy Jolokia security.
- Works with CI pipelines and deterministic Docker health status.

Reviewer notes:
- ENTRYPOINT and CMD unchanged.
- No changes to shutdown/logging behaviour.
- Only adds healthcheck to improve container observability.


jbonofre:
If the healthcheck in Dockerfile is interesting, I'm not sure Jolokia is the best option (or maybe just temporary).

PR #1647, opened 2026-01-31
Upgrade Netty to 4.2.9.Final

hyperxpro:
Motivation:
Netty 4.2 supersedes 4.1, making an upgrade advisable to stay current and supported.

Modification:
1. Upgrade Dependency to 4.2.9.Final
2. Use `MultiThreadIoEventLoopGroup` instead of `NioEventLoopGroup` directly.

Result:
Latest stable release of Netty


jbonofre:
This is tackle by dependabot and dependabot should be updated if needed.

PR #1649, opened 2026-02-03
Refactor AMQ3529v2Test to improve thread handling and logging consistency

jeanouii:
We strictly enforce client threads to be finished.
The Transport thread assertion is relaxed and logged as WARN for the moment after giving it more time to complete.

Transport cleanup can be a bit delayed. We might want to still assert the transport threads to avoid any regression.

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/bugs/AMQ3529v2Test.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/bugs/AMQ3529v2Test.java
@@ -177,18 +177,15 @@ public class AMQ3529v2Test {
             thread.start();
         }
         // interrupt the threads at some random time
-        ExecutorService doTheInterrupts = Executors.newFixedThreadPool(threads.size());
+        final ExecutorService doTheInterrupts = Executors.newFixedThreadPool(threads.size());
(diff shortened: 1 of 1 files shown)


activemq-unit-tests/src/test/java/org/apache/activemq/bugs/AMQ3529v2Test.java
```diff
@@ -200,21 +197,43 @@ public void run() {
...
-            if (t != null && t.isAlive() && !t.isDaemon())
-                assertTrue("Thread completes:" + t, Wait.waitFor(new Wait.Condition() {
-                    @Override
-                    public boolean isSatisified() throws Exception {
-                        LOG.info("Remaining thread: " + t.toString());
-                        return !t.isAlive();
+            if (t == null || !t.isAlive() || t.isDaemon()) {
+                continue;
+            }
+            if (t.getName().startsWith("ActiveMQ Transport")) {
```
jbonofre, line 211: Let me check if it's good enough for transport.

jbonofre: After checking, the thread name is consistent on the transport. So this statement is accurate.

PR #1650, opened 2026-02-04
[#] Ignore JournalFdRecoveryTest due to flakiness

jeanouii:
While we are working on AMQ-9851 in PR #1644 
Let's turn it off so we can have nightly builds


cshannon (approved):
It seems fine to me to turn it off until it is fixed. We could disable it for just MacOS but that might be a pain with maven and if we are going to fix it anyways then just temporarily disabling altogether is probably fine.

PR #1661, opened 2026-02-10
[AMQ-9857] Add SSL support for JMX connector in ManagementContext

jeanouii:
Thoughts on reusing the SSL Context to configure the SSL connector for JMX to avoid duplication of configuration and allow Jasypt to cipher passwords.

### DIFF
--- a/activemq-broker/src/test/java/org/apache/activemq/broker/jmx/ManagementContextSslTest.java
+++ b/activemq-broker/src/test/java/org/apache/activemq/broker/jmx/ManagementContextSslTest.java
@@ -1,0 +1,285 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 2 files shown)


activemq-broker/src/test/java/org/apache/activemq/broker/jmx/ManagementContextSslTest.java
```diff
@@ -192,6 +195,9 @@ public void testConnectorStartsWithSsl() throws Exception {
         SSLContext.setDefault(testSslContext);
         System.setProperty("javax.net.ssl.trustStore", keystoreFile.toString());
         System.setProperty("javax.net.ssl.trustStorePassword", KEYSTORE_PASSWORD);
+        // Force RMI stubs to advertise "localhost" so SSL hostname verification
+        // matches the certificate SAN (dns:localhost) instead of the machine's IP.
+        System.setProperty("java.rmi.server.hostname", "localhost");
```
mattrpav: This can be localized into the RMIServer environment map vs only supporting it via the System Property

jeanouii: Should be better now

jbonofre (commented):
LGTM, thanks.

I would suggest to update the documentation about this 😄

jeanouii:
Good point @jbonofre 

Here it is
https://github.com/apache/activemq-website/pull/173

jbonofre:
This one is worth to consider for 6.4.0.

mattrpav:
@jbonofre this needs to be re-worked to move off of the ThreadLocal. 

I think we should merge the pending SSL refactor [#2133] (removes ThreadLocal, support per-connector SSLContext) first and then revisit this one.

cshannon:
> I think we should merge the pending SSL refactor [#2133] (removes ThreadLocal, support per-connector SSLContext) first and then revisit this one.

I agree we should get #2133 merged first and then rework this with any needed changes

mattrpav:
The SSL refactor is complete and merged. This PR needs to be rebased and then re-rationalized to leverage a static specified SSLContext and fallback to the broker-wide context.

jbonofre:
Yes, that makes sense. Let me ping @jeanouii or I will do the rebase/change myself.

This PR should be included in 6.4.0.

PR #1662, opened 2026-02-11
[#1842] [AMQ-9858] Support purging the first number of messages from a queue

mattrpav:
ref: [#1842] 
ref: [AMQ-9858]

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/QueueViewMBean.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/QueueViewMBean.java
@@ -71,6 +71,14 @@ public interface QueueViewMBean extends DestinationViewMBean {
     void purge() throws Exception;
 
     /**
+     * Removes the first number of messages in the queue.
+     *
+     * @throws Exception
+     */
+    @MBeanInfo("Removes the first number of messages in the queue.")
+    void purge(long numberOfMessages) throws Exception;
+
+    /**
      * Copies a given message to another destination.
      * 
      * @param messageId
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
@@ -1320,9 +1320,20 @@ public class Queue extends BaseDestination implements Task, UsageListener, Index
     }
 
     public void purge() throws Exception {
+        purge(this.destinationStatistics.getMessages().getCount());
+    }
+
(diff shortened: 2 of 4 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/QueueViewMBean.java
```diff
@@ -70,6 +70,14 @@ public interface QueueViewMBean extends DestinationViewMBean {
     @MBeanInfo("Removes all of the messages in the queue.")
     void purge() throws Exception;
 
+    /**
+     * Removes the first number of messages in the queue.
+     *
+     * @throws Exception
+     */
+    @MBeanInfo("Removes the first number of messages in the queue.")
+    void purge(long number) throws Exception;
```
jbonofre: nit: for the reader, we could have use `numberOfMessages` here (to be consistent with `QueueView` interface).

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -1313,9 +1313,15 @@ public QueueMessageReference getMessage(String id) {
     }
 
     public void purge() throws Exception {
+        purge(this.destinationStatistics.getMessages().getCount());
+    }
+
+    public void purge(long numberOfMessages) throws Exception {
         ConnectionContext c = createConnectionContext();
```
jbonofre, line 1332: We should add `numberOfMessages` validation here:


```suggestion
        if (numberOfMessages <= 0) {
            return;
        }
        ConnectionContext c = createConnectionContext();
```

Else, it means that this method will acquire `sendLock`, create `ConnectionContext` and go in the loop even if the number of messages to purge is invalid.

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -1313,9 +1313,15 @@ public QueueMessageReference getMessage(String id) {
...
         List<MessageReference> list = null;
         sendLock.lock();
+
+        long purgeCount = 0L;
         try {
             long originalMessageCount = this.destinationStatistics.getMessages().getCount();
```
jbonofre, line 1338: With the change, `originalMessageCount` is not used anymore, it can be removed.

mattrpav: I now use the originalMessageCount in a log message further down. 

(see latest updated commit)

PR #1664, opened 2026-02-11
AMQ-9812: Allow user to set ACTIVEMQ_OPTS and ACTIVEMQ_OPTS_MEMORY env variables in Docker image

jbonofre:


assembly/src/docker/entrypoint.sh
```diff
@@ -78,4 +78,15 @@ if [ -n "${ACTIVEMQ_WEB_USER}" ]; then
   fi
 fi
 
+if [ -z "${ACTIVEMQ_OPTS_MEMORY}" ]; then
+  ACTIVEMQ_OPTS_MEMORY="-Xms64M -Xmx1G"
+  export ACTIVEMQ_OPTS_MEMORY
+fi
+
+if [ -z "${ACTIVEMQ_OPTS}" ]; then
+  # Make the Web console accesible from outside the container if not already set
```
jeanouii, line 87: Minor type `accessible`

PR #1665, opened 2026-02-11
AMQ-9588: Run ActiveMQ with regular user in Docker container, and hook to easily stop the container

jbonofre:


assembly/src/docker/entrypoint.sh
```diff
@@ -18,6 +18,8 @@
 # limitations under the License.
 ################################################################################
 
+set -e
```
jeanouii: Is it expected?

Any sed command bellow failing will fail silently. Not sure if you wanted that or not

assembly/src/docker/entrypoint.sh
```diff
@@ -89,4 +91,15 @@ if [ -z "${ACTIVEMQ_OPTS}" ]; then
...
+  if [ -n "${child_pid:-}" ] && kill -0 "${child_pid}" 2>/dev/null; then
+    kill -TERM "${child_pid}" 2>/dev/null || true
+  fi
+}
+
+trap _term TERM INT
+
+"$@" &
+child_pid=$!
+wait "${child_pid}"
```
jeanouii, line 103: exit missing right?

`exit $?`

assembly/src/docker/entrypoint.sh
jbonofre: OK to remove the "debug" output.

jbonofre:
@jeanouii good catches, I updated the PR.

PR #1666, opened 2026-02-11
[#1733] Fix race condition in async exception handling by combining JMS ExceptionListener and TransportListener notifications

### LINKED ISSUE #1733: Fix race condition in async exception handling by combining JMS ExceptionListener and TransportListener notifications

jeanouii:
I'm not quite sure if it's possible or not, but looks like there is a race condition where the transport failed listener is invoked before the exception listener. The test requires the exception listener to be invoked first so they need to be invoked sequentially withing the same runnable.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java
@@ -2029,25 +2029,45 @@ public class ActiveMQConnection implements Connection, TopicConnection, QueueCon
      * @param error
      */
     public void onAsyncException(Throwable error) {
-        if (!closed.get() && !closing.get()) {
-            if (this.exceptionListener != null) {
-
-                if (!(error instanceof JMSException)) {
-                    error = JMSExceptionSupport.create(error);
-                }
-                final JMSException e = (JMSException) error;
-                executeAsync(() -> exceptionListener.onException(e));
-
-            } else {
-                LOG.debug("Async exception with no exception listener: {}", error, error);
+        if (this.exceptionListener != null) {
+            if (!(error instanceof JMSException)) {
+                error = JMSExceptionSupport.create(error);
             }
+            final JMSException e = (JMSException) error;
+            // Submit directly to executor bypassing closed/closing guards
+            // to ensure exception notifications are never silently dropped
+            try {
+                executor.execute(() -> exceptionListener.onException(e));
+            } catch (final RejectedExecutionException re) {
+                LOG.debug("Could not notify exception listener asynchronously (executor terminated), notifying inline: {}", error.getMessage());
+                try {
+                    exceptionListener.onException(e);
+                } catch (final Exception ex) {
+                    LOG.debug("Exception during inline ExceptionListener notification", ex);
+                }
+            }
+        } else {
+            LOG.debug("Async exception with no exception listener: {}", error, error);
         }
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


cshannon:
i haven't had a chance to peak at this yet but i will try and look soon

jeanouii:
@cshannon thank you. This one is tricky as well, so needs some more pair of eyes.

activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java
```diff
@@ -2046,20 +2048,40 @@ public void onAsyncException(Throwable error) {
 
     @Override
     public void onException(final IOException error) {
-        onAsyncException(error);
-        executeAsync(() -> {
-            transportFailed(error);
-            ServiceSupport.dispose(ActiveMQConnection.this.transport);
-            brokerInfoReceived.countDown();
+        if (!closed.get() && !closing.get()) {
```
jbonofre, line 2051: `onException()` doesn't call `onAsyncException()` anymore.

With the exception listener "inline" here, it bypass the `onAsyncException()` completely. I'm not sure we have anyone overridden `onAsyncException()`, but it has an impact.

jeanouii: This is the purpose of the change. 
We can have the 2 listeners executed in their own thread, because we can't guarantee the order and we can't guarantee that the exception listener will be triggered before the transport failed.
The only to guarantee the order is to run the 2 in the same async thread.

jbonofre: Yes, I got it, I'm just curious about the impact. If nobody overriddes `onAsyncException()` that's all good to me.

cshannon (commented):
Unless I am missing something I don't think there is actually a race condition here because the executor is [single threaded ](https://github.com/jeanouii/activemq/blob/a5b9308b38b85b64840a6ea27747ca8c08364c0a/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L235) so it will execute tasks in order. onException(error) first calls onAsyncException(error) which will submit the task to the executor to process any configured error listener. Another task is then submitted after to execute the transport failed task.

jeanouii:
> Unless I am missing something I don't think there is actually a race condition here because the executor is [single threaded ](https://github.com/jeanouii/activemq/blob/a5b9308b38b85b64840a6ea27747ca8c08364c0a/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L235) so it will execute tasks in order. onException(error) first calls onAsyncException(error) which will submit the task to the executor to process any configured error listener. Another task is then submitted after to execute the transport failed task.

You are correct @cshannon This is a single threaded (FIFO).
Now the race condition happens because the exception listener may result in a connection.close() before the second task about the transport failure is submitted, resulting in a RejectedExecutionException.

Hopefully this is correct

cshannon:
> i haven't had a chance to peak at this yet but i will try and look soon

2 months later and still haven't looked :smile: We got slammed with tons of security fixes so that's been the focus. We are trying to get 6.3.0 out and other releases going this week so I am going to move this to 6.4.0 for now. We could also fix it for 6.3.1 or another bug fix release too.

jbonofre:
@cshannon I'm doing a new pass as well. 6.4.0 and/or 6.3.1 make sense.

PR #1668, opened 2026-02-11
Use destination statistics instead of global statistics to avoid side…

jeanouii:
… effects in asserts

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/usecases/AdvisoryViaNetworkTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/usecases/AdvisoryViaNetworkTest.java
@@ -22,7 +22,9 @@ import java.util.Arrays;
 import jakarta.jms.MessageConsumer;
 
 import org.apache.activemq.JmsMultipleBrokersTestSupport;
+import org.apache.activemq.advisory.AdvisorySupport;
 import org.apache.activemq.broker.BrokerService;
+import org.apache.activemq.broker.region.Destination;
 import org.apache.activemq.broker.region.DestinationInterceptor;
 import org.apache.activemq.broker.region.RegionBroker;
 import org.apache.activemq.broker.region.virtual.CompositeTopic;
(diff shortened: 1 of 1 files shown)


activemq-unit-tests/src/test/java/org/apache/activemq/usecases/AdvisoryViaNetworkTest.java
```diff
@@ -270,18 +272,27 @@ public void testPrefetchSizePercent() throws Exception {
...
             public boolean isSatisified() throws Exception {
-                RegionBroker regionBroker = (RegionBroker) brokers.get("A").broker.getRegionBroker();
-                LOG.info("A Deq:" + regionBroker.getDestinationStatistics().getDequeues().getCount());
-                LOG.info("A Inflight:" + regionBroker.getDestinationStatistics().getInflight().getCount());
-                return regionBroker.getDestinationStatistics().getDequeues().getCount() == dequeue
-                        && regionBroker.getDestinationStatistics().getInflight().getCount() == inflight;
+                long actualDeq = 0;
+                long actualInflight = 0;
+                for (ActiveMQTopic topic : topics) {
+                    ActiveMQTopic advisory = AdvisorySupport.getConsumerAdvisoryTopic(topic);
```
jbonofre, line 286: Just for context: what we are testing here is not the global `DestinationStatistics`.
We want to test that the consuming actually increase the `dequeue` (after consuming completed) and the `inflight` (when consuming is in progress).

So, it's fine to use that for a specific destination (instead of global).

However, maybe it makes sense to try to keep also the test with global but "isolated" to avoid side effect. Just my $0.01.

jeanouii: The point is that sometimes it's failing because on the global statistics we get advisory (nob statistics). So because the test is focused into testing statistics for one destination only, I changed it to match more closely what's the test is about.

jbonofre: That's fine for me. I just wanted to provide some background.

jbonofre (approved):

PR #1669, opened 2026-02-11
Improve duplex network wait conditions and ports

jeanouii:


cshannon:
This looks pretty good but same comment as #1674 , maybe it's easier to just get rid of all the spring config for the tests entirely?

cshannon:
It looks like SimpleNetworkTest failed, is that something related?

jeanouii:
I've updated the PR but it fails in a different test everytime. So I'm trying again. I don't think it was related, but I'll see if I can make it green for more confidence.

jeanouii:
@cshannon It's green now, can you take a look please?

cshannon (approved):

cshannon:
LGTM now that tests pass

PR #1674, opened 2026-02-12
Support ephemeral ports in network config and Spring

jeanouii:


cshannon:
Would it be easier to just drop the XML config altogether and just configure those brokers without using Spring? It's a Base class so it may cause a lot of cascading refactoring so not sure it's a good idea.

jeanouii:
I agree and wanted to do that, but there were some discussions to drop Spring, so I figured we would do that at the same time.

If you prefer to start now and mix the ephemeral port with the XML file removal, I can give it a go and see what's the impact

cshannon:
I think it's fine to drop spring separately, it's probably better to do that anyways as the changes are a bit unrelated.

jbonofre:
@cshannon assigning PR to milestone is good enough for the release notes. Issue is welcome for "significant" changes.

cshannon:
@jbonofre and @jeanouii - before you merge I think the test failures might be real. I initially thought maybe they were flaky but I looked again at eh results and noticed some failures like `DynamicallyIncludedDestinationsDuplexNetworkTest` that pass on main for me but not this branch

jeanouii:
I investigated a bit deeper and did a small tweak. Thank you so much for the deep analysis.
I think they are now all resolved and should be back to normal. I ran them all in a loop locally and they never failed.

The last run reported `QueueZeroPrefetchLazyDispatchPriorityTest` as failure, but it's a known flaky test that I need to address in another PR.

jeanouii:
I ran it again, and it failed again after in the activemq-http module with 

```
Error:  Failures: 
Error:    AMQ2764Test.testBrokerRestart:145 Had extra messages expected null, but was:<ActiveMQMessage {commandId = 30, responseRequired = true, messageId = ID:runnervmjduv7-46345-1771502273501-24:50:1:1:1, originalDestination = null, originalTransactionId = null, producerId = broker1->broker2-46345-1771502273501-27:2:1:1, destination = queue://RECONNECT.TEST.QUEUE, transactionId = null, deliveryTime = 1771502281971, expiration = 0, timestamp = 1771502281971, arrival = 0, brokerInTime = 1771502281972, brokerOutTime = 1771502281972, correlationId = null, replyTo = null, persistent = true, type = null, priority = 4, groupID = null, groupSequence = 0, targetConsumerId = null, compressed = false, userID = null, content = null, marshalledProperties = null, dataStructure = null, redeliveryCounter = 0, size = 1024, properties = null, readOnlyProperties = true, readOnlyBody = true, droppable = false, jmsXGroupFirstForConsumer = false}>
[INFO] 
Error:  Tests run: 535, Failures: 1, Errors: 0, Skipped: 4

```

I think we are good here, unless you find a link @cshannon 
This looks like a flaky test we need to address. Looks like after reconnect we received an extra and unexpected message.  There is a thread sleep after stop instead of waitUntilStopped() or a WaitFor approach to avoid issues on slow or fast CIs

PR #1678, opened 2026-02-12
Introduce weekly CI build for comprehensive testing, and limit multi …

jeanouii:
…os/jvm to only the quick profile


jeanouii:
This is not critical so I'll move forward and merge

PR #1680, opened 2026-02-12
Fix copy-paste bug in MessageGroupLateArrivalsTest assertion

jeanouii:
testConsumerLateToBigPartyGetsNewGroup had two errors on line 220:
   1. Checked messageCount.get("worker1") instead of "worker2"
   2. Expected 2*perBatch (4) instead of perBatch (2)


jbonofre (approved):
Good catch !

PR #1682, opened 2026-02-13
Update GitHub configuration

jbonofre:
Update GitHub configuration with:
- allow merge and rebase on PR
- need at least one review before merging

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -32,8 +32,8 @@ github:
 
   enabled_merge_buttons:
     squash: true
-    merge: false
-    rebase: false
+    merge: true
+    rebase: true
 
   del_branch_on_merge: true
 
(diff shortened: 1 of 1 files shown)


.asf.yaml
```diff
@@ -32,16 +32,16 @@ github:
...
+    rebase: true
 
   del_branch_on_merge: true
 
   protected_branches:
     main:
       required_pull_request_reviews:
         require_code_owner_reviews: false
-        required_approving_review_count: 0
+        required_approving_review_count: 1
```
mattrpav, line 44: +1 -- agree we should have this for supply chain traceability.

.asf.yaml
```diff
@@ -32,16 +32,16 @@ github:
 
   enabled_merge_buttons:
     squash: true
-    merge: false
-    rebase: false
+    merge: true
+    rebase: true
```
mattrpav, line 36: I think this is ok for now. As a community, we should make a goal by the end of the year to sign all commits for software supply chain verification (which means all merges would need to be done by CLI and not GH web page buttons)

cshannon (approved):

PR #1687, opened 2026-02-17
Add parallel test annotations to multiple test classes

jeanouii:
This PR only adds the annotation to some test classes eligible to run in the parallel surefire execution. 
In order to avoid going through hundred of files, I have NOT included any test fix or production code changes. 

I understand it's a lot of files, but really it's only the annotation added to a subset of classes, so if it becomes green at some point, we can merge I think


jeanouii:
When #1711 is merged I think this should be green or close to.

PR #1688, opened 2026-02-17
Enable GitHub Issues and Discussions

jbonofre:
Following the discussion/vote on the dev mailing list, this PR enables GitHub Issues and Discussions.

It also updates the notification schema accordingly and improve the branch protection policy.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -41,13 +41,16 @@ github:
     main:
       required_pull_request_reviews:
         require_code_owner_reviews: false
+        dismiss_stale_reviews: true
+        require_last_push_approval: false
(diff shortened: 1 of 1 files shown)


cshannon:
What happens with Jira, do we need to get Infra to mark that as read only? I guess we may need to migrate some of the Jiras to GH Issues

jbonofre:
@cshannon yes, I will turn Jira as read-only as soon as this PR is merged.
Then, I will start the "migration". I don't we should to just a migration because there is a lot of Jira that don't make sense. I propose to move the Jira that make sense (we can do that using a label for instance or manually).
I will start this process, any help is welcome 😄

PR #1691, opened 2026-02-17
WIP: [#] Add CONTRIBUTING.md

mattrpav:


CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
+# Contributing Guidelines
```
jbonofre, line 1: The ASF header is missing here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Software Supply Chain  Security
+
+Apache ActiveMQ secures the software supply chain 
+
+1. Apache ActiveMQ committers sign every commit using an SSH key tied to their apache.org email address
+
+2. Apache ActiveMQ releases are monitored with reproducible builds to ensure published binaries match the source tree 
+
+## Credits
```
jbonofre, line 65: Not applicable here.

CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+
+Review our Apache [Security Policy](https://www.apache.org/security/). **Do not** file a public issue for security vulnerabilities.
+
+### Bug Reports and Other Issues
+
+## Feature Requests
+
+## Triaging Issues
+
+## Code Contributions
```
jbonofre, line 39: We can have a GH Issue template to help there.

jbonofre (commented):
Good idea !

FYI, here's `CONTRIBUTING.md` from Polaris that is interesting.

CONTRIBUTING.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Writing Commit Messages
+
+### Pull Request Review
+
+### Coding Style
+
+### Certification of origin
+
+1. Apache ActiveMQ committers should sign all commits using an SSH key tied to their apache.org email address
```
jeanouii, line 53: I configured mine because I agree this is a good practise. But I think you should bring the discussion to the dev@ mailing list and most likely to a vote. 

If you want to mention SSH, you should also mention GPG because most committers already have Apache signing keys for releases and it's pretty straight forward to reuse it to sign commits. (this is what I have done for instance).

jeanouii (changes requested):
Good start

CONTRIBUTING.md
jbonofre: FYI, that's not strictly required from ASF standpoint. I don't want to see anything that can be seen as "limiting" the contributions.
Not sure I'm a big fan here.

mattrpav: This shouldn't be seen as limiting. The benefit of using SSH keys it that Github (or anyone else) never has the private secret. This is more secure all-around and starting to become the standard practice (and requirement!) for SOC2 and ISO security certifications used by enterprises (aka end users of ActiveMQ).

As a project, ActveMQ can present a strong, and modern security stance by having committers sign commits vs relying on GH secrets or passkeys.

edit: I'll make a ssh signing quick-start guide to accompany this to show how easy it is to use over passwords/tokens/passkeys.

jbonofre: My point is that it should not be required, but recommanded.

mattrpav: I agree we can mark the statement as 'should' for now. 

I do think we should try to progress towards _required_. Perhaps we let the new processes settle in this year and review how its working.

mattrpav: > If you want to mention SSH, you should also mention GPG because most committers already have Apache signing keys for releases and it's pretty straight forward to reuse it to sign commits. (this is what I have done for instance).

One benefit of using SSH keys over GPG as it allows one key for git id/authn/authz and code signing.

jeanouii: GPG keys are used for Artifact signing. Github fully supports it for code signing as well.
If you want to use SSH for authentication that's ok, but you need 2 keys then.
Same applies if you use SSH key for code signing as anyways you need an GPG key for artifact signing.

So in the end I don't see any benefit of using one over the other. That's why I suggested to mentioned both.
The 'one key' argument only holds outside the ASF context. For Apache committers, GPG is already a prerequisite, making it the more natural choice for commit signing.

CONTRIBUTING.md
mattrpav: What do you mean by "not applicable"?

jbonofre: It's community driven effort, not a single individual.

I would just remove this section.

CONTRIBUTING.md
mattrpav: Most contributors (non-committers) aren't doing releases. So for 'contributors' an SSH key for authn/z to GH and for signing provides those users a 'one' key solution. 

I am good with updating the note to say 'should SSH (or GPG)..'

CONTRIBUTING.md
mattrpav: oh yes.. sorry. remove the 'credits' for sure, no problem

PR #1697, opened 2026-02-18
Protect the active branches

jbonofre:
The purpose of this PR is to protect our active branches (`activemq-6.2.x` and `activemq-5.19.x` as today):
- It's not allowed anymore to push directly on the branches. PR is required to change all branches.
- However, no review is necessary on the `activemq-6.2.x` and `activemq-5.19.x` (review should have happened on the `main` branch).

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -45,6 +45,20 @@ github:
         require_last_push_approval: false
         required_approving_review_count: 1
       required_linear_history: true
+    activemq-6.2.x:
+      required_pull_request_reviews:
+        require_code_owner_reviews: false
+        dismiss_stale_reviews: true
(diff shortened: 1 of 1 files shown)


jbonofre:
The purpose of using PRs is also to have exhaustive release notes on the branches.

jeanouii (approved):
I'm ok with it.
In practise, it means, we need to create a PR to cherry-pick changes. But that's ok

jbonofre:
@jeanouii that's correct. It's what we do in some projects (like Karaf), and it's pretty convenient to keep the GH Release details exhaustive, and also tests that the change still valid on other branches than `main`.

PR #1702, opened 2026-02-18
Use ephemeral ports in tests to avoid port conflicts

jeanouii:


cshannon:
@jeanouii - it looks like this branch was uploaded to the main report vs your fork, can we delete it?

PR #1705, opened 2026-02-19
Publish test results on failure

cshannon:
Update CI for all workflows to always publish test results even on failures

### DIFF
--- a/.github/workflows/ci-nightly.yml
+++ b/.github/workflows/ci-nightly.yml
@@ -80,13 +80,13 @@ jobs:
         shell: bash
         run: mvn -B -e -fae verify -Pactivemq.tests-quick -Dsurefire.rerunFailingTestsCount=3
       - name: Upload Test Results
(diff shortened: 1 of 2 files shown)


cshannon:
I guess we should cherry-pick this to other branches too? Do we need to open up a new PR now for cherry picks?

jbonofre:
@cshannon yes for the PR on active branches (protected).

cshannon:
@jbonofre and @jeanouii - Are we also missing config to view the html reports? By publishing the reports the goal for me was to be able to easily view the results without having to download an artifact, which appears to be the case now. With Jenkins it was easy to view the HTML report so I wanted to set up the same thing

jbonofre:
@cshannon No, you don't need to download, you have the results as comment (for instance https://github.com/apache/activemq/pull/1695#issuecomment-3920839166).
We can change the action to have a better rendering.

Issue #1699, opened 2026-02-18
ManagementContext should support sslContext injection


jbonofre:
If an user wants to use SSL with JMX, the only option today is to configure via the command line (`-Dcom.sun.management.jmxremote.port=xxxx -Dcom.sun.management.jmxremote.ssl=true`).

This configuration doesn't use the `sslContext` meaning that the user might have to duplicate the SSL configuration two times:
- the `sslContext` in `activemq.xml` that can be used in `transportConnector`
- the same configuration via the command line (`com.sun.management.jmxremote*`)

It would be great to be able to inject the `sslContext` in the `<managementContext/>` to use an unique configuration.

Something like:

```
    <sslContext>
      <sslContext id="fooContext"
        keyStore="file:./path/broker2.ks" keyStorePassword="password"
        trustStore="file:./path/client2.ks" trustStorePassword="password"/>
    </sslContext>

    <managementContext>
        <managementContext createConnector="true" sslContext="#fooContext"/>
    </managementContext> 
```

jeanouii:
Started already on this one, so I'll move it forward up to completion.

anmol-saxena-14:
Hi @jeanouii ,

I was also looking into the same issue. After reviewing your PR and the original context of AMQ-9857, I believe we should also consider adding a broker-level SSL fallback into ManagementContext, as this aligns closely with the original intent of the issue.

**From the issue context (AMQ-9857):**

<img width="965" height="159" alt="Image" src="https://github.com/user-attachments/assets/34e0330e-578f-45f8-9690-57ae1061adc1" />

Today, users can define SSL once in activemq.xml (broker <sslContext>) and reuse it for transport connectors. However, JMX requires a separate SSL setup (often via JVM com.sun.management.jmxremote.* arguments), which forces users to duplicate SSL configuration. The goal is to avoid configuring SSL twice and allow a single source of truth in broker configuration.

**Suggested logic in BrokerService.startManagementContext():**

If managementContext.getSslContext() == null and brokerService.getSslContext() != null, set management SSL from broker SSL.
Then start management context as usual.

To address, I have included additional logic in my PR:
https://github.com/apache/activemq/pull/1710

Adding these changes:
- Will preserves existing explicit behaviour: if managementContext.sslContext is set, that value still wins.
- It removes duplication for the common case: users define broker SSL once, and JMX automatically inherits it.
- It aligns JMX behavior with transport connectors, which already benefit from broker-level SSL defaults.
[...]

PR #1710, opened 2026-02-23
AMQ-9857: Support broker-level SSL fallback in ManagementContext to eliminate duplicate SSL configuration

anmol-saxena-14:


jeanouii:
Hi,
Thanks for the contribution.
I do have already #1661 to cover this. Maybe I missed something?

anmol-saxena-14:
Hi,
I have added a comment on the issue, Can you please check.
https://github.com/apache/activemq/issues/1699#issuecomment-3944335736
Thanks

jeanouii:
@asaxena14 So, if I understand, we need to somehow merge our 2 PRs, right?
Or is it better to keep only one of the 2?

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -2624,6 +2624,12 @@ public ConnectionContext getAdminConnectionContext() throws Exception {
 
     protected void startManagementContext() throws Exception {
         getManagementContext().setBrokerName(brokerName);
+        // Reuse the broker-level SSL context for JMX by default
+        // This avoids duplicating SSL config in activemq.xml while still allowing an
+        // explicit managementContext sslContext to override when one is needed
+        if (getManagementContext().getSslContext() == null && getSslContext() != null) {
```
mattrpav, line 2630: There is a logic flaw here-- just b/c an SslContext exists on the broker, does not mean the intent is to wire it to the management context.

If there is a fallback option, there most likely will need to be a config flag on the managementContext to affirm that ssl should be enabled.

jeanouii: @mattrpav right this was the option I used in my PR #1661

anmol-saxena-14: I agree, the PR #1661 already supports explicit wiring via: " <property name="sslContext" ref="brokerSslContext"/>". Adding a fallback code, guarded by an explicit flag, does not seem necessary and would add complexity without clear benefit.

anmol-saxena-14:
This is already covered by PR #1661.

PR #1726, opened 2026-02-25
[#1843] [AMQ-9687] Add metrics about error and reconnect counts to network co…

### LINKED ISSUE #1727: Add additional observability fields to NetworkConnector
The `NetworkBridgeView` MBean can expose new attributes like:

* `startedTimestamp` to know when the network bridge has started.
* `localExceptionCount` and `remoteExceptionCount` to know the numbers of exceptions on the bridge (local and remote)

I suspect we can consider additional attributes.

### LINKED ISSUE #1843: Add metrics about error and reconnect counts to network connectors and bridges
ref: [AMQ-9687]

mattrpav:
…nnector and network bridges

ref: [#1843]
ref: [AMQ-9687]

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
@@ -37,6 +37,7 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
 import java.util.regex.Pattern;
 
 import javax.management.ObjectName;
@@ -129,6 +130,8 @@ public abstract class DemandForwardingBridgeSupport implements NetworkBridge, Br
     protected final AtomicBoolean localBridgeStarted = new AtomicBoolean(false);
     protected final AtomicBoolean remoteBridgeStarted = new AtomicBoolean(false);
     protected final AtomicBoolean bridgeFailed = new AtomicBoolean();
+    // Ensures a single bridge failure counts exactly one local/remote exception
+    protected final AtomicBoolean bridgeExceptionCounted = new AtomicBoolean();
     protected final AtomicBoolean disposed = new AtomicBoolean();
     protected BrokerId localBrokerId;
     protected ActiveMQDestination[] excludedDestinations;
@@ -171,6 +174,7 @@ public abstract class DemandForwardingBridgeSupport implements NetworkBridge, Br
     private final ExecutorService syncExecutor = Executors.newSingleThreadExecutor();
     private Transport duplexInboundLocalBroker = null;
     private ProducerInfo duplexInboundLocalProducerInfo;
+    private final AtomicLong startedTimestamp = new AtomicLong(0L);
 
     public DemandForwardingBridgeSupport(NetworkBridgeConfiguration configuration, Transport localBroker, Transport remoteBroker) {
         this.configuration = configuration;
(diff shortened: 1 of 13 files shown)


activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
```diff
@@ -171,6 +172,7 @@ public abstract class DemandForwardingBridgeSupport implements NetworkBridge, Br
     private final ExecutorService syncExecutor = Executors.newSingleThreadExecutor();
     private Transport duplexInboundLocalBroker = null;
     private ProducerInfo duplexInboundLocalProducerInfo;
+    private AtomicLong startedTimestamp = new AtomicLong(0L);
```
cshannon:
```suggestion
    private final AtomicLong startedTimestamp = new AtomicLong(0L);
```

activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
```diff
@@ -372,6 +375,7 @@ public void run() {
                     // Once we have all required broker info we can attempt to start
                     // the local and then remote sides of the bridge.
                     doStartLocalAndRemoteBridges();
+                    startedTimestamp.set(System.currentTimeMillis());
```
cshannon: I would move this to the start() method just because the other call to set back to 0 is in the stop() method so it makes more sense. I'd also add it to a try/finally probably. [Here](https://github.com/apache/activemq/blob/5dcc72b779c3b51b32bbad3daa6b44a7f19648b4/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L262) is what i mean.

Something like:

```java
try {
    triggerStartAsyncNetworkBridgeCreation();
} catch (IOException e) {
    LOG.warn("Caught exception from remote start", e);
} finally {
    startedTimestamp.set(System.currentTimeMillis());
}
```

activemq-broker/src/main/java/org/apache/activemq/network/NetworkConnector.java
```diff
@@ -164,22 +172,35 @@ public static ActiveMQDestination[] getDurableTopicDestinations(final Set<Active
...
+        stoppedTimestamp.set(System.currentTimeMillis());
+        startedTimestamp.set(0l);
     }
 
     protected void handleStart() throws Exception {
         if (localURI == null) {
             throw new IllegalStateException("You must configure the 'localURI' property");
         }
         LOG.info("Network Connector {} started", this);
+        bridgeExceptionCounter.setEnabled(true);
```
cshannon: Shouldn't this not be true by default? I thought network statistics were optional and configurable?

activemq-unit-tests/src/test/java/org/apache/activemq/bugs/AMQ4160Test.java
```diff
@@ -351,6 +366,7 @@ public ObjectName getMbeanObjectName() {
                     public void resetStats(){
                         next.resetStats();
                     }
+
```
cshannon: nit: remove extra line

cshannon (changes requested):

activemq-broker/src/main/java/org/apache/activemq/network/NetworkConnector.java
mattrpav: I'm going to follow the pattern of tracking what the brokerService has configured.

cshannon (dismissed):
seems fine to me if all the tests pass

jbonofre (approved):
It's a nice addition and looks good to me.

cshannon (approved):

cshannon:
Our CI is still annoyingly flaky, I kicked off another build. After 6.3.0 I think we need to try and tackle more flaky tests

mattrpav:
> Our CI is still annoyingly flaky, I kicked off another build. After 6.3.0 I think we need to try and tackle more flaky tests

I made an issue to track flaky tests observed during this release and will tackle them as part of 6.4.0:
https://github.com/apache/activemq/issues/2238

Issue #1714, opened 2026-02-25
Implement `Async Send`


jeanouii:
`CompletionListener` API not yet implemented. It is required to fully pass 3.1 TCK

PR #1728, opened 2026-02-25
[#1714] async send completion listener

jeanouii:


cshannon:
This will need to be reviewed in detail, there were problems with the previous attempt to address this in https://github.com/apache/activemq/pull/1364 with not meeting the spec

One of the biggest issues as commented [here](https://github.com/apache/activemq/pull/1364#issuecomment-2535866454) was this part of the spec:
https://jakarta.ee/specifications/messaging/3.1/jakarta-messaging-spec-3.1#use-of-the-completionlistener-by-the-jakarta-messaging-provider

>7.3.8. Use of the CompletionListener by the Jakarta Messaging provider
A session will only invoke one CompletionListener callback method at a time. For a given MessageProducer or JMSContext, callbacks (both onCompletion and onException) will be performed in the same order as the corresponding calls to the asynchronous send method.

@jeanouii - I haven't reviewed this yet but does this PR address that part (7.3.8) of the spec? Have you reviewed the other parts of the spec to verify if this is compatible?

jeanouii:
@cshannon Yes I took a look at the PR and previous discussions.

Yes it explicitely addresses

> 7.3.8. Use of the CompletionListener by the Jakarta Messaging provider
> A session will only invoke one CompletionListener callback method at a time. For a given MessageProducer or 
> JMSContext, callbacks (both onCompletion and onException) will be performed in the same order as the corresponding calls to the asynchronous send method.

Some key points I should have written in the PR description.

1. One callback at a time: the session holds an Executors.newSingleThreadExecutor(). As discussed in the last PR, this goes against parallelism but guarantees the constraint is met. Also we still have our `AsyncCallback` that can be used to increase concurrency.
2. Preserve order: the send is synchronized on the client side with the sendMutex. And the mutex is held until future.get() completes, which guarantees the sequentiality. This was your proposal I just implemented.
3. Callback not in the sender thread:  that's exactly why there is the single thread executor.

BTW by reviewing the PR again in order to verify the 3 points, I discovered a possible deadlock which I just fixed in the last commit. The build is still green and the TCK all pass.

Hope it helps

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
```diff
@@ -168,6 +168,7 @@ public Destination getDestination() throws JMSException {
      */
     @Override
     public void close() throws JMSException {
+        ActiveMQSession.checkNotInCompletionListenerCallback("close");
```
gemmellr: Not seeing anything to satisfy the requirement of close() waiting for CompletionListeners to complete (successfully or otherwise) before returning.

Similarly plays into parent Session and Connection closures that need to do the same (somewhat similar for MessageListener where there [already seems like there could be an issue](https://github.com/apache/activemq/pull/1643#issuecomment-3852789499)).

EDIT: ok so the fact later revealed that this 'asynchronous' send is being done entirely syncronously means this is probably only an issue for the _current_ send (if there is one).

activemq-client/src/main/java/org/apache/activemq/ActiveMQSession.java
```diff
@@ -235,6 +243,20 @@ public static interface DeliveryListener {
...
+        r -> new Thread(r, "ActiveMQ async-send"));
+
+    // Set to true on the executor thread while a CompletionListener callback is executing.
+    // Used to detect illegal session operations (close/commit/rollback) from within a callback.
+    static final ThreadLocal<Boolean> IN_COMPLETION_LISTENER_CALLBACK = ThreadLocal.withInitial(() -> false);
+
+    // Set to true on the dispatch thread while a MessageListener.onMessage() callback is executing.
+    // Used to detect illegal connection/session operations from within a MessageListener callback.
+    static final ThreadLocal<Boolean> IN_MESSAGE_LISTENER_CALLBACK = ThreadLocal.withInitial(() -> false);
+
```
gemmellr, line 258: This doesnt seem correct. The check is meant to be Session-specific, whilst this is JVM-wide, so it will give the wrong result if crossing between Sessions/Connections, which the completion and message listener threads are still free to do (they primarily have restrictions on use of _their own_ Session/Connection, which is what the checks need to cover). Elsewhere we track the Thread in the Session using an AtomicReference, then that can later be compared to currentThread at the appropriate spots.

gemmellr (commented):
I gave the async send bits a skim (ignored all the stuf really from other PRs), and noticed a couple things.

I kinda stopped looking though once I realised it isnt really implementing true asynchronous send at all (more just doing sync sends just with the async API, giving increasing API compatibility) and so various things its clearly not doing are thus far less of an issue right now. Even with that being the case though, the usage restriction checks issue could be problematic for folks using multiple sessions/connections.

activemq-client/src/main/java/org/apache/activemq/ActiveMQSession.java
jeanouii: oh man! Very good catch @gemmellr 
Bad habit to have static thread locals. But you absolutely right, this is too restrictive compared to the spec that only requires per session. I'll fix this.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
jeanouii: Yes, this is a step forward to implement the spec, but you are correct. This is under the cover synchronous. This is purely spec related, and was a clear design choice. We still have the `AsyncCallback` for high throughput. 

That being said, I'm happy to push this as a first implementation attempt and then think about a better way to implement this. Eager to hear your thoughts and happy to give it a try. I'm too young for the moment :-)

activemq-client/src/main/java/org/apache/activemq/ActiveMQSession.java
gemmellr: I could swear I selected both Threadlocals before commenting, but its only highlighting the lower one; the comment applied to both if that wasnt already obvious.

jbertram:
@jeanouii, you said:
> ...This is under the cover synchronous. This is purely spec related, and was a clear design choice.

I expect users will assume that they're getting real async given that's the intention of the spec. If users aren't getting real async then this should be clearly called out in the documentation.

activemq-client/src/main/java/org/apache/activemq/ActiveMQSession.java
jeanouii: Yes yes, I understood that it was the 2 lines and I fully agree. Stupid mistake

cshannon:
> 2. Preserve order: the send is synchronized on the client side with the sendMutex. And the mutex is held until future.get() completes, which guarantees the sequentiality. This was your proposal I just implemented.

As long as the spec is met I (speaking for myself) think that is ok as I feel like that is better than not implementing it at all. Truly async is certainly much harder to implement correctly given the spec requirements, so I'm ok starting with synchronous to at least get the API implemented but others may not agree with me.

If going the sync route, we can document the behavior (including javadocs etc) so users understand it's only one message at a time. The good news is that if we are following the spec we can make a future update to be truly async without breaking anyone.

jeanouii:
@jbertram Yes we can certainly make sure it's clear for the users. Agreed with @cshannon and this is really what this is for. Having a spec compliant implementation. We always improve the boiler plate under the API to support full async. 

All great discussion and feedback.
Do we have a wiki to capture things like this?

On my work projects I usually have an docs/adr with md/adoc files in there to capture technical thoughts and choices.
I like to have them versioned along side with the project, but maybe we use a wiki or any other means?

jbonofre (commented):
For the discussion/feedback, we can use GitHub Discussions (linked to the PR), or (as discussed in the past on the dev@ mailing list), we can also create md files for the website.

I took a quick look on this one and I have some comments:

The PR looks ~90% compliant to the spec (I did a new read on the spec). To my understanding, the main "issue" is about the recover() restriction check.
I'm completely fine with the "not truly async" aspect as soon as it's documented. Ok, it doesn't deliver the performance benefit the spec suggests, but it's still spec compliant.
I think the PR is solid as a first implementation, we can improve it later.

gemmellr:
Yep a fully-synchronous 'async' send is entirely permitted within the spec, it actually explicitly details that possibility in the API docs.

jeanouii:
Added javadoc and implementation details note on the website

https://github.com/apache/activemq-website/pull/174

jbonofre:
If we can get this one for 6.4.0 it would be great (else it will be in 6.5.0).

mattrpav:
@jeanouii how about squashing the commits down a bit?  I don't think we'll need to keep all 9.

We should target this for v6.5.0.

I checked out this branch and it passed all the related tests in the TCK!

cshannon:
I agree 6.5.0 is good, but squashing isn't a big deal, we will squash on merge either way.

PR #1736, opened 2026-03-07
Fix Jakarta 3.1 compliance for Character property types

pradeep85841:
This PR addresses a spec compliance issue in ActiveMQMessage regarding allowed property types.

TCK Alignment: Updated checkValidObject to reject Character types when isNestedMapAndListEnabled is set to false, as required by the Jakarta Messaging 3.1 specification.

Backward Compatibility: Maintained support for Character types when the legacy flag is true (default) to prevent regressions for existing users.

Validation: Added unit tests to verify that MessageFormatException is correctly thrown in strict mode while verifying successful property assignment in legacy mode.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java
@@ -143,8 +143,7 @@ public class ActiveMQConnection implements Connection, TopicConnection, QueueCon
 
     /**
      * If set to true, strict Jakarta Messaging 3.1 compliance is enforced.
-     * This rejects non-standard property types (like Character) and forces
-     * legacy features like nestedMapAndListEnabled to false.
+     * This strictly rejects non-standard property types such as Character, Map, and List.
      */
     private boolean strictCompliance = false;
 
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionFactory.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionFactory.java
@@ -125,8 +125,7 @@ public class ActiveMQConnectionFactory extends JNDIBaseStorable implements Conne
 
     /**
      * If set to true, strict Jakarta Messaging 3.1 compliance is enforced.
-     * This rejects non-standard property types (like Character) and forces
-     * legacy features like nestedMapAndListEnabled to false.
+     * This strictly rejects non-standard property types such as Character, Map, and List.
      */
     private boolean strictCompliance = false;
 
@@ -418,7 +417,6 @@ public class ActiveMQConnectionFactory extends JNDIBaseStorable implements Conne
 
         // Copy the compliance flags from the Factory to the Connection
         connection.setStrictCompliance(isStrictCompliance());
-        connection.setNestedMapAndListEnabled(isNestedMapAndListEnabled());
 
         return connection;
     }
@@ -1017,15 +1015,10 @@ public class ActiveMQConnectionFactory extends JNDIBaseStorable implements Conne
     }
 
     /**
-     * If this flag is set then the connection follows strict Jakarta Messaging
-     * compliance. Enabling this will automatically force nestedMapAndListEnabled to false.
+     * If this flag is set then the connection follows strict Jakarta Messaging compliance.
      */
     public void setStrictCompliance(boolean strictCompliance) {
         this.strictCompliance = strictCompliance;
-        if (strictCompliance) {
-            this.setNestedMapAndListEnabled(false);
-            LOG.info("Strict compliance enabled: nestedMapAndListEnabled has been forced to false to align with Jakarta specifications.");
-        }
     }
 
     public String getClientIDPrefix() {
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
@@ -527,15 +527,20 @@ public class ActiveMQMessage extends Message implements org.apache.activemq.Mess
     protected void checkValidObject(Object value) throws MessageFormatException {
 
         boolean valid = value instanceof Boolean || value instanceof Byte || value instanceof Short || value instanceof Integer || value instanceof Long;
-        valid = valid || value instanceof Float || value instanceof Double || value instanceof String || value == null;
+        valid = valid || value instanceof Float || value instanceof Double || value instanceof Character || value instanceof String || value == null;
 
         ActiveMQConnection conn = getConnection();
-        // Legacy behavior: Character is valid ONLY if strictCompliance is OFF
-        if (conn == null || !conn.isStrictCompliance()) {
-            valid = valid || value instanceof Character;
+
+        // strict rejection for Character
+        if (valid && conn != null && conn.isStrictCompliance() && value instanceof Character) {
+            throw new MessageFormatException("Character type not allowed under strict Jakarta 3.1 compliance");
         }
 
         if (!valid) {
+            // Check strict compliance at validation time without side-effecting the 'nested' flag
+            if (conn != null && conn.isStrictCompliance()) {
+                throw new MessageFormatException("Only objectified primitive objects and String types are allowed under strict Jakarta 3.1 compliance but was: " + value + " type: " + value.getClass());
+            }
             // conn is null if we are in the broker rather than a JMS client
             if (conn == null || conn.isNestedMapAndListEnabled()) {
                 if (!(value instanceof Map || value instanceof List)) {
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQMessagePropertyTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQMessagePropertyTest.java
@@ -26,6 +26,9 @@ import jakarta.jms.Message;
 import jakarta.jms.MessageFormatException;
 import jakarta.jms.Session;
 
+import java.util.HashMap;
+import java.util.Map;
+
 import static org.apache.activemq.command.DataStructureTestSupport.assertEquals;
 import static org.junit.Assert.assertFalse;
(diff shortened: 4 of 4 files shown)


activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQMessagePropertyTest.java
```diff
@@ -0,0 +1,68 @@
+package org.apache.activemq;
```
jbonofre, line 17: The ASF header is missing.

jbonofre:
In order to be Jakarta Messaging 3.1 fully compliant, I would suggest to set `nestedMapAndListEnabled` to `false` by default, in order to not accept `Character`.

pradeep85841:
Set nestedMapAndListEnabled to false by default for Jakarta 3.1 compliance

activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java
```diff
@@ -150,7 +150,7 @@ public class ActiveMQConnection implements Connection, TopicConnection, QueueCon
     private boolean optimizeAcknowledge;
     private long optimizeAcknowledgeTimeOut = 0;
     private long optimizedAckScheduledAckInterval = 0;
-    private boolean nestedMapAndListEnabled = true;
+    private boolean nestedMapAndListEnabled = false;
```
jeanouii: I don't think this has any effect. The factory creating the connection defaults to true and overrides this value I think. To be verified and discarded if I'm not correct.

pradeep85841:
Thank you @jeanouii for the detailed review and suggestions. I agree with your points.

I will update the ActiveMQConnectionFactory as well to set the default to false. This ensures the setting correctly propagates to all connections as intended for Jakarta 3.1 compliance.

And will refactor the unit tests to use try-with-resources for better connection management. I will also ensure the embedded broker is explicitly stopped to avoid any flakiness in the CI.

I am working on these changes and will push the updated code shortly

activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java
jbonofre: Changing `nestedMapAndListEnabled` to `false` by default is a significant behavioral change that goes well beyond `Character` rejection.

With `false`, it means `Map` and `List` property values will also be rejected by default. This will break users who rely on the legacy nested map/list support (it's the behavior for years). 

I would suggest to not change in the context of this PR.

pradeep85841: Thank you for the guidance. I completely understand your point—changing the nestedMapAndListEnabled default is a very big change and will surely break many legacy users who have been using Maps and Lists for years. I will revert this default back to true.

As suggested, I will introduce a specific flag, complianceMode (or strictCompliance), to handle the Character rejection separately. This way, we can achieve Jakarta 3.1 compliance without affecting the long-standing behavior of Maps and Lists.

I will update the PR with these changes and the refactored tests. Please let me know if this sounds correct to you

jeanouii: I like the idea of not having a specific flag just for this. We are just starting working on JMS 3.1 compliance and we may encounter other situations so it may well end up by having many flags.
Having one flag for turning the entire broker into a compatible mode seems relevant to me.
We could as well force (and log of course) `nestedMapAndListEnabled` to `false` when the new strictCompliance flag is enabled? Thoughts?

jbonofre: @jeanouii agree, that was basically my proposal.

pradeep85841:
Introduced a strictCompliance flag in ActiveMQConnection and ActiveMQConnectionFactory.

When strictCompliance=true, it automatically forces nestedMapAndListEnabled=false and logs an info message.

Updated ActiveMQMessage to reject Character properties (and other non-standard types) only when strictCompliance is enabled.

Refactored tests to verify the master switch

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
```diff
@@ -523,17 +525,26 @@ public void setProperties(Map<String, ?> properties) throws JMSException {
...
+            if (conn != null && conn.isStrictCompliance()) {
+                throw new MessageFormatException("Only objectified primitive objects and String types are allowed when strictCompliance is enabled but was: " + value + " type: " + value.getClass());
+            }
+
             // conn is null if we are in the broker rather than a JMS client
             if (conn == null || conn.isNestedMapAndListEnabled()) {
-                if (!(value instanceof Map || value instanceof List)) {
-                    throw new MessageFormatException("Only objectified primitive objects, String, Map and List types are allowed but was: " + value + " type: " + value.getClass());
+                // Standard rules: only primitives and Strings are allowed
+                if (!(value instanceof Map || value instanceof List || value instanceof Character)) {
```
jeanouii: I think now that you have the other flag, you can revert the Character part here, what do you think?

jeanouii (commented):
Small comment to revert the changes in the other flag if possible

jeanouii:
@jbonofre I think that looks good now, what do you think?

It's a good compromise, isn't it?

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
```diff
@@ -525,11 +527,15 @@ public void setProperties(Map<String, ?> properties) throws JMSException {
     protected void checkValidObject(Object value) throws MessageFormatException {
 
         boolean valid = value instanceof Boolean || value instanceof Byte || value instanceof Short || value instanceof Integer || value instanceof Long;
-        valid = valid || value instanceof Float || value instanceof Double || value instanceof Character || value instanceof String || value == null;
+        valid = valid || value instanceof Float || value instanceof Double || value instanceof String || value == null;
```
jbonofre: As pointing out by @jeanouii, you are right here to revert the `Character` change in the validation.

However, as `Character` is now handled in the connection, the original line works. The strict check should only reject `Character` when `strictCompliance` is `true`.

Maybe it's easier (to read/maintenance) to split in two steps:

```
  // Keep original line unchanged:
  valid = valid || value instanceof Float || value instanceof Double
      || value instanceof Character || value instanceof String || value == null;

  // Then add strict rejection:
  if (valid && conn != null && conn.isStrictCompliance() && value instanceof Character) {
      throw new MessageFormatException("Character type not allowed under strict Jakarta 3.1 compliance");
  }
```

activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java
```diff
@@ -141,6 +141,13 @@ public class ActiveMQConnection implements Connection, TopicConnection, QueueCon
     private RedeliveryPolicyMap redeliveryPolicyMap;
     private MessageTransformer transformer;
 
+    /**
+     * If set to true, strict Jakarta Messaging 3.1 compliance is enforced.
+     * This rejects non-standard property types (like Character) and forces
+     * legacy features like nestedMapAndListEnabled to false.
```
jbonofre: I'm not sure it's fully correct.

Javadoc says enabling strict compliance "automatically forces nestedMapAndListEnabled to false." 

This couples two independent concerns. If someone explicitly sets `strictCompliance=true` and then `nestedMapAndListEnabled=true`, the second call would override it, leading to inconsistent state. 

Mayne we should consider either:
* just checking `strictCompliance` at validation time without side-effecting other flags
* or document the ordering dependency clearly in the Javadoc

activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionFactory.java
```diff
@@ -1000,6 +1012,22 @@ public void setNestedMapAndListEnabled(boolean structuredMapsEnabled) {
...
+
+    /**
+     * If this flag is set then the connection follows strict Jakarta Messaging
+     * compliance. Enabling this will automatically force nestedMapAndListEnabled to false.
+     */
+    public void setStrictCompliance(boolean strictCompliance) {
+        this.strictCompliance = strictCompliance;
+        if (strictCompliance) {
+            this.setNestedMapAndListEnabled(false);
+            LOG.info("Strict compliance enabled: nestedMapAndListEnabled has been forced to false to align with Jakarta specifications.");
```
jbonofre: nit: I would suggest to remove this log, or at least use `debug` instead of `info`.

activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionFactory.java
```diff
@@ -408,6 +415,11 @@ protected ActiveMQConnection createActiveMQConnection(String userName, String pa
     protected ActiveMQConnection createActiveMQConnection(Transport transport, JMSStatsImpl stats) throws Exception {
         ActiveMQConnection connection = new ActiveMQConnection(transport, getClientIdGenerator(),
                 getConnectionIdGenerator(), stats);
+
+        // Copy the compliance flags from the Factory to the Connection
+        connection.setStrictCompliance(isStrictCompliance());
+        connection.setNestedMapAndListEnabled(isNestedMapAndListEnabled());
```
jbonofre: I think we should remove this line as it's a behavior change (and unrelated to this PR).
Currently, the connection default is `nestedMapAndListEnabled=false` (if not overridden by the connection URI).
With this line, the factory's value always overrides.

This is probably correct, but I would remove `setNestedMapAndListEnabled()` from this PR.

pradeep85841:
Reverted the original validation line and added a secondary check to reject Character only when strictCompliance is enabled.

Removed the side-effect where strictCompliance would force nestedMapAndListEnabled to false, preventing inconsistent states as suggested.

Removed the INFO log in the setter and limited flag propagation in the factory to only include the new compliance flag.

Updated the unit tests to verify that strictCompliance correctly enforces Jakarta 3.1 rules even if legacy flags are explicitly enabled.

PR #1738, opened 2026-03-08
feat(#1737): Add OAuth2 JWT authentication support via JAAS LoginModule

### LINKED ISSUE #1737: Add OAuth2 support
The purpose here is to add a new JAAS LoginModule allowing authentication based on JWT.

We can use Nimbus JWT dependency to deal with that.

The idea is:
1. Introduce a new `OAuth2LoginModule` to be used by the `JaasAuthenticationPlugin`, similar to the existing `PropertiesLoginModule` or `LDAPLoginModule`.
2. Validates tokens locally, to avoid systematic call to the IdP on every connection.
3. The OAuth2 configuration (issuer, claim, etc) is defined in the `conf/login.config` as other `LoginModules`.
4. The user pass the token as "password"

jbonofre:
Add OAuth2LoginModule to activemq-jaas that validates JWT access tokens using JWKS endpoint for signature verification. 

Clients pass the JWT as the password field, and claims are mapped to UserPrincipal/GroupPrincipal for seamless integration with the existing JaasAuthenticationPlugin.

### DIFF
--- a/activemq-jaas/src/main/java/org/apache/activemq/jaas/OAuth2LoginModule.java
+++ b/activemq-jaas/src/main/java/org/apache/activemq/jaas/OAuth2LoginModule.java
@@ -22,6 +22,7 @@ import java.net.URL;
 import java.security.Principal;
 import java.text.ParseException;
 import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -104,7 +105,7 @@ public class OAuth2LoginModule implements LoginModule {
     private String groupsClaim;
 
     private String user;
-    private final Set<Principal> principals = new HashSet<>();
+    private final LinkedHashSet<Principal> principals = new LinkedHashSet<>();
     private boolean succeeded;
     private boolean commitSucceeded;
 
--- a/activemq-jaas/src/main/java/org/apache/activemq/jaas/OAuth2LoginModule.java
+++ b/activemq-jaas/src/main/java/org/apache/activemq/jaas/OAuth2LoginModule.java
@@ -118,7 +118,15 @@ public class OAuth2LoginModule implements LoginModule {
         this.debug = Boolean.parseBoolean((String) options.get("debug"));
 
         this.jwksUrl = (String) options.get(JWKS_URL_OPTION);
+        if (jwksUrl == null || jwksUrl.isEmpty()) {
+            throw new IllegalArgumentException("OAuth2 JWKS URL (" + JWKS_URL_OPTION + ") is required");
+        }
+
         this.issuer = (String) options.get(ISSUER_OPTION);
+        if (issuer == null || issuer.isEmpty()) {
+            throw new IllegalArgumentException("OAuth2 issuer (" + ISSUER_OPTION + ") is required");
+        }
+
         this.audience = (String) options.get(AUDIENCE_OPTION);
 
         String userClaim = (String) options.get(USERNAME_CLAIM_OPTION);
@@ -135,13 +143,6 @@ public class OAuth2LoginModule implements LoginModule {
 
     @Override
     public boolean login() throws LoginException {
-        if (jwksUrl == null || jwksUrl.isEmpty()) {
-            throw new LoginException("OAuth2 JWKS URL (oauth2.jwksUrl) is required");
-        }
-        if (issuer == null || issuer.isEmpty()) {
-            throw new LoginException("OAuth2 issuer (oauth2.issuer) is required");
-        }
-
         String token = getToken();
         if (token == null || token.isEmpty()) {
             throw new FailedLoginException("No JWT token provided");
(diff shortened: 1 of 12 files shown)


jbonofre:
NB: this is a better approach imho to what I did initially in https://github.com/apache/activemq/pull/1035 (with the same goal).

jbonofre:
The test failure is related to https://github.com/apache/activemq/pull/1741

activemq-jaas/src/main/java/org/apache/activemq/jaas/OAuth2LoginModule.java
```diff
@@ -0,0 +1,342 @@
...
+
+        if (debug) {
+            LOG.debug("OAuth2LoginModule initialized with jwksUrl={}, issuer={}, audience={}, usernameClaim={}, groupsClaim={}",
+                    jwksUrl, issuer, audience, usernameClaim, groupsClaim);
+        }
+    }
+
+    @Override
+    public boolean login() throws LoginException {
+        if (jwksUrl == null || jwksUrl.isEmpty()) {
```
jeanouii: I think if this is required, it could go in the init.
No need to let the broker start and login module initialize if we know upfront it will fail at login, generating tons of similar exceptions.

jeanouii: Same for issuer I think

jbonofre: That's a good point. Let me move in the `init` section.

jeanouii (commented):
Some small details. nothing important

jeanouii (dismissed):
LGTM

mattrpav:
-1 on merging this as part of the activemq-jaas module. This drags in 3rd party dependency to _all_ deployments -- I think this should go into a activemq-jaas-oath2 module.

mattrpav (changes requested):
I think this should go into a separate module and get deployed into an extras folder (with the 3rd party) dependency instead of installed into the classpath by default for _every_ running broker.

activemq-jaas/src/main/java/org/apache/activemq/jaas/OAuth2LoginModule.java
```diff
@@ -0,0 +1,343 @@
...
+    private boolean debug;
+
+    private String jwksUrl;
+    private String issuer;
+    private String audience;
+    private String usernameClaim;
+    private String groupsClaim;
+
+    private String user;
+    private final Set<Principal> principals = new HashSet<>();
```
mattrpav: This should be a LinkedHashSet to ensure ordering (UserPrincipal first)

mattrpav (changes requested):

jbonofre:
> I think this should go into a separate module and get deployed into an extras folder (with the 3rd party) dependency instead of installed into the classpath by default for _every_ running broker.

@mattrpav what is the concern about the Nimbus dependency in the JAAS module ?

At the end, regarding the ActiveMQ distribution, it will be a dependency like another (like Jetty or Rome for instance).

Your concern is about the "embedded" ActiveMQ service in application (`BrokerService`) ?

activemq-jaas/src/main/java/org/apache/activemq/jaas/OAuth2LoginModule.java
jbonofre: I'm fine to use LinkedHashSet but not sure it's strictly required.

jeanouii:
@mattrpav Using LinkedHashSet is a nice to have but not strictly required per spec. Actually Subject.getPrincipals() returns a Set, that's it. The rest is platform dependant. Our JaasAuthenticationBroker for instance gets all the principals and then iterates until it finds an instance of UserPrincipal. 
The other login modules all rely on HashSet, so it is now inconsistent with other ActiveMQ LoginModule.

On the dependency, it will be in the final zip anyways, so until it becomes a problem, I'd prefer to keep in in activemq-jaas instead of creating another module for one class only. It is both simpler for us and for our users that don't have to know they need another module to use oauth2 tokens. We could go like this and if it's a problem split later on. 
Jasypt is there as optional as well, so it's similar here. It won't be pulled as transitive dependency. 
On the other hand, we have a full independant shiro module. Both are valid, Shiro has much more code, so that's understandable. `activemq-jaas-oauth2` definitely makes sense, but it's one class for now so splitting now just makes it harder for us.

mattrpav:
> but it's one class for now so splitting now just makes it harder for us.

I think this line of thinking 'it's just one class' is what gets us into dependency trouble as a project. Yes, it is one class in the ActiveMQ tree, but this pulls in several transitive dependencies-- not just the one listed as optional. Being a security-related component also increases the probability that it will get tagged w/ a vulnerability (or one of its transitive deps will).

I think it would be most correct to have separate modules per security mechanism (the shiro appoarch). We could do an activemq-jaas-extras, but that doesn't seem to make sense b/c rarely are multiple external security providers (used outside of local accounts).

jbonofre:
@mattrpav that's fair. I'm not a big fan of `activemq-jaas-extras`, I would prefer to have `activemq-jaas-oauth2`.

I will update the PR accordingly.

jbonofre:
@jeanouii @mattrpav I moved the `OAuth2LoginModule` in a dedicated `activemq-oauth2` module.

jbonofre:
NB: I will update the `LICENSE`/`NOTICE` to document nimbus.

Issue #1739, opened 2026-03-08
Introduce `requestTimeout` on the `ConnectionFactory` to be used by `sendSyncPacket()`


jbonofre:
This is related to https://issues.apache.org/jira/browse/AMQ-6763.

In order to avoid `setXid()` stuck on `sendSyncPacket()`, I propose to introduce `requestTimeout` (0 by default).

PR #1740, opened 2026-03-08
feat(#1739): Add requestTimeout to ConnectionFactory to prevent hanging threads in syncSendPacket()

jbonofre:
This is a rework of https://github.com/apache/activemq/pull/644 (new investigation).

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/SyncSendPacketTimeoutTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/SyncSendPacketTimeoutTest.java
@@ -24,8 +24,13 @@ import jakarta.jms.JMSException;
 import jakarta.jms.MessageConsumer;
 import jakarta.jms.Session;
 
+import org.apache.activemq.broker.Broker;
+import org.apache.activemq.broker.BrokerFilter;
+import org.apache.activemq.broker.BrokerPlugin;
 import org.apache.activemq.broker.BrokerService;
+import org.apache.activemq.broker.ConnectionContext;
 import org.apache.activemq.broker.TransportConnector;
+import org.apache.activemq.broker.region.Subscription;
 import org.apache.activemq.command.ActiveMQQueue;
 import org.apache.activemq.command.ConsumerInfo;
 import org.apache.activemq.transport.RequestTimedOutIOException;
@@ -122,21 +127,48 @@ public class SyncSendPacketTimeoutTest {
 
     @Test
     public void testSyncSendPacketFailFromTimeout() throws Exception {
+        // Restart the broker with a plugin that delays addConsumer responses
+        broker.stop();
+        broker.waitUntilStopped();
+
+        broker = new BrokerService();
+        broker.setPersistent(false);
+        broker.setUseJmx(false);
+        broker.setPlugins(new BrokerPlugin[]{new BrokerPlugin() {
+            @Override
+            public Broker installPlugin(Broker broker) {
+                return new BrokerFilter(broker) {
+                    @Override
+                    public Subscription addConsumer(ConnectionContext context, ConsumerInfo info) throws Exception {
+                        // Only delay consumers on our test queue, not advisory consumers
+                        if (info.getDestination().getPhysicalName().equals("test")) {
+                            Thread.sleep(5000);
+                        }
+                        return super.addConsumer(context, info);
+                    }
+                };
+            }
+        }});
+        broker.addConnector("tcp://localhost:0");
+        broker.start();
+        broker.waitUntilStarted();
+        brokerUrl = broker.getTransportConnectors().get(0).getPublishableConnectString();
+
         ActiveMQConnectionFactory factory = new ActiveMQConnectionFactory(brokerUrl);
-        // Set to super short 1 millisecond so we always time out
-        factory.setRequestTimeout(1);
+        factory.setWatchTopicAdvisories(false);
+        factory.setRequestTimeout(500);
         try (ActiveMQConnection connection = (ActiveMQConnection) factory.createConnection()) {
+            connection.start();
+            Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+
             Exception exception = null;
             try {
-                connection.start();
-                try (Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
-                     MessageConsumer consumer = session.createConsumer(session.createQueue("test"))) {
-                    assertNotNull("Consumer should be created successfully", consumer);
-                }
+                session.createConsumer(session.createQueue("test"));
                 fail("Expected JMSException due to request timeout");
             } catch (JMSException expected) {
                 exception = expected;
             }
+            assertNotNull("Should have caught a JMSException", exception);
             assertEquals(RequestTimedOutIOException.class,
                     TransportConnector.getRootCause(exception).getClass());
         }


activemq-client/src/main/java/org/apache/activemq/transport/ResponseCorrelator.java
```diff
@@ -39,9 +39,13 @@
 public class ResponseCorrelator extends TransportFilter {
 
     private static final Logger LOG = LoggerFactory.getLogger(ResponseCorrelator.class);
+    // default max timeout for requests without an explicit timeout (0 means no timeout for backwards compatibility)
```
jeanouii: Agreed with the backward compatibility!

activemq-unit-tests/src/test/java/org/apache/activemq/transport/ResponseCorrelatorTest.java
```diff
@@ -0,0 +1,239 @@
...
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import org.apache.activemq.command.ExceptionResponse;
+import org.apache.activemq.command.Response;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+public class ResponseCorrelatorTest {
```
jeanouii: Could have been in the client module, but the Stubs are in activemq-unit-tests, so good with me. You can add the parallel annotation at least if possible. not a blocker, but uses more forks and the current test implementation is fine to run in parallel

activemq-unit-tests/src/test/java/org/apache/activemq/transport/ResponseCorrelatorTest.java
```diff
@@ -0,0 +1,239 @@
...
+                    result.set(correlator.request(new TestCommand()));
+                } catch (Throwable e) {
+                    error.set(e);
+                }
+            });
+
+            assertTrue("Request should have started", requestStarted.await(5, TimeUnit.SECONDS));
+
+            // Wait a bit to confirm the thread is still blocked
+            Thread.sleep(500);
```
jeanouii: Not my friend LOL
Usually causes a lot of flaky tests in CI under load.

jeanouii: Same bellow

activemq-unit-tests/src/test/java/org/apache/activemq/transport/ResponseCorrelatorTest.java
jbonofre: Yes for parallel annotation. And yes, I put this test here for the stubs (just to avoid to import/duplicate too much).

activemq-unit-tests/src/test/java/org/apache/activemq/transport/ResponseCorrelatorTest.java
jbonofre: That's very true, my bad. Let me improve that.

activemq-unit-tests/src/test/java/org/apache/activemq/transport/ResponseCorrelatorTest.java
jbonofre: The test is not parallel compliant.

activemq-unit-tests/src/test/java/org/apache/activemq/transport/ResponseCorrelatorTest.java
jbonofre: I removed `Thread.sleep()` for something more robust.

cshannon:
What is the use case for this? When were/are you seeing timeouts?

jbonofre:
@cshannon it's my new work on an old issue from Jira ([AMQ-6763](https://issues.apache.org/jira/browse/AMQ-6763)). I was able to reproduce it when using XA:  the broker never sent back a response to these XA commands. 

In my case, I reproduced the issue with simulating network failure: the TCP connection to the broker silently dropped, the request was sent but the response never arrives, and no TCP error was detected, meaning the XA thread hangs forever.

I believe that the similar behavior can happen:
- when the broker is not responsible (overloaded): the broker received the command but is stuck processing it (slow processing)
- if `wireFormat.maxInactivityDuration` is too high (or disabled), dead connections aren't detected promptly.

This PR aims to set timeout to avoid the thread handing forever.

cshannon:
@jbonofre -

Usually the timeout is set on a Connection and then passed when making the request as seen [at this line](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L1466).

Maybe we need to be passing the timeout value in more spots?  I think the code just defaults to 0 and not set in many places so the timeout is not used. I'm also not a huge fan of system properties as they are global and can't be easily set for a single connection, for example unless there is no other option.

jbonofre:
I can pass the timeout from the `Connection` (via the URL).

The System property is more for "transition" as I think we should have a timeout (0/no timeout is not a good idea for `FutureResponse` imho).

cshannon:
> @cshannon it's my new work on an old issue from Jira ([AMQ-6763](https://issues.apache.org/jira/browse/AMQ-6763)). I was able to reproduce it when using XA: the broker never sent back a response to these XA commands.
> 

Looking at it closer, it's getting stuck here in [setXid() ](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/TransactionContext.java#L759) when calling syncSendPacket() on the connection.

If you want to add a new configuration option for a default timeout I think you should move it one step deeper into the ActiveMQConnection, inside the [syncSendPacket()](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L1512) method. 

You can replace the 0 that is passed with a new configuration option (that will of course be 0 by default for compatibility). The configuration option can be configured on the connection as a URL parameter or setter like any other connection property and we can avoid the system property and also avoid having the timeout stuff in different locations.

There's already a [closeTimeout](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L157) so this could fit in with that and be called something like `requestTimeout` that defaults to 0.

tabish121:
> > @cshannon it's my new work on an old issue from Jira ([AMQ-6763](https://issues.apache.org/jira/browse/AMQ-6763)). I was able to reproduce it when using XA: the broker never sent back a response to these XA commands.
> 
> Looking at it closer, it's getting stuck here in [setXid() ](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/TransactionContext.java#L759) when calling syncSendPacket() on the connection.
> 
> If you want to add a new configuration option for a default timeout I think you should move it one step deeper into the ActiveMQConnection, inside the [syncSendPacket()](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L1512) method.
> 
> You can replace the 0 that is passed with a new configuration option (that will of course be 0 by default for compatibility). The configuration option can be configured on the connection as a URL parameter or setter like any other connection property and we can avoid the system property and also avoid having the timeout stuff in different locations.
> 
> There's already a [closeTimeout](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L157) so this could fit in with that and be called something like `requestTimeout` that defaults to 0.

There's also a sendTimeout and possibly a connectTimeout in ActiveMQConnection.  I recall adding sendTimeout a long time ago for similar reasons.

cshannon:
> There's also a sendTimeout and possibly a connectTimeout in ActiveMQConnection. I recall adding sendTimeout a long time ago for similar reasons.

Ah yeah, i see the [sendTimeout](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L160) already defined just below it, forgot about that. It is just used to pass to the producer (which makes sense it is only used on message sending).

It would be nice to re-used a timeout but I think we likely need a new one that is more broad, ie `requestTimeout`. This new ew configuration would be more of a default for all requests that were not already covered by other specific cases such as connection, close, and send.

tabish121:
> > There's also a sendTimeout and possibly a connectTimeout in ActiveMQConnection. I recall adding sendTimeout a long time ago for similar reasons.
> 
> Ah yeah, i see the [sendTimeout](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnection.java#L160) already defined just below it, forgot about that. It is just used to pass to the producer (which makes sense it is only used on message sending).
> 
> It would be nice to re-used a timeout but I think we likely need a new one that is more broad, ie `requestTimeout`. This new ew configuration would be more of a default for all requests that were not already covered by other specific cases such as connection, close, and send.

Yes you would likely add a new timeout option such as 'requestTimeout'. We have similar options in Qpid JMS which cover all cases of blocking calls, connectTimeout, closeTimeout, sendTimeout, requestTimeout and drainTimeout.

cshannon:
@jbonofre - One more thing, if adding a new requestTimeout property for the connection, you should added the property to the `ActiveMQConnectionFactory`. [Here](https://github.com/apache/activemq/blob/6060f9bcd3b38c47225db1162260be6309590c44/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionFactory.java#L143) is where sendTimeout is on the factory, so adding requestTimeout means you could do something like:` tcp://localhost:61656?jms.requestTimeout=30000`

jbonofre:
@cshannon @tabish121 @jeanouii I changed the approach as you suggested.

I introduce a configurable `requestTimeout` (default 0, no timeout as today) on `ActiveMQConnectionFactory` and `ActiveMQConnection`, similar to the existing `sendTimeout`. 

When set, `syncSendPacket(Command)` uses `poll(timeout)` instead of `take()`, throwing `RequestTimedOutIOException` when the timeout expires.
    
NB: it can be configured programmatically via `factory.setRequestTimeout(ms)` or via URL parameter `jms.requestTimeout=ms`.

activemq-unit-tests/src/test/java/org/apache/activemq/SyncSendPacketTimeoutTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+    @Test
+    public void testSyncSendPacketSucceedsWithRequestTimeout() throws Exception {
+        ActiveMQConnectionFactory factory = new ActiveMQConnectionFactory(brokerUrl);
+        factory.setRequestTimeout(5000);
+        ActiveMQConnection connection = (ActiveMQConnection) factory.createConnection();
+        try {
+            connection.start();
+            // Creating a session triggers syncSendPacket internally — should succeed within timeout
+            Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
```
cshannon: This is not actually testing syncSendPacket. Creating a session actually uses an async call [here](https://github.com/apache/activemq/blob/a2fe8919932463076671e3318c8a874e59c01f94/activemq-client/src/main/java/org/apache/activemq/ActiveMQSession.java#L259).

You could add an extra line to create a consumer and that would work as createConsumer() does indeed use syncSendPacket when creating the consumer [here](https://github.com/apache/activemq/blob/a2fe8919932463076671e3318c8a874e59c01f94/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java#L301). This should be done anywhere else that is relevant.

```java
connection.start();
Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
// Creating a consumer triggers syncSendPacket internally — should succeed within timeout
session.createConsumer(session.createQueue("test"));
```

activemq-unit-tests/src/test/java/org/apache/activemq/SyncSendPacketTimeoutTest.java
```diff
@@ -0,0 +1,119 @@
...
+            connection.close();
+        }
+    }
+
+    @Test
+    public void testSyncSendPacketSucceedsWithoutRequestTimeout() throws Exception {
+        ActiveMQConnectionFactory factory = new ActiveMQConnectionFactory(brokerUrl);
+        // requestTimeout=0 means no timeout (default)
+        ActiveMQConnection connection = (ActiveMQConnection) factory.createConnection();
+        try {
```
cshannon: Nit: You could use try-with-resources for all of these tests, but not a huge deal either way.

cshannon (changes requested):
The requestTimeout change looks good but I noticed a few things to fix.

**1.** I made a [comment](https://github.com/apache/activemq/pull/1740#discussion_r2908309160) inline but you need to do something else besides create a session, like creating a consumer, to trigger the sync send. creating a session is actually done [async](https://github.com/apache/activemq/blob/a2fe8919932463076671e3318c8a874e59c01f94/activemq-client/src/main/java/org/apache/activemq/ActiveMQSession.java#L259).

**2.** Inside of TransactionContext there is a spot that needs to be updated to use the request timeout value [here](https://github.com/apache/activemq/blob/a2fe8919932463076671e3318c8a874e59c01f94/activemq-client/src/main/java/org/apache/activemq/TransactionContext.java#L284). The line should be:

```java
this.connection.syncSendPacket(info, this.connection.isClosing() ?
    this.connection.getCloseTimeout() : this.connection.getRequestTimeout());
```

**3.** You should add a test to shows the timeout actually works, such as:

```java
@Test
public void testSyncSendPacketFailFromTimeout() throws Exception {
    ActiveMQConnectionFactory factory = new ActiveMQConnectionFactory(brokerUrl);
    // Set to super short 1 millisecond so we always time out
    factory.setRequestTimeout(1);
    try (ActiveMQConnection connection = (ActiveMQConnection) factory.createConnection()) {
        final Exception exception = assertThrows(JMSException.class,
            () -> {
                connection.start();
                Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
                session.createConsumer(session.createQueue("test"));
            });
        assertEquals(RequestTimedOutIOException.class,
                TransportConnector.getRootCause(exception).getClass());
    }
}
```
**4.** You could also add a test to verify that the method that passes along the timeout value overrides the default as intended:

```java
@Test
public void testSyncSendPacketOverrideDefaultRequestTimeout() throws Exception {
    ActiveMQConnectionFactory factory = new ActiveMQConnectionFactory(brokerUrl);
    try (ActiveMQConnection connection = (ActiveMQConnection) factory.createConnection()) {
        connection.start();
        ActiveMQSession session = (ActiveMQSession) connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
        // After session creation set the timeout default to be very short to test that
        // overriding directly works
        connection.setRequestTimeout(1);
        ConsumerInfo info = new ConsumerInfo(session.getSessionInfo(),
                session.getNextConsumerId().getValue());
        info.setDestination(new ActiveMQQueue("test"));
        // Send info packet with timeout override
        assertNotNull("Consumer should be created successfully with no timeout",
                connection.syncSendPacket(info, 5000));
    }
}
```

jbonofre:
@cshannon I addressed your comments.

activemq-unit-tests/src/test/java/org/apache/activemq/SyncSendPacketTimeoutTest.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    @Test
+    public void testSyncSendPacketSucceedsWithoutRequestTimeout() throws Exception {
+        ActiveMQConnectionFactory factory = new ActiveMQConnectionFactory(brokerUrl);
+        // requestTimeout=0 means no timeout (default)
+        try (ActiveMQConnection connection = (ActiveMQConnection) factory.createConnection()) {
+            connection.start();
+            try (Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE)) {
```
cshannon: You need to create a new consumer here as well to test the sync

activemq-unit-tests/src/test/java/org/apache/activemq/SyncSendPacketTimeoutTest.java
```diff
@@ -0,0 +1,156 @@
...
+    @Test
+    public void testSyncSendPacketSucceedsWithRequestTimeout() throws Exception {
+        ActiveMQConnectionFactory factory = new ActiveMQConnectionFactory(brokerUrl);
+        factory.setRequestTimeout(5000);
+        try (ActiveMQConnection connection = (ActiveMQConnection) factory.createConnection()) {
+            connection.start();
+            // Creating a session triggers syncSendPacket internally — should succeed within timeout
+            try (Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE);
+                 // Creating a consumer triggers syncSendPacket internally — should succeed within timeout
+                 MessageConsumer consumer = session.createConsumer(session.createQueue("TEST.QUEUE"))) {
```
cshannon: The formatting/indentation seems a bit off here

jbonofre: The formatting is correct here as it's part of the "try-on-resource" (following the session).

jbonofre: Let me check if we have a similar statement for the formatting.

jbonofre:
@cshannon updated according to your comments.

cshannon (dismissed):

jbonofre:
This closes #1739

jeanouii:
LGTM

I think all feedback has been addressed. We need probably a formal approve from @cshannon and @tabish121  and it's ready to go.

jbonofre:
@jeanouii @cshannon I did a rebase, so your approvals have been lost.

PR #1745, opened 2026-03-09
Improve PortfolioPublishServlet encoding

cshannon:
Use HTML encoding for output instead of URL encoding


activemq-web/src/main/java/org/apache/activemq/web/PortfolioPublishServlet.java
```diff
@@ -132,6 +134,6 @@ protected int getNumberOfMessages(HttpServletRequest request) {
     }
 
     protected String escape(String text) throws IOException {
-        return java.net.URLEncoder.encode(text, "UTF-8");
+        return text != null ? HtmlUtils.htmlEscape(text, "UTF-8") : null;
```
jbonofre, line 137: I'm fine using Spring `HtmlUtils` here for now.

jbonofre (approved):

PR #1751, opened 2026-03-10
fix(test): fix MQTTTest#testReceiveMessageSentWhileOffline flakiness

jbonofre:
`subscribe()` returns on `SUBACK` but the broker processes `ConsumerInfo` asynchronously, so the durable subscription may not be fully reactivated when `receive()` is called. Add `Wait.waitFor` on `isSubscriptionActive()` after subscribe to ensure the broker has activated the subscription before attempting to receive queued messages.

Also remove the try/catch that silently swallowed assertion failures mid-loop, which masked the real failure with a misleading count mismatch.


jbonofre:
@jeanouii I know you worked on the MQTT tests. I observed some flakiness still on MQTT.

I'm trying to fix here. Not sure it would be enough though.

jeanouii:
@jbonofre I fixed hundred of them. Some don't show up anymore, a few are still unstable sometimes. And a lot to fix I'm sure. From my experience, there is no magic receipe, push and see if it works in Github Actions.

I tried many local scripts to run in parallel or in a batch to see of some tests fails. They usually always work locally but fail in the CI.

PR #1757, opened 2026-03-11
feat(assembly): add Docker image build support via Maven profile (#1701)

### LINKED ISSUE #1701: Create docker image as part of the build
Currently, we are using a shell script to create Docker images (`assembly/src/docker`).
It means it's "out" of the build and the CI.

The Docker image should created within the Maven reactor.

jbonofre:
This closes #1701


assembly/pom.xml
```diff
@@ -703,5 +703,105 @@
...
+      </build>
+    </profile>
+    <profile>
+      <id>docker-deploy</id>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>io.fabric8</groupId>
+            <artifactId>docker-maven-plugin</artifactId>
+            <version>0.45.1</version>
```
jeanouii: We use the version twice. Not a blocker, but would be better to extract it and reuse it.

jeanouii: Also do we need goal -> push?
The docker images are built locally but never pushed, right?

jbonofre: We need the push for buildx to be able to do multi-platform images.

That's the different between "local" (building the Docker image for the local host platform) and the "push" that actually do multi-platforms image (`linux/amd64` and `linux/arm64`).

Good point for the plugin version, I will create a property to avoid the duplicate.

jbonofre:
1. The previous `build.sh` is not used anymore. I kept for "transition" purpose but I'm fine to delete it (to avoid confusion).
2. Good point about the Java version, let me improve this.
3. Let me create the `latest` alias too.

Thanks @jeanouii !

jbonofre:
@jeanouii do you mind to do a new pass ? I think I addressed your comments.

jeanouii:
I think we can merge.

Look good.
We can improve and have latest when necessary and see if we need or not the goal. It was there in the script before that's why I raised the point.

I'll approve the PR

PR #1758, opened 2026-03-11
fix(assembly): add missing jolokia-core to assembly descriptor

jbonofre:
The jolokia-core artifact (containing `org.jolokia.core.api.LogHandler`) is a transitive dependency of jolokia-server-core in Jolokia 2.x but was not listed in the assembly descriptor's explicit includes, causing a `NoClassDefFoundError` at runtime.


jeanouii (approved):
I noticed that the other day as well.

PR #1759, opened 2026-03-11
fix(test): batch of fixes for weekly only tests

jeanouii:


activemq-unit-tests/src/test/java/org/apache/activemq/transport/failover/FailoverComplexClusterTest.java
```diff
@@ -338,72 +364,201 @@ protected void setUp() throws Exception {
...
+     * Adds a network bridge to a broker and starts it immediately.
+     * This is used when the broker is already running.
+     */
+    private void addAndStartNetworkBridge(final BrokerService broker, final String bridgeName,
+                                          final String uri, final boolean duplex, final String destinationFilter) throws Exception {
+        final NetworkConnector network = broker.addNetworkConnector(uri);
+        network.setName(bridgeName);
+        network.setDuplex(duplex);
+        if (destinationFilter != null && !destinationFilter.isEmpty()) {
+            network.setDestinationFilter(bridgeName);
```
jbonofre: I think it should be `destinationFilter` here as argument (not `bridgeName`), right ?

jeanouii: Bad copy/paste sorry.
Thanks for catching it

Issue #1752, opened 2026-03-10
Introduce a Kubernetes Operator to manage ActiveMQ brokers


jbonofre:
As part of the cloud friendly effort, a nice feature would be to provide a Kubernetes Operator to manage ActiveMQ brokers.

PR #1761, opened 2026-03-13
feat(helm): add Helm chart for ActiveMQ broker deployment

jbonofre:
Add a production-ready Helm chart for deploying ActiveMQ on Kubernetes, including StatefulSet with persistent KahaDB storage, multi-protocol services (OpenWire, AMQP, STOMP, MQTT, WebSocket), security hardening, Ingress, HPA, PDB, Prometheus ServiceMonitor, and example values files.

@kenliao94 do you mind to review ? It's a merge between what you did and what I did :)


helm/activemq/templates/hpa.yaml
```diff
@@ -0,0 +1,51 @@
...
+software distributed under the License is distributed on an
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+*/}}
+
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
```
jeanouii: This is a stupid and dummy question lol
Is this going to create isolated brokers without Network Of Broker to link them all and actually split the load?

helm/activemq/templates/statefulset.yaml
```diff
@@ -0,0 +1,223 @@
...
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      containers:
+        - name: {{ .Chart.Name }}
+          securityContext:
+            {{- toYaml .Values.securityContext | nindent 12 }}
+          image: {{ include "activemq.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          env:
+            - name: ACTIVEMQ_CONNECTION_USER
```
jeanouii, line 65: I double checked all variables and they are all valid in the docker image for activemq

jeanouii (commented):
This is all out of my skills to be honest. I just asked stupid questions and did some sanity check. Sorry to not be more useful.

kenliao94:
Awesome @jbonofre taking a look right now

helm/activemq/Chart.yaml
```diff
@@ -0,0 +1,38 @@
...
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+apiVersion: v2
+name: activemq
+description: Apache ActiveMQ - the most popular open source, multi-protocol, Java-based message broker
+type: application
+version: 6.3.0-SNAPSHOT
```
kenliao94, line 22: Are we launching this on 6.3? Not back porting it to 6.2?

jbonofre: I'm fine to create a backport on 6.2.x branch.

That said, regarding the number of improvements for 6.3.0, I will focus on 6.3.0 as first citizen 😄

helm/activemq/values.yaml
```diff
@@ -0,0 +1,287 @@
...
+  tcpSocket:
+    port: openwire
+  initialDelaySeconds: 10
+  periodSeconds: 10
+  timeoutSeconds: 5
+  failureThreshold: 3
+
+# -- Horizontal Pod Autoscaler configuration
+autoscaling:
+  # -- Enable HPA
```
kenliao94: How does HPA works in this case? If it spins up more replica, the lock on KahaDB will guarantee there is only one active instance. How does HPA work with Network of Broker?

jbonofre: That's an anticipation on https://github.com/apache/activemq/issues/1771 and https://github.com/apache/activemq/issues/1752.

But you are right, as it's not yet fully supported, maybe better to remove it.

Thoughts ?

helm/activemq/values.yaml
```diff
@@ -0,0 +1,287 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# -- Number of ActiveMQ broker replicas
+replicaCount: 1
+
+image:
+  # -- Container image repository
+  repository: apache/activemq-classic
```
kenliao94: if i run `helm install my-activemq ./helm/activemq` it will run the chart appVersion which is now set to 6.3 which doesn't exist on dockerhub yet. Should we change doc and set the appVersion to 6.2?

kenliao94 (changes requested):

helm/activemq/templates/hpa.yaml
jbonofre: I removed the HPA for now.

helm/activemq/values.yaml
jbonofre: I removed the HPA for now (waiting network of broker with k8s discovery).

helm/activemq/values.yaml
jbonofre: Good catch, also the `repository` is not the one we should use.

jbonofre: Updated.

jbonofre:
@jeanouii @kenliao94 by the way (as mentioned earlier), I plan to add the `index.yaml` on `dist.apache.org` allowing user to directly add the ActiveMQ Helm repo.

kenliao94:
LGTM thanks @jbonofre for the PR!

kenliao94 (approved):

jeanouii (approved):

chrisalbright:
Is this helm chart likely to be published?

PR #1762, opened 2026-03-13
fix(broker): improve error message for RMI registry creation failure

jbonofre:
Include the port number and actionable guidance in the debug log when creating the local RMI registry fails, so operators can quickly diagnose and resolve the issue.

This error usually happens if another process is using the Management Connector port. The purpose is to improve the message for the users.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagementContext.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagementContext.java
@@ -567,9 +567,9 @@ public class ManagementContext implements Service {
             Attribute attr = new Attribute("Port", connectorPort);
             mbeanServer.setAttribute(namingServiceObjectName, attr);
         } catch(ClassNotFoundException e) {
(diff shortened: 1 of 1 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ManagementContext.java
```diff
@@ -569,7 +569,9 @@ private void createConnector(MBeanServer mbeanServer) throws IOException {
         } catch(ClassNotFoundException e) {
             LOG.debug("Probably not using JRE 1.4: {}", e.getLocalizedMessage());
         } catch (Throwable e) {
-            LOG.debug("Failed to create local registry. This exception will be ignored.", e);
+            LOG.debug("Failed to create local RMI registry on port {}. This may happen when another process or broker instance "
```
jbonofre: I would also propose to use `warn` instead of `debug` here (to clearly inform the user).

jeanouii: I agree!
Improving the log message is great, but it has to be visible. DEBUG does not seem adequate here so +1 for WARN instead

PR #1769, opened 2026-03-13
feat(otel): Add OpenTelemetry distributed tracing support

### LINKED ISSUE #1768: Add OpenTelemetry plugin
The idea is to have OpenTelemetry plugin that intercept broker events and trace what's happening in the broker, a.k.a creating span for:
* message production
* message consumption
* message ack

It means we will be able to add a OpenTelemetry exporter, allowing to populate Prometheus, Zipkin, ...

jbonofre:
New activemq-opentelemetry module that instruments the broker using the OpenTelemetry API. The plugin traces send (PRODUCER), dispatch (CONSUMER), and acknowledge (INTERNAL) operations with W3C TraceContext propagation and standard OTel messaging semantic conventions.

Depends on opentelemetry-api only; users bring their own SDK and exporter at runtime. Included in the distribution with example configuration. Maybe we should add a default SDK/exporter in the ActiveMQ distribution, just wondering 😄

### DIFF
--- a/activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/ActiveMQMessageTextMapSetter.java
+++ b/activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/ActiveMQMessageTextMapSetter.java
@@ -1,0 +1,41 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.broker.util.opentelemetry;
+
+import java.io.IOException;
+
+import io.opentelemetry.context.propagation.TextMapSetter;
+import org.apache.activemq.command.Message;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ActiveMQMessageTextMapSetter implements TextMapSetter<Message> {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ActiveMQMessageTextMapSetter.class);
+
+    public static final ActiveMQMessageTextMapSetter INSTANCE = new ActiveMQMessageTextMapSetter();
+
+    @Override
+    public void set(Message message, String key, String value) {
+        try {
+            message.setProperty(key, value);
+            message.setMarshalledProperties(null);
+        } catch (IOException e) {
+            LOG.warn("Failed to set trace context property '{}' on message", key, e);
+        }
+    }
+}
--- a/activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/ActiveMQMessageTextMapSetter.java
+++ b/activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/ActiveMQMessageTextMapSetter.java
@@ -33,6 +33,7 @@ public class ActiveMQMessageTextMapSetter implements TextMapSetter<Message> {
     public void set(Message message, String key, String value) {
         try {
             message.setProperty(key, value);
+            // Clear the cached marshalled form so the injected property is included on next serialization.
             message.setMarshalledProperties(null);
         } catch (IOException e) {
             LOG.warn("Failed to set trace context property '{}' on message", key, e);
--- a/activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
+++ b/activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
@@ -1,0 +1,202 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.broker.util.opentelemetry;
+
+import java.util.concurrent.ConcurrentHashMap;
+
+import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.SpanBuilder;
+import io.opentelemetry.api.trace.SpanKind;
+import io.opentelemetry.api.trace.StatusCode;
+import io.opentelemetry.api.trace.Tracer;
+import io.opentelemetry.context.Context;
+import io.opentelemetry.context.propagation.TextMapPropagator;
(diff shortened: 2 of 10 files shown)


activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+                LOG.warn("Failed to create deliver span", e);
+            }
+        }
+        super.preProcessDispatch(messageDispatch);
+    }
+
+    @Override
+    public void postProcessDispatch(MessageDispatch messageDispatch) {
+        if (messageDispatch != null) {
+            Span span = dispatchSpans.remove(messageDispatch);
```
jeanouii, line 159: Because by nature, this can't be done in a try/finally block, some spans may leak in the map forever.
The map needs to be either bound or we need to implement a TTL to remove old entries in my opinion.

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+                .setAttribute("messaging.system", "activemq")
+                .setAttribute("messaging.destination.name", destinationName)
+                .setAttribute("messaging.operation", "publish");
+
+        if (messageSend.getMessageId() != null) {
+            spanBuilder.setAttribute("messaging.message.id", messageSend.getMessageId().toString());
+        }
+        if (producerExchange.getConnectionContext() != null
+                && producerExchange.getConnectionContext().getClientId() != null) {
+            spanBuilder.setAttribute("messaging.client_id", producerExchange.getConnectionContext().getClientId());
```
jeanouii, line 106: We use `client_id` a lot, but naming conventions require `.` (dot)

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+        Tracer tracer = GlobalOpenTelemetry.getTracer(INSTRUMENTATION_NAME);
+
+        // Extract any existing context from the message (e.g., set by client)
+        Context parentContext = propagator.extract(Context.current(), messageSend, ActiveMQMessageTextMapGetter.INSTANCE);
+
+        String destinationName = messageSend.getDestination().getPhysicalName();
+        SpanBuilder spanBuilder = tracer.spanBuilder(destinationName + " publish")
+                .setSpanKind(SpanKind.PRODUCER)
+                .setParent(parentContext)
+                .setAttribute("messaging.system", "activemq")
```
jeanouii, line 97: "messaging.system" --> "messaging.system.name"?

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+                .setAttribute("messaging.destination.name", destinationName)
+                .setAttribute("messaging.operation", "publish");
```
jeanouii, line 99: "messaging.operation" --> "messaging.operation.type"?

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @Override
+    public void send(ProducerBrokerExchange producerExchange, Message messageSend) throws Exception {
+        if (!traceProducer) {
+            super.send(producerExchange, messageSend);
+            return;
+        }
+
+        TextMapPropagator propagator = GlobalOpenTelemetry.getPropagators().getTextMapPropagator();
+        Tracer tracer = GlobalOpenTelemetry.getTracer(INSTRUMENTATION_NAME);
```
jeanouii, line 88: Shouldn't these 2 lines be done in the constructor only and not for each invocation?

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/ActiveMQMessageTextMapSetter.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(ActiveMQMessageTextMapSetter.class);
+
+    public static final ActiveMQMessageTextMapSetter INSTANCE = new ActiveMQMessageTextMapSetter();
+
+    @Override
+    public void set(Message message, String key, String value) {
+        try {
+            message.setProperty(key, value);
+            message.setMarshalledProperties(null);
```
jeanouii, line 36: This is to force the properties to be injected, correct?
what's the performance impact?

activemq-opentelemetry/src/test/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPluginTest.java
```diff
@@ -0,0 +1,244 @@
...
+    public void testPublishSpanCreatedOnSend() throws Exception {
+        MessageProducer producer = session.createProducer(session.createQueue(queueName));
+        producer.setDeliveryMode(DeliveryMode.NON_PERSISTENT);
+
+        TextMessage message = session.createTextMessage("test");
+        producer.send(message);
+        producer.close();
+
+        // Allow async broker processing to complete
+        Thread.sleep(500);
```
jeanouii, line 110: Can you use the WaitFor pattern instead?

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
```diff
@@ -0,0 +1,202 @@
...
+        if (traceConsumer && messageDispatch != null && messageDispatch.getMessage() != null) {
+            try {
+                TextMapPropagator propagator = GlobalOpenTelemetry.getPropagators().getTextMapPropagator();
+                Tracer tracer = GlobalOpenTelemetry.getTracer(INSTRUMENTATION_NAME);
+
+                Message message = messageDispatch.getMessage();
+                Context extractedContext = propagator.extract(Context.current(), message, ActiveMQMessageTextMapGetter.INSTANCE);
+
+                String destinationName = message.getDestination().getPhysicalName();
+                SpanBuilder spanBuilder = tracer.spanBuilder(destinationName + " deliver")
```
jeanouii, line 136: How is this going to show up for topics?
Many spans for the same message?

jeanouii (commented):
Great PR and a good start.
I've added a couple of notes and questions. 

It's been a while since I worked with this.

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
jbonofre: Yes, it's multiple messages per topic here.

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/ActiveMQMessageTextMapSetter.java
jbonofre: No performance impact here as we don't marshal the message.

activemq-opentelemetry/src/main/java/org/apache/activemq/broker/util/opentelemetry/OpenTelemetryBrokerPlugin.java
jbonofre: Addressed: replaced the unbounded `ConcurrentHashMap` with a `Collections.synchronizedMap`-wrapped `LinkedHashMap` that overrides `removeEldestEntry`. When the map exceeds `MAX_DISPATCH_SPANS` (1000 entries) the oldest span is ended and evicted automatically, bounding memory regardless of whether `postProcessDispatch` is ever called.

jbonofre:
@cshannon wdyt about this one for 6.3.0?

cshannon:
> @cshannon wdyt about this one for 6.3.0?

I haven't looked at this at all and this definitely needs to wait. Did 

> @cshannon wdyt about this one for 6.3.0?

I think 6.4.0 makes more sense, I don't think anyone else has paid attention to this (I know I haven't) and adding a major new framework like this is a pretty big deal, even if it's off by default. I think it would be good to have some more time.

jbonofre:
6.4.0 is good idea.

PR #1793, opened 2026-03-17
Improve FactoryFinder Validation

cshannon:
Also modernizes with generics and adds improved testing

**Edit:**

Adding more details on the changes that were done here now that [CVE-2026-33227](https://www.cve.org/CVERecord?id=CVE-2026-33227) has been published.

* FactoryFinder and and our default ObjectFactory impls now verify any classes loaded implement the required interface
* FactoryFinder and ObjectFactory optionally takes an allow list to ensure classes loaded are on the list. The config can be null (everything allowed), empty (nothing allowed), or a list of classes.
* FactoryFinder and ObjectFactory now use generics which makes the code nicer where factories are used 
* ObjectFactory has a new default method that takes the required interface and allowlist as arguments. If not implemented it defaults back to the original to not break existing users who may have switched it out as it's pluggable. Our two impls (the default and OSGi Activator) do not allow using the original method.
* FactoryFinder now performs path validation to prevent path traversal before invoking ObjectFactory
* All spots using FactoryFinder were updated with the correct interface type. 
* ProtocolConverter (Stomp) has had its FactoryFinder for translators updated to limit the classes to the defaults as users provide the value to load translators. There is a system property `org.apache.activemq.stomp.FRAME_TRANSLATOR_CLASSES` to control it. The value is a CSV list of classes with empty string allowing none and asterisk allowing all.
* QueueBrowseServlet (web) has had its FactoryFinder for views updated to limit the classes as user input is also used for views. There is a system property `org.apache.activemq.web.view.QUEUE_BROWSE_CLASSES` to control it. The value is a CSV list of classes with empty string allowing none and asterisk allowing all.
* A bunch of tests were added for everything

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
+++ b/activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
@@ -54,6 +54,18 @@ public class FactoryFinder<T> {
          */
         Object create(String path) throws IllegalAccessException, InstantiationException, IOException, ClassNotFoundException;
 
+        /**
+         * This method loads objects by searching for classes in the given path.
+         * A requiredType and Set of allowed implementations are provided for
+         * implementations to use for validation. Note is up to the actual implementations
+         * that implement {@link ObjectFactory} to decide how to use both parameters.
+         * By default, the method just delegates to {@link #create(String)}
+         *
+         * @param path the full service path
+         * @param requiredType the requiredType any objects must implement
+         * @param allowedImpls The set of allowed impls
+         * @return
+         */
         @SuppressWarnings("unchecked")
         default <T> T create(String path, Class<T> requiredType, Set<Class<? extends T>> allowedImpls)
                 throws IllegalAccessException, InstantiationException, IOException, ClassNotFoundException {
@@ -152,10 +164,13 @@ public class FactoryFinder<T> {
     private final Class<T> requiredType;
     private final Set<Class<? extends T>> allowedImpls;
 
-    public FactoryFinder(String path, Class<T> requiredType) {
-        this(path, requiredType, null);
-    }
-
+    /**
+     *
+     * @param path The path to search for impls
+     * @param requiredType Required interface type that any impl must implement
+     * @param allowedImpls The list of allowed implementations. If null or asterisk
+     * then all impls of the requiredType are allowed.
+     */
     public FactoryFinder(String path, Class<T> requiredType, String allowedImpls) {
         this.path = Objects.requireNonNull(path);
         this.requiredType = Objects.requireNonNull(requiredType);
@@ -210,7 +225,7 @@ public class FactoryFinder<T> {
         // Example: "/dir1/dir2/dir3/../file" becomes "/dir1/dir2/file"
         final Path resolvedPath = Path.of(path).resolve(key).normalize();
 
-        // Validate the resoled path is still within the original defined
+        // Validate the resolved path is still within the original defined
         // root path and throw an error of it is not.
         if (!resolvedPath.startsWith(path)) {
             throw new InstantiationException("Provided key escapes the FactoryFinder configured directory");
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerFactory.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerFactory.java
@@ -32,7 +32,8 @@ import org.apache.activemq.util.IOExceptionSupport;
 public final class BrokerFactory {
 
     private static final FactoryFinder<BrokerFactoryHandler> BROKER_FACTORY_HANDLER_FINDER
-            = new FactoryFinder<>("META-INF/services/org/apache/activemq/broker/", BrokerFactoryHandler.class);
+            = new FactoryFinder<>("META-INF/services/org/apache/activemq/broker/", BrokerFactoryHandler.class,
+            null);
 
     private BrokerFactory() {        
     }
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/group/GroupFactoryFinder.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/group/GroupFactoryFinder.java
@@ -26,7 +26,8 @@ import org.apache.activemq.util.URISupport;
 
 public class GroupFactoryFinder {
     private static final FactoryFinder<MessageGroupMapFactory> GROUP_FACTORY_FINDER =
-            new FactoryFinder<>("META-INF/services/org/apache/activemq/groups/", MessageGroupMapFactory.class);
+            new FactoryFinder<>("META-INF/services/org/apache/activemq/groups/", MessageGroupMapFactory.class,
+                    null);
 
     private GroupFactoryFinder() {
     }
--- a/activemq-broker/src/main/java/org/apache/activemq/transport/auto/AutoTcpTransportServer.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/transport/auto/AutoTcpTransportServer.java
@@ -81,11 +81,13 @@ public class AutoTcpTransportServer extends TcpTransportServer {
     protected int protocolDetectionTimeOut = 30000;
 
     private static final FactoryFinder<TransportFactory> TRANSPORT_FACTORY_FINDER =
-            new FactoryFinder<>("META-INF/services/org/apache/activemq/transport/", TransportFactory.class);
+            new FactoryFinder<>("META-INF/services/org/apache/activemq/transport/", TransportFactory.class,
+                    null);
     private final ConcurrentMap<String, TransportFactory> transportFactories = new ConcurrentHashMap<String, TransportFactory>();
 
     private static final FactoryFinder<WireFormatFactory> WIREFORMAT_FACTORY_FINDER =
-            new FactoryFinder<>("META-INF/services/org/apache/activemq/wireformat/", WireFormatFactory.class);
+            new FactoryFinder<>("META-INF/services/org/apache/activemq/wireformat/", WireFormatFactory.class,
+                    null);
 
     public WireFormatFactory findWireFormatFactory(String scheme, Map<String, Map<String, Object>> options) throws IOException {
         WireFormatFactory wff = null;
--- a/activemq-client/src/main/java/org/apache/activemq/transport/TransportFactory.java
+++ b/activemq-client/src/main/java/org/apache/activemq/transport/TransportFactory.java
@@ -37,9 +37,11 @@ import org.apache.activemq.wireformat.WireFormatFactory;
 public abstract class TransportFactory {
 
     private static final FactoryFinder<TransportFactory> TRANSPORT_FACTORY_FINDER
-            = new FactoryFinder<>("META-INF/services/org/apache/activemq/transport/", TransportFactory.class);
+            = new FactoryFinder<>("META-INF/services/org/apache/activemq/transport/", TransportFactory.class,
+            null);
     private static final FactoryFinder<WireFormatFactory> WIREFORMAT_FACTORY_FINDER
-            = new FactoryFinder<>("META-INF/services/org/apache/activemq/wireformat/", WireFormatFactory.class);
+            = new FactoryFinder<>("META-INF/services/org/apache/activemq/wireformat/", WireFormatFactory.class,
+            null);
     private static final ConcurrentMap<String, TransportFactory> TRANSPORT_FACTORYS = new ConcurrentHashMap<String, TransportFactory>();
 
     private static final String WRITE_TIMEOUT_FILTER = "soWriteTimeout";
--- a/activemq-client/src/main/java/org/apache/activemq/transport/discovery/DiscoveryAgentFactory.java
+++ b/activemq-client/src/main/java/org/apache/activemq/transport/discovery/DiscoveryAgentFactory.java
@@ -27,7 +27,8 @@ import org.apache.activemq.util.IOExceptionSupport;
 public abstract class DiscoveryAgentFactory {
 
     private static final FactoryFinder<DiscoveryAgentFactory> DISCOVERY_AGENT_FINDER =
-            new FactoryFinder<>("META-INF/services/org/apache/activemq/transport/discoveryagent/", DiscoveryAgentFactory.class);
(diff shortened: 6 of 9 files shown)


activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/StompTest.java
```diff
@@ -23,15 +23,19 @@
 import static org.junit.Assert.assertTrue;
 import static org.junit.Assert.fail;
 
+import jakarta.jms.Destination;
+import java.awt.Frame;
```
jbonofre: I guess this is a typo or IDE "auto-completion" here 😄 
It can be removed.

activemq-broker/src/main/java/org/apache/activemq/util/osgi/Activator.java
```diff
@@ -172,7 +172,15 @@ protected void unregister(long bundleId) {
     // ================================================================
 
     @Override
-    public Object create(String path) throws IllegalAccessException, InstantiationException, IOException, ClassNotFoundException {
+    public Object create(String path)
+            throws IllegalAccessException, InstantiationException, IOException, ClassNotFoundException {
+        throw new UnsupportedOperationException("Create is not supported without requiredType and allowed impls");
```
jbonofre, line 177: nit: as it's a "breaking change", we should remember to note in the release note.

jbonofre (approved):

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
```diff
@@ -138,21 +149,114 @@ public static void setObjectFactory(ObjectFactory objectFactory) {
...
+        // Example: "/dir1/dir2/dir3/../file" becomes "/dir1/dir2/file"
+        final Path resolvedPath = Path.of(path).resolve(key).normalize();
+
+        // Validate the resoled path is still within the original defined
+        // root path and throw an error of it is not.
+        if (!resolvedPath.startsWith(path)) {
+            throw new InstantiationException("Provided key escapes the FactoryFinder configured directory");
+        }
+
+        return resolvedPath.toString();
```
jeanouii, line 234: With your comment above, this might fail on Windows, because `/dir1/dir2/file` will probably be in reality `\dir1\dir2\file` so I think you will probably need to consider doing 
return resolvedPath.toString().replace('\\', '/');

What do you think?

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
```diff
@@ -44,51 +52,54 @@ public interface ObjectFactory {
...
     protected static class StandaloneObjectFactory implements ObjectFactory {
-        final ConcurrentMap<String, Class> classMap = new ConcurrentHashMap<String, Class>();
+        final ConcurrentMap<String, Class<?>> classMap = new ConcurrentHashMap<>();
 
         @Override
-        public Object create(final String path) throws InstantiationException, IllegalAccessException, ClassNotFoundException, IOException {
-            Class clazz = classMap.get(path);
+        public Object create(final String path)
+                throws IllegalAccessException, InstantiationException, IOException, ClassNotFoundException {
+            throw new UnsupportedOperationException("Create is not supported without requiredType and allowed impls");
```
jeanouii, line 85: Same comment as @jbonofre did above for the Activator.

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
```diff
@@ -44,51 +52,54 @@ public interface ObjectFactory {
          * @param path the full service path
          * @return
          */
-        public Object create(String path) throws IllegalAccessException, InstantiationException, IOException, ClassNotFoundException;
+        Object create(String path) throws IllegalAccessException, InstantiationException, IOException, ClassNotFoundException;
 
+        @SuppressWarnings("unchecked")
+        default <T> T create(String path, Class<T> requiredType, Set<Class<? extends T>> allowedImpls)
+                throws IllegalAccessException, InstantiationException, IOException, ClassNotFoundException {
+            return (T) create(path);
```
jeanouii, line 72: This is bypassing all the class validation, path validation and allowed impls. Anyone setting a custom ObjectFactory can bypass validations.
Is it a design choice?
Maybe a javadoc comment if that's the case.

activemq-broker/src/main/java/org/apache/activemq/util/osgi/Activator.java
cshannon: I would not call this a breaking change because ActiveMQ never calls this method. FactoryFinder now calls the new method in the ObjectFactory interface so this new method is never invoked unless someone extended or implemented the interface themselves or wrote something custom. If someone has their own version of ObjectFactory everything works fine because the new method just delegates to the legacy. This is only breaking if it's expected users would invoke Activator themselves and not just the broker but I don't see why we would expect that.

activemq-stomp/src/test/java/org/apache/activemq/transport/stomp/StompTest.java
cshannon: This is probably just the IDE as you mentioned. At some point it would be nice be nice to set up a plugin like spotless to check imports

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
```diff
@@ -138,21 +149,114 @@ public static void setObjectFactory(ObjectFactory objectFactory) {
...
-        this.path = path;
+    public FactoryFinder(String path, Class<T> requiredType, String allowedImpls) {
+        this.path = Objects.requireNonNull(path);
+        this.requiredType = Objects.requireNonNull(requiredType);
+        this.allowedImpls = loadAllowedImpls(requiredType, allowedImpls);
     }
 
+    @SuppressWarnings("unchecked")
+    private static <T> Set<Class<? extends T>> loadAllowedImpls(Class<T> requiredType, String allowedImpls) {
+        // If allowedImpls is either null or an asterisk (allow all wild card) then set to null so we don't filter
```
jeanouii, line 182: Useful comment to move to javadoc maybe so it can benefit developers/contributors.

Now, I have one concern. The default 2 args contructor defaults to allow-all. Usually from a security perspective, we tend to deny-all by default. This PR being about improving security, I'm surprised. That being said, I see that for internet facing factories (Stomp and Web) are using a strict list of impl which is great.
For the internal factories (like TransportFactory), the default is applied and we use allow all whereas we know in advance the list, so we could switch to deny all by default and provide the list of known implementations for internal factories as well.
What do you think?

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
cshannon: This was done by default for backwards compatibility and also I don't know why we would do anything if someone plugs in their own impl. If someone provides their own custom impl then it is on them to do whatever they want in my opinion.

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
cshannon: There's not really much point of having a 2 arg constructor if the default is deny all as it's pointless to create a FactoryFinder that can't do load anything. We could just only have the 3 arg constructor and then just update all the uses to provide the known lists

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
cshannon: I don't think we need to do this because forward slashes should work fine with windows in java as this validation check doesn't result in any different output than previously. Previously we just used the path with forward slashes and add on the key, now we do the same but before returning just resolve it and check it first. So the resulting value should be the same as before on both windows/linux if it passes validation.

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
cshannon: Actually there are a ton of impls for some of them and the impls are not all loaded on the same classpath so it's not easy to build the list. I think we can just get rid of the 2 arg and have the 3 arg and at least make the calling code explicilty set null if not caring to check

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
cshannon: I added a javadoc that makes mention it's up to the implementations to decide if the parameters are used. Maybe in 6.3.0 or another version where we can have breaking changes we could just drop the create(path) method entirely and just have the new method. Ultimately though, as I said, it's just an interface so implementations can ignore or do what they want.

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
```diff
@@ -138,21 +161,117 @@ public static void setObjectFactory(ObjectFactory objectFactory) {
...
+
+    Class<T> getRequiredType() {
+        return requiredType;
+    }
+
+    private String resolvePath(final String key) throws InstantiationException {
+        // Normalize the base path with the given key. This
+        // will resolve/remove any relative ".." sections of the path.
+        // Example: "/dir1/dir2/dir3/../file" becomes "/dir1/dir2/file"
+        final Path resolvedPath = Path.of(path).resolve(key).normalize();
```
rmannibucau, line 226: need to move toAbsolutePath somehow (both paths) to be comparable otherwise the startsWith can fail even if "true"

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
```diff
@@ -138,21 +161,117 @@ public static void setObjectFactory(ObjectFactory objectFactory) {
...
+
+        return resolvedPath.toString();
+    }
+
+    public static String buildAllowedImpls(Class<?>...classes) {
+        return Arrays.stream(Objects.requireNonNull(classes, "List of allowed classes may not be null"))
+                .map(Class::getName).collect(Collectors.joining(","));
+    }
+
+    public static <T> void validateClass(Class<?> clazz, Class<T> requiredType,
```
rmannibucau, line 242: if you want to validate the class you must validate the class name before loading it otherwise you can have side effects and defeat the validation no?

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
```diff
@@ -138,21 +161,117 @@ public static void setObjectFactory(ObjectFactory objectFactory) {
...
+        ClassLoader loader = Thread.currentThread().getContextClassLoader();
+        if (loader != null) {
+            try {
+                clazz = loader.loadClass(className);
+            } catch (ClassNotFoundException e) {
+                // ignore
+            }
+        }
+        if (clazz == null) {
+            clazz = FactoryFinder.class.getClassLoader().loadClass(className);
```
rmannibucau, line 271: classloader can be null there so a fallback on the system classloader is likely needed

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
jeanouii: I agree with this compromise.

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
cshannon: The class is is fine to load, what you want to do is avoid initialization and calling classloader.loadClass() does not initialize

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
cshannon: This was existing behavior so could be changed in a future PR if needed

activemq-client/src/main/java/org/apache/activemq/util/FactoryFinder.java
rmannibucau: You actually can't assume that with AMQ since it runs in complex classloader envs where it is wrong several (in some OSGi or EE setup for ex)

snicoll:
We've upgraded to ActiveMQ 6.2.2 and it looks like this change makes our integration tests fail on a Windows runner:

```
Caused by: java.io.IOException: Could not find factory class for resource: META-INF\services\org\apache\activemq\transport\vm	
at org.apache.activemq.util.FactoryFinder$StandaloneObjectFactory.loadProperties(FactoryFinder.java:127)	
at org.apache.activemq.util.FactoryFinder$StandaloneObjectFactory.create(FactoryFinder.java:93)	
at org.apache.activemq.util.FactoryFinder.newInstance(FactoryFinder.java:211)	
at org.apache.activemq.transport.TransportFactory.findTransportFactory(TransportFactory.java:186)	
at org.apache.activemq.transport.TransportFactory.connect(TransportFactory.java:68)	
at org.apache.activemq.ActiveMQConnectionFactory.createTransport(ActiveMQConnectionFactory.java:361)	
at org.apache.activemq.ActiveMQConnectionFactory.createActiveMQConnection(ActiveMQConnectionFactory.java:376)	
at org.apache.activemq.ActiveMQConnectionFactory.createActiveMQConnection(ActiveMQConnectionFactory.java:334)	
at org.apache.activemq.ActiveMQConnectionFactory.createConnection(ActiveMQConnectionFactory.java:226)	
at org.springframework.jms.connection.SingleConnectionFactory.doCreateConnection(SingleConnectionFactory.java:466)	
at org.springframework.jms.connection.SingleConnectionFactory.initConnection(SingleConnectionFactory.java:425)	
at org.springframework.jms.connection.SingleConnectionFactory.getConnection(SingleConnectionFactory.java:331)	
at org.springframework.jms.connection.SingleConnectionFactory.createConnection(SingleConnectionFactory.java:236)
```

The same test doesn't fail on Linux.

cshannon:
> We've upgraded to ActiveMQ 6.2.2 and it looks like this change makes our integration tests fail on a Windows runner:
> 
> ```
> The same test doesn't fail on Linux.

This might be related to this comment https://github.com/apache/activemq/pull/1793#discussion_r2952460699 which I didn't think would be an issue. Our windows CI is passing so it might depend on the environment

cshannon:
I opened up #1831 to fix this

PR #1796, opened 2026-03-18
feat(#1797): add message delivery tracking to QueueMessageReference

### LINKED ISSUE #1797: Track message delivery state in QueueMessageReference
Add a delivered flag to QueueMessageReference to distinguish between messages that are pending dispatch and messages that have been dispatched to a consumer but  
  not yet acknowledged.                                                                                                                                           
                                                                                                                                                                  
  Currently, once a message is dispatched to a consumer, there is no way to know whether the consumer has confirmed receipt (via delivered ack) or if the message is
   still in-flight. This adds a delivered boolean that is set to true when processing DeliveredAck in PrefetchSubscription, enabling better visibility into message
  lifecycle state.

jeanouii:
Add isDelivered()/setDelivered() to QueueMessageReference interface with implementations in IndirectMessageReference (boolean field) and NullMessageReference (no-op). PrefetchSubscription.acknowledge() now marks messages as delivered=true when processing DeliveredAck, enabling downstream consumers to distinguish delivered-but-unacked messages.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/IndirectMessageReference.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/IndirectMessageReference.java
@@ -35,6 +35,8 @@ public class IndirectMessageReference implements QueueMessageReference {
     private boolean dropped;
     /** Has the message been acked? */
     private boolean acked;
+    /** Has the message been acked? */
+    private boolean delivered;
     /** Direct reference to the message */
     private final Message message;
     private final MessageId messageId;
@@ -197,16 +199,25 @@ public class IndirectMessageReference implements QueueMessageReference {
 
     @Override
     public synchronized int getSize() {
-       return message.getSize();
+        return message.getSize();
     }
 
     @Override
     public boolean isAdvisory() {
-       return message.isAdvisory();
+        return message.isAdvisory();
     }
 
     @Override
     public boolean canProcessAsExpired() {
         return message.canProcessAsExpired();
+    }
+
+    @Override
+    public boolean isDelivered() {
+        return delivered;
+    }
+
+    public void setDelivered(final boolean delivered) {
+        this.delivered = delivered;
     }
 }
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/IndirectMessageReference.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/IndirectMessageReference.java
@@ -35,7 +35,7 @@ public class IndirectMessageReference implements QueueMessageReference {
     private boolean dropped;
     /** Has the message been acked? */
     private boolean acked;
-    /** Has the message been acked? */
+    /** Has the message been delivered? */
     private boolean delivered;
     /** Direct reference to the message */
     private final Message message;
@@ -213,11 +213,12 @@ public class IndirectMessageReference implements QueueMessageReference {
     }
 
     @Override
-    public boolean isDelivered() {
+    public synchronized boolean isDelivered() {
         return delivered;
     }
 
-    public void setDelivered(final boolean delivered) {
+    @Override
+    public synchronized void setDelivered(final boolean delivered) {
         this.delivered = delivered;
     }
 }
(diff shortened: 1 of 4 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/IndirectMessageReference.java
```diff
@@ -35,6 +35,8 @@ public class IndirectMessageReference implements QueueMessageReference {
     private boolean dropped;
     /** Has the message been acked? */
     private boolean acked;
+    /** Has the message been acked? */
```
jbonofre: The javadoc doesn't seem correct to me. I think it's a copy from `acked` javadoc.


```suggestion
    /** Has the message been delivered? */.
```

activemq-broker/src/main/java/org/apache/activemq/broker/region/IndirectMessageReference.java
```diff
@@ -197,16 +199,25 @@ public boolean isExpired() {
...
+        return message.isAdvisory();
     }
 
     @Override
     public boolean canProcessAsExpired() {
         return message.canProcessAsExpired();
     }
+
+    @Override
+    public boolean isDelivered() {
```
jbonofre: I guess it should be `synchronized`.


```suggestion
    public synchronized boolean isDelivered() {
```

activemq-broker/src/main/java/org/apache/activemq/broker/region/IndirectMessageReference.java
```diff
@@ -197,16 +199,25 @@ public boolean isExpired() {
...
+        return delivered;
+    }
+
+    public void setDelivered(final boolean delivered) {
```
jbonofre: I guess it should be `synchronized`.


```suggestion
    public synchronized void setDelivered(final boolean delivered) {
```

jbonofre: `@Override` seems missing here.


```suggestion
    @Override
    public void setDelivered(final boolean delivered) {
```

activemq-broker/src/main/java/org/apache/activemq/broker/region/PrefetchSubscription.java
```diff
@@ -270,6 +270,27 @@ public final void acknowledge(final ConnectionContext context,final MessageAck a
...
+                        if (ack.getLastMessageId().equals(messageId)) {
+                            destination = (Destination) node.getRegionDestination();
+                            callDispatchMatched = true;
+                            break;
+                        }
+                    }
+                }
+
                 int index = 0;
                 for (Iterator<MessageReference> iter = dispatched.iterator(); iter.hasNext(); index++) {
```
jbonofre: It seems this loop overrides what has been done in the previous loop (line 275). Maybe we should merge the logic in one loop ?

cshannon:
What caused you to make this change, did you see bugs show up or specific issues with tests? There are no test updates here to demonstrate the issue and fix. I'm not saying the change is incorrect (i haven't reviewed it) but i am wondering what triggered the change here.

jbonofre:
@cshannon I think it's coming for a flaky test (or needed by a test). @jeanouii am I right ?

cshannon:
> @cshannon I think it's coming for a flaky test (or needed by a test). @jeanouii am I right ?

Yeah that would make sense, if there is an existing test already that is flaky that this fixes then that demonstrates the issue and fix but it would be helpful to know which one

jeanouii:
@cshannon @jbonofre Not a specific test, but I asked myself many times the question when debugging flaky tests. And I realized even in the Web Console we don't have information on hand. 
I don't recall about a very specific test to be honest. 
If you guys think we don't need it, I'm ok with it, I'll put it back locally when I need to debug again in the future

cshannon:
> If you guys think we don't need it, I'm ok with it, I'll put it back locally when I need to debug again in the future

I just need time to look at it more, this week has been pretty busy so I'll try to get to the various backlogs of PR reviews next week.

jbonofre (approved):

jbonofre:
This PR looks good to me but I would like @cshannon's opinion.

cshannon (approved):

cshannon:
this closes #1797

PR #1801, opened 2026-03-18
Backport: remove the annotation on the ProtocolConverterTest (#1793)

jbonofre:


jbonofre:
@cshannon the `ParallelTest` class introduced for the STOMP converter test was missing from the backport PR.

PR #1809, opened 2026-03-19
Enforce non-negative delivery delay in strictCompliance mode

pradeep85841:
- Implemented setDeliveryDelay and getDeliveryDelay in ActiveMQMessageProducer
- Added Jakarta 3.1 compliance check to reject negative delays when strictCompliance is enabled
- Added unit tests to verify both strict enforcement and legacy behaviour

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
@@ -228,7 +228,7 @@ public class ActiveMQMessageProducer extends ActiveMQMessageProducerSupport impl
 
         // Jakarta 3.1 Compliance Guard
         if (deliveryDelay < 0 && session.connection.isStrictCompliance()) {
-            throw new jakarta.jms.MessageFormatException("Delivery delay cannot be negative in strictCompliance mode.");
+            throw new jakarta.jms.MessageFormatException("Delivery delay cannot be negative when strictCompliance is enabled.");
         }
 
         this.deliveryDelay = deliveryDelay;
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQMessageProducerTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQMessageProducerTest.java
@@ -18,8 +18,6 @@ package org.apache.activemq;
 
 import jakarta.jms.Connection;
 import jakarta.jms.MessageProducer;
-import jakarta.jms.Session;
-import jakarta.jms.Topic;
 import org.junit.Test;
 
 import org.slf4j.Logger;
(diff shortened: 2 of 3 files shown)


pradeep85841:
- Updated ActiveMQJMS2ContextTest to reflect that these methods are now fully supported, removing legacy UnsupportedOperationException assertions.
- Added validation to throw MessageFormatException for negative delays when strictCompliance is enabled, as required by Jakarta Messaging 3.1.
- Maintained backward compatibility for legacy/non-strict mode.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
```diff
@@ -84,6 +84,8 @@ public class ActiveMQMessageProducer extends ActiveMQMessageProducerSupport impl
     private MessageTransformer transformer;
     private MemoryUsage producerWindow;
 
+    protected long deliveryDelay = 0;
```
jbonofre, line 87: `deliveryDelay` is not used in `send()`. 

It means that it's not effective.

To be effective, the delay needs to be applied during message dispatch (so either passing it through to the session send, or setting `AMQ_SCHEDULED_DELAY` on the message).

Also `ActiveMQProducer.setDeliveryDelay()` (for JMS 2.0 `JMSProducer`) is not updated and still throws `UnsupportedOperationException`.

I think it's better to have the PR atomic and consistent.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
```diff
@@ -219,6 +221,25 @@ protected void checkClosed() throws IllegalStateException {
...
      * @since 1.1
      */
+
+    @Override
+    public void setDeliveryDelay(long deliveryDelay) throws JMSException {
+        checkClosed();
+
+        // Jakarta 3.1 Compliance Guard
+        if (deliveryDelay < 0 && session.connection.isStrictCompliance()) {
+            throw new jakarta.jms.MessageFormatException("Delivery delay cannot be negative when strictCompliance is enabled.");
```
jbonofre, line 231: This is wrong imho.

Jakarta Messaging 3.1 spec (`MessageProducer.setDeliveryDelay`) doesn't mandate `MessageFormatException` for invalid arguments.

`MessageFormatException` is for message body/property type conversion issues (like in `setBody`).
An `IllegalArgumentException` or `JMSException` (generic) would be more appropriate.
The JMS Spec says nothing about negative delays being illegal, the behavior is undefined and it's up to us to decide the validation logic.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
```diff
@@ -219,6 +221,25 @@ protected void checkClosed() throws IllegalStateException {
      * @see jakarta.jms.Session#createProducer
      * @since 1.1
      */
+
+    @Override
+    public void setDeliveryDelay(long deliveryDelay) throws JMSException {
```
jbonofre, line 226: Rather than defining here, why not updated `ActiveMQMessageProducerSupport` ?

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
jbonofre, line 87: Do we need really `protected` here ?
As we have getter/setter access, we can go with `private` and using mock in the test, right ?

mattrpav:
Updates to DeliveryDelay work should be applied to the pre-existing WIP PR: [#1157]

pradeep85841:
Closing this in favor of consolidating the Delivery Delay work into #1157 as requested by @mattrpav. I will move the Jakarta 3.1 compliance and 'effective delay' logic there.

PR #1815, opened 2026-03-20
[AMQ-8320] Implement DeliveryDelay and Jakarta 3.1 compliance

pradeep85841:
- Consolidated DeliveryDelay state into ActiveMQMessageProducerSupport for better encapsulation.
- Implemented Jakarta 3.1 strictCompliance check to reject negative delays when enabled.
- Fixed unreachable code and incorrect return types in ActiveMQProducer (JMS 2.0 wrapper).
- Ensured delivery delay is effective by injecting AMQ_SCHEDULED_DELAY and JMSDeliveryTime properties during send.
- Updated ActiveMQJMS2ContextTest to reflect that DeliveryDelay is now a supported feature.
- Added functional tests for strict vs legacy behaviour.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
@@ -27,9 +27,9 @@ import jakarta.jms.InvalidDestinationException;
 import jakarta.jms.JMSException;
 import jakarta.jms.Message;
 
-import org.apache.activemq.command.ActiveMQMessage;
-import org.apache.activemq.command.ActiveMQDestination;
 import org.apache.activemq.command.ProducerAck;
+import org.apache.activemq.command.ActiveMQDestination;
+import org.apache.activemq.command.ActiveMQMessage;
 import org.apache.activemq.command.ProducerId;
 import org.apache.activemq.command.ProducerInfo;
 import org.apache.activemq.management.JMSProducerStatsImpl;
@@ -327,12 +327,33 @@ public class ActiveMQMessageProducer extends ActiveMQMessageProducerSupport impl
             }
         }
 
-        long delay = getDeliveryDelay();
-        if (delay > 0) {
-            message.setLongProperty("AMQ_SCHEDULED_DELAY", delay);
-            long deliveryTime = System.currentTimeMillis() + delay;
-            message.setLongProperty(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY, deliveryTime);
+        long timeStamp = System.currentTimeMillis();
+        long delay = 0;
+
+        // Try to get delay from message property first (set by Wrapper)
+        Object delayProp = message.getObjectProperty(ScheduledMessage.AMQ_SCHEDULED_DELAY);
+        if (delayProp instanceof Number) {
+            delay = ((Number) delayProp).longValue();
         }
+
+        // If not on message, use the Producer's own setting (set by MessageProducer API)
+        if (delay <= 0) {
+            delay = getDeliveryDelay();
+        }
+
+        // If a delay exists, make sure it's on the message so the Broker schedules it!
+        if (delay > 0) {
+            // Ensure properties are writable before injecting the broker delay
+            if (message instanceof ActiveMQMessage) {
+                ((ActiveMQMessage)message).setReadOnlyProperties(false);
+            }
+            message.setLongProperty(ScheduledMessage.AMQ_SCHEDULED_DELAY, delay);
+        }
+
+        if (!disableMessageTimestamp) {
+            message.setJMSTimestamp(timeStamp);
+        }
+        message.setJMSDeliveryTime(timeStamp + delay);
 
         this.session.send(this, dest, message, deliveryMode, priority, timeToLive, disableMessageID, disableMessageTimestamp, producerWindow, sendTimeout, onComplete);
 
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducerSupport.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducerSupport.java
@@ -59,9 +59,11 @@ public abstract class ActiveMQMessageProducerSupport implements MessageProducer,
     @Override
     public void setDeliveryDelay(long deliveryDelay) throws JMSException {
         checkClosed();
-        // This should now compile after the rebase!
-        if (deliveryDelay < 0 && session.connection.isStrictCompliance()) {
-            throw new jakarta.jms.JMSException("Delivery delay cannot be negative.");
+        if (deliveryDelay < 0) {
+            if (session.connection.isStrictCompliance()) {
+                throw new jakarta.jms.JMSException("Delivery delay cannot be negative.");
+            }
+            deliveryDelay = 0; // Clamping for legacy
         }
         this.deliveryDelay = deliveryDelay;
     }
@@ -75,7 +77,6 @@ public abstract class ActiveMQMessageProducerSupport implements MessageProducer,
      */
     @Override
     public long getDeliveryDelay() throws JMSException {
-        checkClosed();
         return this.deliveryDelay;
     }
     
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
@@ -92,11 +92,9 @@ public class ActiveMQProducer implements JMSProducer {
                 }
             }
 
-            // [AMQ-8320] Producer setting for deliveryDelay will override user-specified ActiveMQ Scheduled Delay property
             if(this.deliveryDelay != null) {
-                long deliveryTimeMillis = System.currentTimeMillis() + this.deliveryDelay;
-                message.setLongProperty(ScheduledMessage.AMQ_SCHEDULED_DELAY, this.deliveryDelay);
-                message.setLongProperty(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY, deliveryTimeMillis);
+                // Explicitly set as long to avoid any Object/Long wrapper confusion
+                message.setLongProperty(ScheduledMessage.AMQ_SCHEDULED_DELAY, this.deliveryDelay.longValue());
             }
 
             activemqMessageProducer.send(destination, message, getDeliveryMode(), getPriority(), getTimeToLive(), getDisableMessageID(), getDisableMessageTimestamp(), null);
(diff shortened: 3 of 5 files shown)


jbonofre:
@pradeep85841 can you please upgrade to PR title ?

activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
```diff
@@ -246,12 +255,23 @@ public long getTimeToLive() {
...
+            // JMS 2.0 requires converting checked exceptions to RuntimeExceptions
+            throw JMSExceptionSupport.convertToJMSRuntimeException(e);
+        }
+        return this;
     }
 
     @Override
     public long getDeliveryDelay() {
-        throw new UnsupportedOperationException("getDeliveryDelay() is not supported");
+        return this.deliveryDelay;
```
jeanouii: Shouldn't this be implemented in a similar way as getTimeToLive() or getPriority() to avoid the NPE because of the unboxing of the Long?

activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
```diff
@@ -90,6 +92,13 @@ public JMSProducer send(Destination destination, Message message) {
                 }
             }
 
+            // [AMQ-8320] Producer setting for deliveryDelay will override user-specified ActiveMQ Scheduled Delay property
+            if(this.deliveryDelay != null) {
+                long deliveryTimeMillis = System.currentTimeMillis() + this.deliveryDelay;
```
jeanouii: I don't think that is really necessary here. 
For JMS 2+, the path is ActiveMQProducer.send() → ActiveMQMessageProducer.send() → ActiveMQSession.send().

So applying on ActiveMQMessageProducer should cover both JMS 1 and 2+
For now, we set a value here and we override it with a different value in ActiveMQMessageProducer. 

Not a big deal functionally, but having the logic duplicated will make maintenance harder in the future and may introduce discrepancies.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
```diff
@@ -326,6 +327,13 @@ public void send(Destination destination, Message message, int deliveryMode, int
             }
         }
 
+        long delay = getDeliveryDelay();
+        if (delay > 0) {
+            message.setLongProperty("AMQ_SCHEDULED_DELAY", delay);
+            long deliveryTime = System.currentTimeMillis() + delay;
+            message.setLongProperty(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY, deliveryTime);
```
jeanouii: To be checked with ActiveMQSession which sets to now without delay, correct?

activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQDeliveryDelayTest.java
```diff
@@ -0,0 +1,92 @@
...
+
+            MessageProducer producer = sess.createProducer(sess.createQueue("TEST.EFFECTIVE"));
+            long delay = 5000L;
+            producer.setDeliveryDelay(delay);
+
+            ActiveMQMessage msg = (ActiveMQMessage) sess.createTextMessage("Hello");
+            producer.send(msg);
+
+            // Verify Broker-side scheduling property
+            assertEquals("Broker delay property missing",
```
jeanouii: Because of the session comment above, if you also check the header instead of the custom property, it might fail?

To be checked

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
```diff
@@ -326,6 +327,13 @@ public void send(Destination destination, Message message, int deliveryMode, int
             }
         }
 
+        long delay = getDeliveryDelay();
+        if (delay > 0) {
+            message.setLongProperty("AMQ_SCHEDULED_DELAY", delay);
```
jeanouii: Can you use the constante like bellow ?

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducerSupport.java
```diff
@@ -56,7 +58,12 @@ public ActiveMQMessageProducerSupport(ActiveMQSession session) {
      */
     @Override
     public void setDeliveryDelay(long deliveryDelay) throws JMSException {
-        throw new UnsupportedOperationException("setDeliveryDelay() is not supported");
+        checkClosed();
+        // This should now compile after the rebase!
```
jeanouii: If it does eventually compile, please remove this comment ;-)

activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQDeliveryDelayTest.java
```diff
@@ -0,0 +1,92 @@
...
+ */
+package org.apache.activemq;
+
+import jakarta.jms.Connection;
+import jakarta.jms.MessageProducer;
+import jakarta.jms.Session;
+import org.apache.activemq.command.ActiveMQMessage;
+import org.junit.Test;
+
+import static org.apache.activemq.command.DataStructureTestSupport.assertEquals;
```
jeanouii: Wrong one??

activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQDeliveryDelayTest.java
```diff
@@ -0,0 +1,92 @@
...
+            ActiveMQMessage msg = (ActiveMQMessage) sess.createTextMessage("Hello");
+            producer.send(msg);
+
+            // Verify Broker-side scheduling property
+            assertEquals("Broker delay property missing",
+                    delay, msg.getLongProperty("AMQ_SCHEDULED_DELAY"));
+
+            // Verify Consumer-side visibility property (matching #1157 logic)
+            assertTrue("JMSDeliveryTime property missing or incorrect",
+                    msg.getLongProperty(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY) >= System.currentTimeMillis() + delay - 100);
```
jeanouii: In the CI, this might randomly fail under load or when parallelism is used. 

You could do something like this instead

```
final  long before = System.currentTimeMillis();
  producer.send(msg);

...

  long deliveryTime = msg.getLongProperty(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY);
  assertTrue(deliveryTime >= before + delay);
  assertTrue(deliveryTime <= System.currentTimeMillis() + delay);
```

activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
jbonofre: That's correct and a blocker to me.

As this class declares `private Long deliveryDelay = null` (boxed), here you do auto-unboxes to `long`. 
When `deliveryDelay` is `null` (the default), this will throw `NullPointerException`.

As @jeanouii said, you have to use a similar approach as in `getTimeToLive()` or `getPriority()` methods.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
jbonofre: We have something weird here. As `AMQ_SCHEDULED_DELAY` and `JMSDeliveryTime` properties are set in both `ActiveMQProducer.send()` and `ActiveMQMessageProducer.send()`.

As `ActiveMQProducer.send()` delegates to `ActiveMQMessageProducer.send()`, the properties get set twice with potentially different `System.currentTimeMillis()` values.

Imho, the logic should be in `ActiveMQMessageProducer.send()` only.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducer.java
jbonofre: Agree, to be consistent with the other parts of this class, we should use a constant here.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducerSupport.java
```diff
@@ -56,7 +58,12 @@ public ActiveMQMessageProducerSupport(ActiveMQSession session) {
...
+        if (deliveryDelay < 0 && session.connection.isStrictCompliance()) {
```
jbonofre: The check is correct to validate the parameter.
However, the validation only rejects negative values when `strictCompliance` is `true`.
When `strictCompliance` is `false`, negative delays are silently accepted but will produce nonsensical `AMQ_SCHEDULED_DELAY` values (negative scheduling).

I suggest:
1. Always rejecting negative values (JMS spec says delay must be >= 0) 
or
2. Clamping to 0

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageConsumer.java
```diff
@@ -636,6 +636,9 @@ public void execute() throws Exception {
                 }
             });
         }
+        if (m.propertyExists(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY)) {
+            m.setJMSDeliveryTime(m.getLongProperty(ActiveMQMessage.JMS_DELIVERY_TIME_PROPERTY));
```
jbonofre, line 640: Here, `JMSDeliveryTime` is restored from a message property.
This needs verification to ensure the consumer-side restoration isn't overwritten.

activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQDeliveryDelayTest.java
jbonofre: I think so 😄

activemq-unit-tests/src/test/java/org/apache/activemq/ActiveMQDeliveryDelayTest.java
jbonofre: Yes, it smells flakiness here 😄

mattrpav:
@pradeep85841 In the future, please add your changes to the existing PR (#982) (PR-on-the-PR). This PR is not carrying forward the discussion and feedback from the previous PR.

Key feedback:

> Since message objects can be reused/resent, and might already have an existing value for JMSDeliveryTime, it should always be setting the JMSDeliveryTime value on send even if there isnt a delay, to ensure the JMSDeliveryTime value on the message after send reflects behaviour of that send and not some previous send or receive.

> That doesnt mean it has to transit a delivery time on the wire...ideally it wouldnt if there is no delay being requested for this send...just that after send the local message object should report the delivery time for that send. Elsewhere I added an internal message method to set the value and note whether it is for transmission.

> It doesnt feel like it should be setting a message property at all, only a header. I don't expect the TCK tests which check additional properties arent being added to the messages are using delivery-delay at the time..but they could. I know the older 'message property extension' scheduling stuff uses one, but its not clear to me this needs to. They could just be treated as separate mechanisms, which they kind of are already as you have to choose to respect the new value on the producer or the old one on the message. Theres no reason the broker couldnt do the same, and this new mechanism essentially ignore that old prop.

PR #1824, opened 2026-03-23
[#1823] Align DestinationView and DestinationViewMBean

mattrpav:


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/DestinationView.java
```diff
@@ -136,6 +136,12 @@ public long getStoreMessageSize() {
         return messageStore != null ? messageStore.getMessageStoreStatistics().getMessageSize().getTotalSize() : 0;
     }
 
+    @Override
+    public long getMessagesCachedCount() {
+        return destination.getDestinationStatistics().getMessagesCached().getCount();
+    }
+
+    @Deprecated(forRemoval = true) // see: getMessagesCachedCount() instead. This method name is inconsistent
```
jbonofre, line 144: I agree with `@Deprecated` here. For consistency, I suggest to also deprecated `getMessagesCached()` method in the MBean.

mattrpav: Currently, there is no 'getMessagesCached()' method. This PR adds the new field method with aligned naming

```
public long getMessagesCachedCount();
```

jbonofre: Oh you renamed the MBean operation. It makes sense. Thanks.

PR #1829, opened 2026-03-25
feat(#1828): Expose NetworkConnector URI and local URI in JMX MBean

### LINKED ISSUE #1828: NetworkConnectorMBean should provide configured URI
If we have the `NetworkBridgeMBean` that provides details about the connection (remote and local addresses), we should be able to retrieve a network connector configuration about URI (remote and local).
If the `NetworkConnectorMBean` currently provides details about `name`, `duplex`, etc, it doesn't provide the URI.

jbonofre:
Add getUri() and getLocalUri() attributes to NetworkConnectorViewMBean so that the network connector URI is visible via JMX. Add abstract getUri() to NetworkConnector base class and implement it in MulticastNetworkConnector.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/network/MulticastNetworkConnector.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/network/MulticastNetworkConnector.java
@@ -77,6 +77,11 @@ public class MulticastNetworkConnector extends NetworkConnector {
         this.remoteTransport = remoteTransport;
     }
 
+    @Override
+    public URI getUri() {
+        return remoteURI;
+    }
+
     public URI getRemoteURI() {
         return remoteURI;
     }
--- a/activemq-broker/src/main/java/org/apache/activemq/network/MulticastNetworkConnector.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/network/MulticastNetworkConnector.java
@@ -77,9 +77,10 @@ public class MulticastNetworkConnector extends NetworkConnector {
         this.remoteTransport = remoteTransport;
     }
 
+    // Satisfies the abstract getUri() contract from NetworkConnector; delegates to getRemoteURI()
     @Override
     public URI getUri() {
-        return remoteURI;
+        return getRemoteURI();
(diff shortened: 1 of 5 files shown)


activemq-broker/src/main/java/org/apache/activemq/network/MulticastNetworkConnector.java
```diff
@@ -77,6 +77,11 @@ public void setRemoteTransport(Transport remoteTransport) {
         this.remoteTransport = remoteTransport;
     }
 
+    @Override
+    public URI getUri() {
+        return remoteURI;
```
jbonofre: I know it's redundant with `getRemoteURI()` here, but it's for consistency.

I'm fine to use `getRemoveURI()` everywhere if it makes more sense.

jeanouii: It's a bit weird indeed. Maybe add a small comment for us, so no one goes in one month from now and says, it's duplicate I'll remove it

jeanouii (dismissed):

activemq-broker/src/main/java/org/apache/activemq/network/MulticastNetworkConnector.java
mattrpav: I think it'd make more sense to make MulticastNetworkConnector have a 'getUri()' method instead.

I'm pretty against doing any field name mapping at the admin view b/c it is miscommunicating to users

jbonofre: OK, I was hesitating. Thanks for your inputs, let's use `getRemoteURI()` then.

jbonofre: I addressed your comment.

jbonofre:
@mattrpav @jeanouii I addressed your comments.

mattrpav (approved):
LGTM

PR #1847, opened 2026-03-26
Also validate nested composite URIs used with BrokerView

cshannon:
Add a check for VM transports that are in a nested composite URI

This is a follow on to #1840

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
@@ -604,21 +604,32 @@ public class BrokerView implements BrokerViewMBean {
 	}
 
     private static void validateAllowedUrl(String uriString) throws URISyntaxException {
-        validateAllowedUri(new URI(uriString));
+        validateAllowedUri(new URI(uriString), 0);
     }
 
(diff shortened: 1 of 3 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
```diff
@@ -603,18 +603,22 @@ public long getTotalMaxUncommittedExceededCount() {
...
 
-        // If composite, also check all schemes for each component
+        // If composite, iterate and check each of the composite URIs
         if (URISupport.isCompositeURI(uri)) {
             URISupport.CompositeData data = URISupport.parseComposite(uri);
             for (URI component : data.getComponents()) {
-                validateAllowedScheme(component.getScheme());
+                // Each URI could be a nested composite URI so
+                // call validateAllowedUri() to validate it
+                validateAllowedUri(component);
```
tabish121: Need to create some limit on the recursion depth here otherwise bad input could cause an overflow

cshannon: Good catch, I will limit it to something reasonable like 5 (although there's probably not much of a use case past 3 at the moment).

PR #1852, opened 2026-03-28
(Fix/test) flaky testForceExpirationZeroOverrideDLQ in MessageInterceptorStrategyTest

gurpartap3697:
The MessageInterceptorStrategyTest.testForceExpirationZeroOverrideDLQ test has been a source of flakiness, occasionally failing on slower CI environments with the error `MessageInterceptorStrategyTest.testForceExpirationZeroOverrideDLQ:231 expected:<1> but was:<0>`

ref: https://github.com/apache/activemq/actions/runs/23632687744/job/68835701471

The test relied on Thread.sleep(250L) to wait for a message to be routed to the Dead Letter Queue (DLQ). If a machine was under execution load or CPU starvation, 250ms was not enough time for the broker to process and dispatch the message.

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/broker/policy/MessageInterceptorStrategyTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/broker/policy/MessageInterceptorStrategyTest.java
@@ -222,6 +222,7 @@ public class MessageInterceptorStrategyTest extends TestSupport {
                 return originalQueueEmpty && dlqHasMessage;
             }
         });
+        assertTrue("Message did not reach DLQ in time", reachedDlq);
 
         queueBrowser = session.createBrowser(queue);
         Enumeration<?> browseEnumeration = queueBrowser.getEnumeration();


activemq-unit-tests/src/test/java/org/apache/activemq/broker/policy/MessageInterceptorStrategyTest.java
```diff
@@ -207,7 +208,16 @@ public void testForceExpirationZeroOverrideDLQ() throws Exception {
         Message sendMessageP = session.createTextMessage("expiration=zero-no-dlq-expiry");
         producer.send(queue, sendMessageP);
 
-        Thread.sleep(250l);
+        boolean reachedDlq = Wait.waitFor(new Wait.Condition() {
+            @Override
+            public boolean isSatisified() throws Exception {
+                boolean originalQueueEmpty = !session.createBrowser(queue).getEnumeration().hasMoreElements();
+                QueueBrowser dlqQueueBrowser = session.createBrowser(createQueue("mis.forceExpiration.zero-no-dlq-expiry.dlq"));
```
jbonofre: You create a browser here, but you never close it. With a default wait of 30 seconds with 1-second polling, it means it could create 60 unclosed `QueueBrowser` instances.

You should use try-with-resources or explicity `close()` in the iteration. 

That's not a huge blocker here because we are in a test, but worth to address.

activemq-unit-tests/src/test/java/org/apache/activemq/broker/policy/MessageInterceptorStrategyTest.java
```diff
@@ -207,7 +208,16 @@ public void testForceExpirationZeroOverrideDLQ() throws Exception {
...
+                boolean dlqHasMessage = dlqQueueBrowser.getEnumeration().hasMoreElements();
+                return originalQueueEmpty && dlqHasMessage;
+            }
+        });
+        assertTrue("Message should be routed to DLQ", reachedDlq);
 
         queueBrowser = session.createBrowser(queue);
```
jbonofre, line 227: Nit: you are doing the verification in the `Wait.watiFor()` and you are doing the verification again here. It's useless: if you are out of the iteration, you have the condition satisfied.

gurpartap3697:
Thanks @jbonofre! for the review. I've updated the `Wait.waitFor()` loop as well as the final assertions to use try-with-resources.

activemq-unit-tests/src/test/java/org/apache/activemq/broker/policy/MessageInterceptorStrategyTest.java
```diff
@@ -207,7 +208,20 @@ public void testForceExpirationZeroOverrideDLQ() throws Exception {
         Message sendMessageP = session.createTextMessage("expiration=zero-no-dlq-expiry");
         producer.send(queue, sendMessageP);
 
-        Thread.sleep(250l);
+        boolean reachedDlq = Wait.waitFor(new Wait.Condition() {
```
jbonofre, line 211: `reachedDlq` result is unused.

If the condition times out (return `false` after 30s), the test will silently proceed and likely fail.

I would add an assertion to prevent that:

```
assertTrue("Message did not reach DLQ in time", reachedDlq);
```

gurpartap3697: Added the missing `assertTrue("Message did not reach DLQ in time", reachedDlq);`

PR #1853, opened 2026-03-28
Enforce provider-only JMSX properties under strictCompliance

pradeep85841:
- Throws a MessageNotWriteableException if a client attempts to set JMSXDeliveryCount, JMSXRcvTimestamp, JMSXState, JMSXProducerTXID, or JMSXConsumerTXID.
- Client-set properties (like JMSXGroupID and JMSXGroupSeq) are completely bypassed and function normally.
- Added StrictComplianceProviderJMSXPropertyTest using an embedded BrokerService to explicitly verify both strict and legacy behaviors.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
@@ -18,7 +18,16 @@ package org.apache.activemq.command;
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
-import java.util.*;
+
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.Enumeration;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.Vector;
 
 import jakarta.jms.DeliveryMode;
 import jakarta.jms.Destination;
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
@@ -510,7 +510,7 @@ public class ActiveMQMessage extends Message implements org.apache.activemq.Mess
                     "JMSXState".equals(name) ||
                     "JMSXProducerTXID".equals(name) ||
                     "JMSXConsumerTXID".equals(name)) {
-                throw new MessageNotWriteableException("Provider-set JMSX property '" + name + "' cannot be set by a client under strict compliance.");
+                throw new JMSException("Provider-set JMSX property '" + name + "' cannot be set by a client under strict compliance.");
             }
         }
 
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
@@ -501,6 +501,19 @@ public class ActiveMQMessage extends Message implements org.apache.activemq.Mess
 
         checkValidObject(value);
         value = convertScheduled(name, value);
+
+        // Strict Compliance Check For Provider-Set JMSX Properties
+        ActiveMQConnection conn = getConnection();
+        if (conn != null && conn.isStrictCompliance()) {
+            if ("JMSXDeliveryCount".equals(name) ||
+                    "JMSXRcvTimestamp".equals(name) ||
+                    "JMSXState".equals(name) ||
+                    "JMSXProducerTXID".equals(name) ||
+                    "JMSXConsumerTXID".equals(name)) {
+                throw new MessageNotWriteableException("Provider-set JMSX property '" + name + "' cannot be set by a client under strict compliance.");
+            }
+        }
+
         PropertySetter setter = JMS_PROPERTY_SETERS.get(name);
 
         if (setter != null && value != null) {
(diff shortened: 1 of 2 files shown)


activemq-unit-tests/src/test/java/org/apache/activemq/StrictComplianceProviderJMSXPropertyTest.java
```diff
@@ -0,0 +1,138 @@
...
+        ActiveMQConnectionFactory factory = new ActiveMQConnectionFactory(connectionUri);
+
+        // Legacy mode (default)
+        factory.setStrictCompliance(false);
+
+        try (Connection connection = factory.createConnection();
+             Session session = connection.createSession(false, Session.AUTO_ACKNOWLEDGE)) {
+
+            connection.start();
+            Message message = (Message) session.createMessage();
```
jbonofre: I don't see `Message` in the import, I think it won't compile 😄

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
```diff
@@ -501,6 +501,19 @@ public void setObjectProperty(String name, Object value, boolean checkReadOnly)
...
+
+        // Strict Compliance Check For Provider-Set JMSX Properties
+        ActiveMQConnection conn = getConnection();
+        if (conn != null && conn.isStrictCompliance()) {
+            if ("JMSXDeliveryCount".equals(name) ||
+                    "JMSXRcvTimestamp".equals(name) ||
+                    "JMSXState".equals(name) ||
+                    "JMSXProducerTXID".equals(name) ||
+                    "JMSXConsumerTXID".equals(name)) {
+                throw new MessageNotWriteableException("Provider-set JMSX property '" + name + "' cannot be set by a client under strict compliance.");
```
jbonofre, line 523: I'm not sure we should use `MessageNotWriteableException` here.

`MessageNotWriiteableException` is specified in JMS for when a message is in read-only mode (received). For a provider-only property that clients should never set, I think a `JMSException` with a clear message might be more semantically accurate. 

The distinction matters because a client catching `MessageNotWriteableException` might assume calling `clearProperties()` would fix it, but it won't (since this is a permanent restriction, not a state-based one).

jbonofre (changes requested):

pradeep85841:
Hi @jbonofre, just a quick ping on this! The changes are pushed with JMSException and the corrected imports. Let me know if anything else is needed

jbonofre:
@pradeep85841 thanks ! Let me do a new pass.

pradeep85841:
Updated the "Wait.waitFor" condition in "testNetworkAdvancedStatistics". The test now waits for the broker to finish updating its background metrics (like enqueues and dequeues) before doing the final checks.

pradeep85841:
​Hi @jbonofre , just checking in! Both the main PR updates (exception/imports) and the fix for the flaky CI test are pushed. Let me know if you can kick off the CI checks again!

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
```diff
@@ -501,6 +501,19 @@ public void setObjectProperty(String name, Object value, boolean checkReadOnly)
 
         checkValidObject(value);
         value = convertScheduled(name, value);
+
+        // Strict Compliance Check For Provider-Set JMSX Properties
+        ActiveMQConnection conn = getConnection();
+        if (conn != null && conn.isStrictCompliance()) {
+            if ("JMSXDeliveryCount".equals(name) ||
```
mattrpav: I think a constant final public static set would make more sense here vs each property listed out by string name.

mattrpav:
Is the network metrics fix commit already merged? That should probably be a separate PR that can be merged quickly

pradeep85841:
Hi @mattrpav, I've split the flaky network metrics test fix into its own PR here: #1960
And updated this branch to use a public static final Set for the property checks as you suggested.

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
```diff
@@ -18,11 +18,7 @@
 
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
-import java.util.Enumeration;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Vector;
+import java.util.*;
```
mattrpav: Please don't reduce imports to wildcard. It is a common IDE setting, but we don't have the code style rules applied to catch it.

pradeep85841:
@mattrpav, My IDE auto-collapsed them when I added the Set classes. I've expanded them back to explicit imports and updated my local settings so it won't happen again. The fix is pushed

activemq-client/src/main/java/org/apache/activemq/command/ActiveMQMessage.java
```diff
@@ -501,6 +515,15 @@ public void setObjectProperty(String name, Object value, boolean checkReadOnly)
 
         checkValidObject(value);
         value = convertScheduled(name, value);
+
+        // Strict Compliance Check For Provider-Set JMSX Properties
+        ActiveMQConnection conn = getConnection();
+        if (conn != null && conn.isStrictCompliance()) {
+            if (STRICT_PROVIDER_JMSX_PROPERTIES.contains(name)) {
+                throw new JMSException("Provider-set JMSX property '" + name + "' cannot be set by a client under strict compliance.");
```
mattrpav: Please update this exception to throw MessageNotWriteableException to be spec-aligned.

mattrpav (changes requested):

pradeep85841:
Hi @mattrpav, I've updated the exception to MessageNotWriteableException.

Issue #1858, opened 2026-03-30
Fix AMQP race condition with concurrent dispatch and copy of messages


cshannon:
As discussed in detail in https://github.com/apache/activemq/pull/1851, there is a race condition with AMQP where text messages that are published to a topic can be delivered with a null body when sent over a network bridge and local consumer. 

The root cause is the that ActiveMQTextMessage can be mutated during copy and marshaling which can lead the the body becoming null as the message is optimized to only store the data in one format (binary or String). The messages are not supposed to be accessed by multiple threads and copied first, but with topics this can be violated to avoid having a ton of copies in memory and can lead to the issue.

The long term solution is to improve the thread safety of either the messages or how the messages are used in the broker. This a longer term goal and there will be a follow on issue to discuss and implement improvements.

The short term solution as a quick fix is to ensure the AMQP protocol converter marshals the message before passing it along to the broker. This brings the behavior inline with OpenWire (non VM at least) because the message will be already marshaled so when the broker later goes to make copies or sends to consumers there won't be an issue as no mutation will be attempted as the data is already marshaled.

PR #1860, opened 2026-03-30
[6.2.x] Ensure AMQP protocol marshals messages before passing to broker (#1859)

cshannon:
This change brings the behaior inline with other protocols and will prevent a future race condition during copy/dispatch as the data will already be marshaled.

This closes https://github.com/apache/activemq/issues/1858

(cherry picked from commit e8f10dd24a26a04892fe9d7b10fd697a2e138d19)


cshannon:
This can be merged once the pending 6.2.3 release is done and we bump the version as this targets 6.2.4

PR #1862, opened 2026-03-30
Fix flaky XACompletionTest by adding wait for async states

gurpartap3697:
XACompletionTest.testStatsAndBrowseAfterAckPreparedRolledback was failing intermittently:
```
Error:    XACompletionTest.testStatsAndBrowseAfterAckPreparedRolledback:254 size 0 expected:<0> but was:<10>
```
ref: https://github.com/apache/activemq/actions/runs/23623154131/job/68806975829

The test relied on immediate JMX state, which is asynchronous and can lead to flakiness.

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/store/jdbc/XACompletionTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/store/jdbc/XACompletionTest.java
@@ -249,7 +249,7 @@ public class XACompletionTest extends TestSupport {
         resource.recover(XAResource.TMSTARTRSCAN);
         resource.recover(XAResource.TMNOFLAGS);
 
-        Wait.waitFor(() -> proxy.getInFlightCount() == 0L && proxy.cursorSize() == 0);
+        assertTrue("Timed out waiting for queue state to stabilize", Wait.waitFor(() -> proxy.getInFlightCount() == 0L && proxy.getQueueSize() == 10L && proxy.cursorSize() == 0));
         assertEquals("prefetch 0", 0, proxy.getInFlightCount());
         assertEquals("size 0", 10, proxy.getQueueSize());
         assertEquals("size 0", 0, proxy.cursorSize());


activemq-unit-tests/src/test/java/org/apache/activemq/store/jdbc/XACompletionTest.java
```diff
@@ -249,6 +249,7 @@ public void testStatsAndBrowseAfterAckPreparedRolledback() throws Exception {
         resource.recover(XAResource.TMSTARTRSCAN);
         resource.recover(XAResource.TMNOFLAGS);
 
+        Wait.waitFor(() -> proxy.getInFlightCount() == 0L && proxy.cursorSize() == 0);
```
jbonofre: I suggest to wrap `Watit.waitFor()` in a `assertTrue` for better diagnostics on failure:


```suggestion
        assertTrue("Timed out waiting for inFlight and cursor to clear",                                                                                                                                                                            
          Wait.waitFor(() -> proxy.getInFlightCount() == 0L && proxy.cursorSize() == 0)); 
```

jbonofre: The wait condition doesn't cover `getQueueSize()`. The wait checks `getInFlightCount() == 0 && cursorSize() == 0`, but the very next assertion also checks `getQueueSize() == 10`.

If `getQueueSize()` is also subject to async updates, it could still flake. 

Worth considering whether to include it in the wait condition.

gurpartap3697: Good point — thanks for calling that out.

I initially assumed that inFlightCount == 0 would imply the queue state (including queueSize) had fully settled. I've updated the wait condition to include getQueueSize()

gurpartap3697:
Thanks @jbonofre for feedback! I updated the pr with suggested changes.

jbonofre:
@gurpartap3697 good job ! I will merge as soon as CI is green.

PR #1865, opened 2026-04-01
Fix network bridge silent death when transport exception occurs during broker info handshake

### LINKED ISSUE #1864: Network bridge silently dies without reconnection when transport exception occurs during broker info handshake
## Description

In `DemandForwardingBridgeSupport.start()`, the local and remote transport `onException()` handlers check `if (!futureBrokerInfo.isDone())`. When the future is NOT done (i.e., the broker info handshake has not yet completed), the handler cancels the future and **returns early** — skipping `serviceLocalException()` / `serviceRemoteException()` entirely.

```java
// localBroker onException handler (~line 227)
public void onException(IOException error) {
    if (!futureLocalBrokerInfo.isDone()) {
        LOG.info("Error with pending local brokerInfo on: {} ({})", localBroker, error.getMessage());
        LOG.debug("Peer error: ", error);
        futureLocalBrokerInfo.cancel(true);
        return;   // <-- BUG: skips serviceLocalException()
    }
    serviceLocalException(error);
}
```

The same pattern exists in the `remoteBroker` `onException()` handler (~line 247).

## Impact

The reconnection chain — `serviceLocalException()`/`serviceRemoteException()` → `fireBridgeFailed()` → `NetworkBridgeListener.bridgeFailed()` → `discoveryAgent.serviceFailed()` — is never triggered. The network bridge silently dies and is never re-established. This requires a broker restart to recover.

## Evidence this is a bug

The `duplexInboundLocalBroker` `onException()` handler in the same file does NOT have this guard — it unconditionally calls `serviceLocalException(error)`, confirming the intended behavior.

## Fix

Remove the two `return` statements so that `serviceLocalException()` / `serviceRemoteException()` is always called regardless of the future's state. The future cancellation is still performed, but execution continues to trigger the reconnection path.

## Production Validation

Tested in a production environment with 15 network connectors over 3.5 months. The bug path was hit 12 times (transport exceptions during broker info handshake), and all 12 resulted in successful reconnection after the fix. Prior to the fix, these would have been permanent connector losses.

## Component

activemq-broker / Network Connectors

TylerMurali:
## Summary

Fixes #1864

- **Bug**: In `DemandForwardingBridgeSupport.start()`, the `onException()` handlers for both local and remote transports return early when `futureBrokerInfo` is not done, skipping `serviceLocalException()`/`serviceRemoteException()`. This prevents the reconnection chain (`fireBridgeFailed()` → `discoveryAgent.serviceFailed()`) from firing, causing the bridge to silently die without reconnection.
- **Fix**: Remove the two `return` statements so the exception service methods always fire. The future cancellation is preserved. This is a 2-line change.
- **Evidence**: The `duplexInboundLocalBroker` handler in the same file unconditionally calls `serviceLocalException()`, confirming the intended behavior.

## Production Validation

Tested in production with 15 network connectors over 3.5 months. The bug path was hit 12 times (transport exceptions during broker info handshake), and all 12 resulted in successful reconnection after the fix. Prior to the fix, these were permanent connector losses requiring broker restarts.

## Test Plan

- [x] `testBridgeReconnectsAfterHandshakeFailure` — **Reproduces the exact bug path**: uses a fake server socket that accepts TCP connections but never sends `BrokerInfo`, so `onException` fires while `futureBrokerInfo` is NOT done. **Fails without the fix, passes with it.**
- [x] `testBridgeReconnectsAfterRemoteBrokerRestart` — General reconnection test: single remote broker stop/restart cycle
- [x] Existing tests pass: `DemandForwardingBridgeTest`, `DemandForwardingBridgeSupportTest`, `NetworkReconnectTest`

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/network/NetworkBridgeReconnectOnHandshakeFailureTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/network/NetworkBridgeReconnectOnHandshakeFailureTest.java
@@ -19,8 +19,12 @@ package org.apache.activemq.network;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import java.net.ServerSocket;
+import java.net.Socket;
 import java.net.URI;
+import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import jakarta.jms.Connection;
 import jakarta.jms.MessageConsumer;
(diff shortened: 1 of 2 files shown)


activemq-unit-tests/src/test/java/org/apache/activemq/network/NetworkBridgeReconnectOnHandshakeFailureTest.java
```diff
@@ -0,0 +1,260 @@
...
+        // Close the accepted socket — this fires onException while
+        // futureBrokerInfo is NOT done (the exact bug path)
+        Socket s = clientSocket.get();
+        if (s != null) {
+            s.close();
+            LOG.info("Fake server closed client socket — triggering onException with futureBrokerInfo not done");
+        }
+
+        // Wait a bit for the bridge to process the exception and attempt reconnection
+        Thread.sleep(2000);
```
jbonofre: This is a little fragile here compared to `Wait.waitFor()` but acceptable.

mattrpav: Agreed— it is fragile and we should request it changed to the wait. 

Also, we probably need to call out not using Thread.sleep() in favor of Wait in the CONTRIBUTING.md going forward

TylerMurali: Re designed the test to avoid the Thread.sleep

activemq-unit-tests/src/test/java/org/apache/activemq/network/NetworkBridgeReconnectOnHandshakeFailureTest.java
```diff
@@ -0,0 +1,331 @@
...
+
+        for (int i = 0; i < remoteExceptions.size(); i++) {
+            Throwable ex = remoteExceptions.get(i);
+            LOG.info("serviceRemoteException call [{}]: {} ({})",
+                    i, ex.getClass().getName(), ex.getMessage());
+        }
+
+        boolean hasIOException = remoteExceptions.stream()
+                .anyMatch(ex -> ex instanceof IOException);
+        assertTrue(
```
cshannon, line 207: I am still investigating this, but the first observation is the test case here might need some improvement. I ran this test without the included fix and I tried commenting out the assertion here and the tests continued and both passed and the bridges seemed to reconnect so it's not really demonstrating the issue of silently not reconnecting as described from what I can tell.

cshannon:
So I am not saying there is not an issue here, but I'm trying to figure out what I am missing because the reconnect logic still seems to be fired even without this fix because the TimeoutException is still being passed along to bridge failed. Maybe there is a race condition. 

Here is what I am seeing when I run the test case and step through things without the fix:

1. EOFException does happen first and is caught here and [aborts](https://github.com/apache/activemq/blob/27c17441d681e2b325e5e0700220a7b0d302e780/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L231) as noted
2. However, after that happens the call to collectBrokerInfos() is called [here](https://github.com/apache/activemq/blob/27c17441d681e2b325e5e0700220a7b0d302e780/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L370).
3. collectBrokerInfos() triggers a timeout exception and is caught [here](https://github.com/apache/activemq/blob/27c17441d681e2b325e5e0700220a7b0d302e780/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L398)
4. bridgeFailedError(timeoutException) is called [here](https://github.com/apache/activemq/blob/27c17441d681e2b325e5e0700220a7b0d302e780/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L663) which passes in that timeout exception and called bridgeFailed() on the listener
5. discoveryServiceAgent.sericeFailed(event) is invoked [here](https://github.com/apache/activemq/blob/27c17441d681e2b325e5e0700220a7b0d302e780/activemq-broker/src/main/java/org/apache/activemq/network/DiscoveryNetworkConnector.java#L254) which in turn kicks off the reconnect logic

cshannon:
So digging in a bit more I see that when the exception happens it doesn't pass the IOException obviously, but `futureRemoteBrokerInfo.cancel(true); `is called which triggers the `TimeoutException`. That does get passed to `serviceRemoteException()` in `collectBrokerInfos()` which should trigger the reconnect.

So the part that I am confused about is your saying that returning early prevents fireBridgeFailed() from firing, but from what I can see returning early first calls cancel on the futureRemoteBrokerInfo which in turn cause the TimeoutException which is caught later and triggers fireBridgeFailed()

TylerMurali:
I might have reasoned is wrongly, It was an hard one to track down, used to happens once in a while across our hundreds of bridge installs. One customer with the network instability it was happening few times in a week bringing system down. I believe a combination of Inactivity monitor exception and failure to negotiate the broker info after SSL is established and FutureBrokerInfo object state are contributing for not triggering a restart on the bridge connector. Throwing the ServiceException() in a more deterministic way has fixed the issue.

cshannon:
Thanks that makes sense, I will see if i can dig into this more. It does seems like there's more to do this based on my initial testing so you might be right that it has to do with the Inactivity monitor. The change seams reasonable but it would be good to understand the scenario a bit better of how it can happen.

cshannon (approved):
I took a look at this again I think this fix is good. It makes sense to pass in the original exception to the error handler method. Canceling the future will also normally trigger a timeout exception and cause a reconnect as I previously pointed out, and this causes the reconnect to typically work.

However, taking a closer look, I think there is a race condition. What could happen is in between the time the check is done to see if the future was completed, and the actual call to cancel the future, the future might have been completed successfully so that when cancel gets called on the future it becomes a no op and the exception gets lost and is never passed to the exception handler.

With this change things are more deterministic in that we always guarantee the exception handler is called at least once with the original error. It's now possible (or even likely) that the serviceRemoteException() or serviceLocalException() could be called more than once (the timeout exception being the other time) but that is ok because there is a bridgeFailed AtomicBoolean that uses compare and set so it will only fire the failure logic at most once to reconnect and not duplicate. Also, the disposed flag gets checked too.

mattrpav:
@cshannon I’ve got another fix coming where a fast reconnect gets a failure due to clientId dupe. Might be good to look at these two together

cshannon:
> @cshannon I’ve got another fix coming where a fast reconnect gets a failure due to clientId dupe. Might be good to look at these two together

It can just be looked at separate, this was already merged and I think this fix is good either way

PR #1900, opened 2026-04-10
Minor bug fix for BrokerView#validateAllowedUri

cshannon:
The wrong variable is being referenced in the nested loop. This does not cause the validation to actually break due to other checks done during the recursive call, but is incorrect either way. With this fix the counter needed tweaking for limiting the number of nested components as well.

Follow on to #1847


tabish121 (approved):

mattrpav (approved):
LGTM

jbonofre (approved):

PR #1906, opened 2026-04-11
Fix flaky DurableSubInBrokerNetworkTest by replacing fixed sleeps with Wait.waitFor polling

gurpartap3697:
### Description:
DurableSubInBrokerNetworkTest was intermittently failing in CI due to timing issues.  `setUp()` started the `DiscoveryNetworkConnector` and immediately let tests run, with no guarantee the bridge was established.
After creating/removing durable subscriptions, the tests slept 100ms and then checked state with a single synchronous query. On slow or loaded CI machines, 100ms is insufficient for subscription changes to propagate across the broker network.  

### Error:
```
Error:    DurableSubInBrokerNetworkTest>CombinationTestSupport.runBare:113->CombinationTestSupport.runBare:107->testDurableSubNetwork:103
```
ref: https://github.com/apache/activemq/actions/runs/24242348177/job/70780252784

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/usecases/DurableSubInBrokerNetworkTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/usecases/DurableSubInBrokerNetworkTest.java
@@ -30,6 +30,7 @@ import org.apache.activemq.ActiveMQConnectionFactory;
 import org.apache.activemq.network.DiscoveryNetworkConnector;
 import org.apache.activemq.network.NetworkConnector;
 import org.apache.activemq.network.NetworkTestSupport;
+import org.apache.activemq.util.Wait;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.junit.experimental.categories.Category;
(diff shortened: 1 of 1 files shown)


cshannon (changes requested):
Thanks for the test fix! The only suggestion I have after a quick look is I think it would be better to speed up the polling time. The default calls to Wait.waitFor() wait up to 30 seconds and retry every second. I think it would be better to speed up the checks quicker and reduce the 30 second time so we timeout faster if it will fail. The default of 1 second is something I've considered changing as a default to be faster anyways as we often override it.  Example:

```java
// retries every 100 millis and times out after 15 seconds
assertTrue("Network bridge did not establish in time",
    Wait.waitFor(() -> !nc.activeBridges().isEmpty(), 15000, 100));
```

It probably makes sense to apply that to all the Wait.waitFor() calls

gurpartap3697:
Thanks for the suggestion! I’ve updated the `Wait.waitFor()` calls to use a shorter polling interval.

jbonofre (approved):
LGTM, and thanks to have considered @cshannon suggestion.

cshannon (approved):

cshannon:
Should be good if CI passes tests

cshannon:
Tests keep failing (probably another flaky test) but going to try one more time anyways

jbonofre:
CI is failing on MQTT, so not on the DurableSubInBrokerNetworkTest.

PR #1910, opened 2026-04-13
Restrict URL protocol types loaded by XBeanBrokerFactory

### LINKED ISSUE #1899: Review options for ConnectionFactory configuration for allow/deny list of protocol schemes
3rd-parties using ActiveMQ for integration (SaaS, Cloud products, etc) would benefit from having a way to deny the creation of vm:// transport + remote loading of xml bean to limit security exposure.

cshannon:
This adds a new system property to control which protocol types are valid for loading resources using the XBeanBrokerFactory. By default only file and classpath resources can be loaded.

The goal of this is to prevent possible future security issues by hardening what is allowed to be loaded by default. There have been a lot of previous CVEs reported that were made possible by allowing remoting loading of Spring contexts (ie using http) so this will help stop any future vulnerabilities from being discovered but just no longer allowing that out of the box.

This PR does the following:

1. Restrict which URL protocols can be used to load resources by the `XBeanBrokerFactory` (and vm transport that uses it) with classpath and file allowed by default.
2. This adds a new property that users can set to control what is allowed: `org.apache.activemq.xbean.XBEAN_BROKER_FACTORY_PROTOCOLS`
3. The default is set to `classpath,file `so we only allow those resources. `*`can be used to allow all and empty string to allow none.
4. The default behavior of the order that resources are tried is preserved like before. A URL is always tired first as a file on the file system to see if it is found, then it checks if it's a URI, and lastly falls back to classpath. This still happens assuming the configuration allows it (ie file or classpath could be skipped if either or both are not allowed)
5. It loads correctly both file or classpath using either the fully qualified prefix or not (just like before)
6. I added a ton of unit tests, including full code coverage for the Utils class which does all the loading and validation.

Closes #1899

### DIFF
--- a/activemq-spring/src/main/java/org/apache/activemq/xbean/XBeanBrokerFactory.java
+++ b/activemq-spring/src/main/java/org/apache/activemq/xbean/XBeanBrokerFactory.java
@@ -17,23 +17,16 @@
 package org.apache.activemq.xbean;
 
 import java.beans.PropertyEditorManager;
-import java.io.IOException;
 import java.net.MalformedURLException;
 import java.net.URI;
 
 import java.util.Arrays;
-import java.util.List;
 import java.util.Set;
 import java.util.stream.Collectors;
-import org.apache.activemq.broker.BrokerContextAware;
 import org.apache.activemq.broker.BrokerFactoryHandler;
 import org.apache.activemq.broker.BrokerService;
 import org.apache.activemq.spring.SpringBrokerContext;
 import org.apache.activemq.spring.Utils;
-import org.apache.activemq.transport.stomp.FrameTranslator;
-import org.apache.activemq.transport.stomp.JmsFrameTranslator;
-import org.apache.activemq.transport.stomp.LegacyFrameTranslator;
-import org.apache.activemq.util.FactoryFinder;
 import org.apache.activemq.util.IntrospectionSupport;
 import org.apache.activemq.util.URISupport;
 import org.apache.xbean.spring.context.ResourceXmlApplicationContext;
@@ -44,7 +37,6 @@ import org.springframework.beans.BeansException;
 import org.springframework.beans.FatalBeanException;
 import org.springframework.beans.factory.xml.XmlBeanDefinitionReader;
 import org.springframework.context.ApplicationContext;
-import org.springframework.context.ApplicationContextAware;
 import org.springframework.core.io.Resource;
 
 /**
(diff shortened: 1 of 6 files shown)


mattrpav (dismissed):
LGTM

activemq-spring/src/main/java/org/apache/activemq/xbean/XBeanBrokerFactory.java
```diff
@@ -17,14 +17,23 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.Set;
+import java.util.stream.Collectors;
 import org.apache.activemq.broker.BrokerContextAware;
 import org.apache.activemq.broker.BrokerFactoryHandler;
 import org.apache.activemq.broker.BrokerService;
 import org.apache.activemq.spring.SpringBrokerContext;
 import org.apache.activemq.spring.Utils;
+import org.apache.activemq.transport.stomp.FrameTranslator;
```
jbonofre: It seems these imports are not actually used in the factory.

Maybe coming from previous dev ?

activemq-spring/src/main/java/org/apache/activemq/xbean/XBeanBrokerFactory.java
```diff
@@ -42,13 +51,33 @@
  * 
  */
 public class XBeanBrokerFactory implements BrokerFactoryHandler {
-    private static final transient Logger LOG = LoggerFactory.getLogger(XBeanBrokerFactory.class);
+    private static final Logger LOG = LoggerFactory.getLogger(XBeanBrokerFactory.class);
+
+    public static final String XBEAN_BROKER_FACTORY_PROTOCOLS_PROP =
+            "org.apache.activemq.xbean.XBEAN_BROKER_FACTORY_PROTOCOLS";
+    public static final String DEFAULT_ALLOWED_PROTOCOLS =
+            String.join(",", Set.of(Utils.FILE_PROTOCOL, Utils.CLASSPATH_PROTOCOL));
```
jbonofre: There's no guarantee of order here: it could be `file,classpath` or `classpath,file`.

Why not just using a simple `Utils.FILE_PROTOCOL + "," + Utils.CLASSPATH_PROTOCOL`.

jbonofre (dismissed):

activemq-spring/src/main/java/org/apache/activemq/xbean/XBeanBrokerFactory.java
cshannon: Order shouldn't matter but doing that is fine too and easier to read

activemq-spring/src/main/java/org/apache/activemq/xbean/XBeanBrokerFactory.java
cshannon: Probably, we really need a plugin to check imports (I've mentioned this before), it's too easy for imports to leak in by mistake when making a lot of changes and refactoring.

jbonofre (approved):

PR #1918, opened 2026-04-13
Add Http discovery transport to denied list for JMX

cshannon:
This also prevents the Http discovery transport from being added as a connector or network connector through JMX and Jolokia

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
@@ -44,6 +44,8 @@ public class BrokerView implements BrokerViewMBean {
 
     private static final Logger LOG = LoggerFactory.getLogger(BrokerView.class);
 
+    private static final Set<String> DENIED_TRANSPORT_SCHEMES = Set.of("vm", "http");
+
     ManagedRegionBroker broker;
 
     private final BrokerService brokerService;
@@ -402,7 +404,7 @@ public class BrokerView implements BrokerViewMBean {
 
     @Override
     public String addConnector(String discoveryAddress) throws Exception {
-        // Verify VM transport is not used
+        // Verify a denied transport scheme is not used
         validateAllowedUrl(discoveryAddress);
         TransportConnector connector = brokerService.addConnector(discoveryAddress);
         if (connector == null) {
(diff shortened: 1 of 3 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
```diff
@@ -44,6 +44,8 @@ public class BrokerView implements BrokerViewMBean {
 
     private static final Logger LOG = LoggerFactory.getLogger(BrokerView.class);
 
+    private static final Set<String> DENIED_TRANSPORT_SCHEMES = Set.of("vm", "http");
```
jbonofre, line 47: Should we consider `https` here?

activemq-unit-tests/src/test/java/org/apache/activemq/jmx/JmxCreateNCTest.java
```diff
@@ -82,35 +80,44 @@ public void testBridgeRegistration() throws Exception {
...
+            fail("Should have failed trying to add connector bridge");
         } catch (IllegalArgumentException e) {
-            assertEquals("VM scheme is not allowed", e.getMessage());
+            assertEquals("Transport scheme '" + scheme + "' is not allowed", e.getMessage());
         }
 
         try {
-            proxy.addNetworkConnector("multicast:(vm://localhost)");
+            proxy.addNetworkConnector("multicast:(" + scheme + "://localhost)");
             fail("Should have failed trying to add vm connector bridge");
```
jbonofre: nit: `vm` can be removed here (for consistency to other).


```suggestion
            fail("Should have failed trying to add connector bridge");
```

jbonofre (dismissed):

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
cshannon: Surprisingly (or maybe not given the age) there is no https discovery protocol, the broker only supports http for that.

There is http and https transports but those are not the same as the http discovery.

Here is the location for the http/https transport references: https://github.com/apache/activemq/tree/main/activemq-http/src/main/resources/META-INF/services/org/apache/activemq/transport

And here is the location for http discovery: https://github.com/apache/activemq/tree/main/activemq-http/src/main/resources/META-INF/services/org/apache/activemq/transport/discoveryagent

cshannon: A future enhancement could be to add https if people are using it

jbonofre (approved):

PR #1931, opened 2026-04-15
fix(webconsole): the webconsole now redirect to the slave.jsp when required

jbonofre:
When using `<broker ... startAsync="true">` in master/slave mode, the Jetty server can be started on slave.

In that case, the broker is not a "full active" broker yet, and the Web Console is poorly failing with an exception (and long stack trace in the log).

This PR fixes the Web Console, now dealing with slave broker and redirecting cleanly to `slave.jsp`.

NB: no impact on standalone/master broker, same behavior as today.


cshannon:
@mattrpav - Does this look ok to you? I think redirecting the page makes sense if it is running so the user knows.

cshannon (dismissed):

mattrpav (dismissed):
LGTM

cshannon (approved):

PR #1938, opened 2026-04-16
Queue browse improvements in webconsole

cshannon:
This change makes sure we always use the correct content type for the output based on the configured view and also escape xml content when displaying.

This version is different from #1937 because it uses our own method for XML escaping and it does not require adding new dependencies. We do use common-text as a testing only dependency just to verify correctness.


activemq-web/src/main/java/org/apache/activemq/web/util/ViewUtils.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.web.util;
+
+import com.rometools.rome.feed.atom.Link;
```
jbonofre, line 19: This import is not used.

I guess it was an IDE auto-completion 😄

activemq-web/src/main/java/org/apache/activemq/web/util/ViewUtils.java
```diff
@@ -0,0 +1,67 @@
...
+    public static final String AMP = "&";
+    public static final String QUOTE = "\"";
+    public static final String LT = "<";
+    public static final String GT = ">";
+    public static final String APOS = "'";
+
+    public static final String XML_ESCAPED_AMP = "&amp;";
+    public static final String XML_ESCAPED_QUOTE = "&quot;";
+    public static final String XML_ESCAPED_LT = "&lt;";
+    public static final String XMl_ESCAPED_GT = "&gt;";
```
jbonofre, line 36:
```suggestion
    public static final String XML_ESCAPED_GT = "&gt;";
```

jbonofre (dismissed):
LGTM. Just a couple of minor fixes to do.

jbonofre (approved):

PR #1950, opened 2026-04-21
Add remote file filtering for XBeanBrokerFactory

cshannon:
This expands on the previous work in #1910 to add treat remote files as a separate protocol category that can be controlled independently of regular files when loading resources. By default remote files will be disabled.

Follow on to #1910

### DIFF
--- a/activemq-spring/src/main/java/org/apache/activemq/spring/Utils.java
+++ b/activemq-spring/src/main/java/org/apache/activemq/spring/Utils.java
@@ -32,6 +32,9 @@ public class Utils {
 
     public static final String FILE_PROTOCOL = "file";
     public static final String CLASSPATH_PROTOCOL = "classpath";
+    // Special marker to indicate we want to allow remote files such as
+    // Windows UNC, etc
+    public static final String REMOTE_FILE_PROTOCOL = "remote-" + FILE_PROTOCOL;
 
     public static Resource resourceFromString(String uri) throws MalformedURLException {
         // default allows all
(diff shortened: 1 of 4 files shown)


activemq-spring/src/main/java/org/apache/activemq/spring/Utils.java
```diff
@@ -77,22 +80,51 @@ public static Resource resourceFromString(String uri, Set<String> allowedProtoco
...
+        return isQualifiedRemoteFile(uriString) ? REMOTE_FILE_PROTOCOL :
+                new URI(uriString).getScheme();
+    }
+
+    private static boolean isUnqualifiedRemoteFile(String uri) {
+        return uri.startsWith("//") || uri.startsWith("\\\\");
+    }
+
+    private static boolean isQualifiedRemoteFile(String uri) {
+        return uri.startsWith(FILE_PROTOCOL + "://") ||  uri.startsWith(FILE_PROTOCOL + ":\\\\");
```
jbonofre, line 128: I guess `file:///` match here, meaning it's local but not remove. Should we not check the `///` to be sure we are actually local?

```suggestion
      return (uri.startsWith(FILE_PROTOCOL + "://") && !uri.startsWith(FILE_PROTOCOL + ":///"))
          || uri.startsWith(FILE_PROTOCOL + ":\\\\");   
```

jbonofre (dismissed):
It looks good. I just have a doubt about `isQualifiedRemoteFile()` test.

activemq-spring/src/main/java/org/apache/activemq/spring/Utils.java
cshannon: That is a good point but I think the safest thing here is to just only allow using the single slash or relative path. No one needs to use `file:///` they can just use `file:/` and that will simplify things and prevent any bugs.

We can make a note of that when this is released and we write some documentation.

jbonofre: Yes, that works for me. Thanks!

mattrpav (approved):
LGTM

PR #1960, opened 2026-04-23
Fix flaky NetworkAdvancedStatisticsTest race condition

pradeep85841:
This was originally included in PR #1853, but I have split it out into this separate PR per @mattrpav review so it can be merged quickly.

updated the Wait.waitFor block so the test now waits for both the messages to arrive and the statistics to finish updating before running the final assertions.


mattrpav (approved):
LGTM

PR #1964, opened 2026-04-25
Fix for race condition when sending text messages using ajax to a network of borkers

arnoudja:
The race condition as described in https://github.com/apache/activemq/pull/1851 is solved for AMQP publishers but still there for AJAX publishers. This change should fix it for those as well.


arnoudja:
Is there additional information needed for this pr?

jbonofre:
I will do a new pass.

Issue #1816, opened 2026-03-20
Refactore and improve activemq CLI


jbonofre:
If we already have a CLI (`activemq`), I strongly believe we should improve the CLI with more "user facing" operations (similar to what we can find on the Web Console):

1. Listing all resources (destinations, transport connectors, network connectors, etc)
2. Managing resources (create/delete/update destinations, transport connectors, network connectors, etc)
3. Managing messages (browse/produce/consume)
4. Getting internal broker resources (scheduler, broker attributes, ...)

The purpose is not to "expose" the JMX MBeans (as we can do today with `activemq bstat`, `activemq dstat`, or `activemq query -Qqueue`, etc).
The purpose is to façade with user friendly CLI:

```
activemq queues create FOO
activemq queues delete FOO
activemq queues purge FOO
...
```

Why do I think it's important ?

Because we would be able to "integrate" easily ActiveMQ via the CLI into Claude Code and similar (no need to create a MCP server anymore if the CLI is good enough).

anmol-saxena-14:
Hi @jbonofre,

Can I work on this ?

Thanks

PR #1966, opened 2026-04-26
feat(console): add user-friendly CLI commands for queues, topics, broker, and network connectors(#1816).

anmol-saxena-14:
Adds four new action-oriented CLI commands as described in **Issue#1816**, making it straightforward to manage a running broker from the shell(and from AI agents such as Claude Code) without requiring JMX tooling.

**New commands**
    `activemq queues`   list | create | delete | purge | info | browse | produce | pause | resume
    `activemq topics`   list | create | delete | info
    `activemq broker`   info | connectors [list|add|remove] | scheduler
    `activemq network-connectors`  list | add | remove

Each command is registered via the existing ServiceLoader SPI `(META-INF/services/...Command)` so no existing code is modified to   discover them.

The consume action is intentionally not implemented in the new commands. Receiving and acknowledging messages requires an active JMS connection,which is outside the scope of the JMX-based CLI facade. The existing `activemq consumer` command covers that use case.

**Tests :** 37 new unit tests across `BrokerCommandTest`, `NetworkConnectorsCommandTest`, `QueuesCommandTest`, and `TopicsCommandTest` covering all actions and error-handling paths.


anmol-saxena-14:

activemq-console/src/test/java/org/apache/activemq/console/NetworkConnectorsCommandTest.java
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    @Test(timeout = 30000)
+    public void testNoActionShowsHelp() throws Exception {
+        String result = execute();
+        assertTrue("help content shown", result.contains("Actions:"));
+    }
+
+    @Test(timeout = 30000)
+    public void testAddMissingUriThrows() throws Exception {
```
jbonofre, line 94: This test can pass silently: the `try/catch` has no `fail()` after `execute("add")`, so if the `IllegalArgumentException` is never thrown the assertion in the catch block simply doesn't run and the test still passes. It would keep passing even if the validation were removed. Please add a `fail("expected IllegalArgumentException")` after the `execute(...)` call, or switch to `assertThrows(IllegalArgumentException.class, () -> execute("add"))`.

activemq-console/src/test/java/org/apache/activemq/console/NetworkConnectorsCommandTest.java
```diff
@@ -0,0 +1,128 @@
...
+        try {
+            execute("add");
+        } catch (IllegalArgumentException e) {
+            assertTrue(e.getMessage().contains("URI required"));
+        }
+    }
+
+    @Test(timeout = 30000)
+    public void testRemoveMissingNameThrows() throws Exception {
```
jbonofre, line 100: Same issue as `testAddMissingUriThrows`: no `fail()` after `execute("remove")`, so this passes even if no exception is thrown. Use `assertThrows(...)` or add a `fail(...)` after the call.

activemq-console/src/test/java/org/apache/activemq/console/QueuesCommandTest.java
```diff
@@ -0,0 +1,277 @@
...
+
+        String result = execute("produce", "Q1", "test-message-body");
+        assertTrue("confirmation with message ID", result.contains("Message sent to Q1"));
+        assertTrue("message ID in output", result.contains("ID:"));
+
+        assertEquals("queue has 1 message after produce", 1, proxy.getQueueSize());
+    }
+
+    @Test(timeout = 30000)
+    public void testProduceMissingBody() throws Exception {
```
jbonofre, line 183: Same silent-pass problem here: the `try/catch` around `execute("produce", "Q1")` has no `fail()`, so the test passes whether or not the exception is thrown. Please use `assertThrows(IllegalArgumentException.class, ...)` or add a `fail(...)`.

activemq-console/src/main/java/org/apache/activemq/console/command/QueuesCommand.java
```diff
@@ -0,0 +1,321 @@
...
+        }
+        ObjectName name = results.get(0).getObjectName();
+        QueueViewMBean q = MBeanServerInvocationHandler.newProxyInstance(
+                createJmxConnection(), name, QueueViewMBean.class, true);
+        q.resume();
+        context.print("Queue resumed: " + queueName);
+    }
+
+    @SuppressWarnings("unchecked")
+    private BrokerViewMBean getBrokerMBean() throws Exception {
```
jbonofre: `getBrokerMBean()` and `dashes()` are duplicated verbatim across all four new command classes (`QueuesCommand`, `TopicsCommand`, `BrokerCommand`, `NetworkConnectorsCommand`). Consider lifting a shared helper (e.g. `getBrokerViewMBean()`) into `AbstractJmxCommand` so all commands reuse it. Non-blocking, but it would avoid the copy-paste drift.

jbonofre (commented):
Thanks for the contribution, this is a nice addition and follows the existing `AbstractJmxCommand` / ServiceLoader pattern cleanly. The tests spinning up real embedded brokers and verifying side effects via JMX proxies are much appreciated. A few comments inline — the main one to address before merge is the negative tests that can pass silently. The rest are cleanups (duplication, dead code) and are non-blocking.

activemq-console/src/main/java/org/apache/activemq/console/command/QueuesCommand.java
```diff
@@ -0,0 +1,321 @@
...
+                createJmxConnection(), brokerName, BrokerViewMBean.class, true);
+    }
+
+    private void requireQueueName(List<String> tokens, String action) throws Exception {
+        if (tokens.isEmpty()) {
+            throw new IllegalArgumentException("Queue name required for '" + action + "'.");
+        }
+    }
+
+    private static String dashes(int count) {
```
jbonofre: `dashes(int)` can just be `"-".repeat(count)` (Java 11+ is available here), which would also let this duplicated helper disappear once the shared base is in place.

activemq-console/src/main/java/org/apache/activemq/console/command/QueuesCommand.java
```diff
@@ -0,0 +1,321 @@
...
+        context.print("Queue created: " + queueName);
+    }
+
+    private void deleteQueue(String queueName) throws Exception {
+        getBrokerMBean().removeQueue(queueName);
+        context.print("Queue deleted: " + queueName);
+    }
+
+    @SuppressWarnings("unchecked")
+    private void purgeQueue(String queueName) throws Exception {
```
jbonofre, line 178: `purge`, `info`, `browse`, `produce`, `pause`, and `resume` all repeat the same ~6-line block: query by `destinationName`, bail with a not-found message if empty, otherwise build a `QueueViewMBean` proxy from the first result. Extracting a `private QueueViewMBean lookupQueue(String name)` helper (returning `null` and printing the not-found message on a miss) would remove most of the repetition and centralize the not-found behavior.

activemq-console/src/test/java/org/apache/activemq/console/BrokerCommandTest.java
```diff
@@ -0,0 +1,156 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.console;
+
+import static org.junit.Assert.assertFalse;
```
jbonofre: `assertFalse` is imported but never used in this test — please remove the unused import.

activemq-console/src/test/java/org/apache/activemq/console/TopicsCommandTest.java
```diff
@@ -0,0 +1,162 @@
...
+
+        LinkedList<String> tokens = new LinkedList<>();
+        for (String arg : args) {
+            tokens.add(arg);
+        }
+        cmd.execute(tokens);
+        return out.toString();
+    }
+
+    private TopicViewMBean getTopicProxy(String topicName) throws Exception {
```
jbonofre: `getTopicProxy(...)` is never called anywhere in this test class (unlike `getQueueProxy` in `QueuesCommandTest`, which is used). Please remove this dead helper.

jbonofre (changes requested):

PR #1967, opened 2026-04-26
fix(console): fix NPE in AbstractJmxCommand.execute and AbstractCommand.handleException

anmol-saxena-14:
`AbstractJmxCommand.execute` called `jmxServiceUrl.toString()` without a null check. `jmxServiceUrl` is never set in local JMX mode (--jmxlocal), so any exception thrown in that mode produced a **NullPointerException** instead of the real error.

`AbstractCommand.handleException` started cause-chain traversal from `exception.getCause()`, which is null for exceptions thrown without a wrapped cause (e.g. IllegalArgumentException). This caused a **NullPointerException** that masked the original exception and swallowed the error message entirely.


anmol-saxena-14:

jbonofre (approved):

PR #1968, opened 2026-04-26
build: make surefire enableProcessChecker overridable for Windows 11 compatibility.

anmol-saxena-14:
**## Problem**
Surefire 3.x uses `wmic` to periodically check whether the parent Maven process is still alive. `wmic` was removed in Windows 11 22H2, so the check always fails — the forked test, JVM concludes the parent has died and calls System.exit(1) before any test runs.

  **Symptoms on Windows 11:**
      [ERROR] The forked VM terminated without properly saying goodbye.
      [ERROR] Process Exit Code: 1

 **## Change**
Extract `enableProcessChecker` into an overridable property with **default** `all` . Windows 11 developers can now run tests using: **`mvn test -Dsurefire.enableProcessChecker=ping`**.

No behaviour change for CI or any other environment.

**Note:**  `ping` uses the stdin pipe to detect parent-process death rather than `wmic`, and works correctly on all platforms.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -41,6 +41,8 @@
          preventing CI from hanging indefinitely on Windows file-lock issues or
          any other test that never terminates. Override per-module if needed. -->
     <surefire.forkedProcessTimeout>300</surefire.forkedProcessTimeout>
+    <!-- Set to 'ping' on Windows 11+ where wmic is unavailable: -Dsurefire.enableProcessChecker=ping -->
+    <surefire.enableProcessChecker>all</surefire.enableProcessChecker>
 
     <siteId>activemq-${project.version}</siteId>
     <projectName>Apache ActiveMQ</projectName>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -967,7 +969,7 @@
             <forkCount>1</forkCount>
             <reuseForks>false</reuseForks>
             <forkedProcessTimeoutInSeconds>${surefire.forkedProcessTimeout}</forkedProcessTimeoutInSeconds>
-            <enableProcessChecker>all</enableProcessChecker>
+            <enableProcessChecker>${surefire.enableProcessChecker}</enableProcessChecker>
```
jbonofre, line 972: It's already the case right?

My understanding is that you are using just a Maven property instead of having the setting `all` inline (in the `test` profile).

Can you elaborate?

anmol-saxena-14: Yes, that is exactly what this change does.

Before, the value was hardcoded inline in the surefire plugin configuration:
`<enableProcessChecker>all</enableProcessChecker>`

After, it is driven by a Maven property with **all** as the default:
**`<!-- in <properties> -->`**
`<surefire.enableProcessChecker>all</surefire.enableProcessChecker>`

**`<!-- in surefire plugin config -->`**
`<enableProcessChecker>${surefire.enableProcessChecker}</enableProcessChecker>`

The default value is unchanged so CI and all existing environments behave identically. The only difference is that a developer on  Windows 11 (where wmic is no longer available) can now override it on the command line without patching the pom:
`mvn test -Dsurefire.enableProcessChecker=ping`

This change follows the same pattern already used in this pom for `surefire.forkedProcessTimeout` and `surefire.argLine`.

anmol-saxena-14:

pom.xml
jbonofre: Understood, and it makes sense. Thanks for the clarification.

jbonofre (approved):

PR #2000, opened 2026-05-11
Disable the message servlet by default

cshannon:
The servlet will still be enabled by default in the web demo project

Edit: I marked the servlet as deprecated and also added comments about why

### DIFF
--- a/assembly/src/release/webapps/api/WEB-INF/web.xml
+++ b/assembly/src/release/webapps/api/WEB-INF/web.xml
@@ -23,7 +23,12 @@
     <display-name>Apache ActiveMQ REST API</display-name>
 
     <!--
-    Uncomment to enable the MessageServlet
+
+    WARNING: The MessageServlet should be used with caution as it is deprecated and unmaintained
+    and there are multiple security related issues. This servlet is primarily meant for demo
+    purposes only and will be removed entirely in a future release. It is recommended to
+    keep it disabled.
+
     <servlet>
         <servlet-name>MessageServlet</servlet-name>
         <servlet-class>org.apache.activemq.web.MessageServlet</servlet-class>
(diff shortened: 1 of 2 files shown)


jbertram:
Does this mean that we're essentially dropping support for the [REST messaging API](https://activemq.apache.org/components/classic/documentation/rest)?

assembly/src/release/webapps/api/WEB-INF/web.xml
```diff
@@ -22,29 +22,6 @@
 
     <display-name>Apache ActiveMQ REST API</display-name>
 
-    <servlet>
```
jbonofre, line 25: Maybe worth to keep in the file, but commented, just in case the user absolutely wants to use it.

jbonofre (dismissed):

mattrpav:
> Does this mean that we're essentially dropping support for the [REST messaging API](https://activemq.apache.org/components/classic/documentation/rest)?

No. This is a security stance of being more “locked down” by default. Administrators are then encouraged to enable features and service they need

assembly/src/release/webapps/api/WEB-INF/web.xml
cshannon: I originally just commented it out but then removed it entirely (it's still the demo project). I can go either way though. @mattrpav what do you think?

cshannon: This was my original commit before force pushing: https://github.com/apache/activemq/commit/ad8ac8770e2976f02dbc5a2eafb7cea377921f79

mattrpav: I'm good with comment it out as a good step here. It'll surface questions from users needing to reanble it and we'll get a sense of how it is used.

Going forward (v7) there is probably a modernization task for this.

mattrpav (approved):
LGTM

assembly/src/release/webapps/api/WEB-INF/web.xml
jbonofre: That sounds like a good plan.

PR #2003, opened 2026-05-11
Block the XBeanBrokerFactory by default inside VMTransportFactory

cshannon:
By default the VMTransportFactory will not be allowed to use XBeanBrokerFactory to create new brokers. Only the default and properties factories will be enabled. To enable the XBeanBrokerFactory, the property `org.apache.activemq.transport.VM_TRANSPORT_FACTORY_SCHEMES_ENABLED` can be configured.


mattrpav (approved):
Code changes and test coverage looks good. Ready to merge, once tests are confirmed to pass.

PR #2004, opened 2026-05-12
Handle validation for Composite URIs without parens

cshannon:
Parentheses are optional when creating composite and nested URIs so this updates the validation to handle missing parens as well.

Follow on to #1847


mattrpav (approved):
LGTM. Good change to use the same uri parser that is used to construct the objects

PR #2006, opened 2026-05-12
[#2005] Fix authorization check on removeDestination

mattrpav:


activemq-broker/src/main/java/org/apache/activemq/security/AuthorizationBroker.java
```diff
@@ -75,12 +75,22 @@ protected SecurityContext checkSecurityContext(ConnectionContext context) throws
...
     }
 
-    protected boolean checkDestinationAdmin(SecurityContext securityContext, ActiveMQDestination destination) {
-        Destination existing = this.getDestinationMap(destination).get(destination);
-        if (existing != null) {
+    protected boolean checkDestinationAdminAdd(SecurityContext securityContext, ActiveMQDestination destination) {
+        if (this.getDestinationMap(destination).get(destination) != null) {
+            return true;
+        }
+
```
cshannon: Minor nit, might want to remove this empty line to be consistent with checkDestinationAdminRemove

cshannon (dismissed):

cshannon:
This looks ok to me as long as tests are good. I looked at the core logic changes and they look correct so that the short circuit now looks good (it short circuits on remove if the dest doesn't exist and short circuits on add if the dest already exists)

activemq-broker/src/main/java/org/apache/activemq/security/AuthorizationBroker.java
mattrpav: Fixed in latest push

cshannon (dismissed):

mattrpav:
- [x] Updated tests to include temp dest to cover both removeDestination and removeDestinationInfo methods.
- [x] Reverted usage of new switch syntax as it does not backport

cshannon (approved):

PR #2019, opened 2026-05-14
#2020 Add expiryCheckEnabled option to MessageEvictionStrategy to skip O(n) expiry scan

mwashburn2:
When a slow-consumer's pending-message buffer exceeds the high-water mark TopicSubscription.add() calls removeExpiredMessages() on every single message add. That method iterates every pending message and calls isExpired(), which is an O(n) scan over the full buffer. When messages carry no TTL (isExpired() always returns false), this scan provides no benefit -- it iterates the entire buffer on every add with zero messages removed. With a large pending limit (e.g. 20,000), this adds up to millions of no-op iterations per second on busy topics. This commit adds an enableExpiry boolean (default true) to PolicyEntry. Setting it to false inserts a single guard in TopicSubscription.add():
    if (enableExpiry && !matched.isEmpty() && matched.size() > max) {
        removeExpiredMessages();
    }

Configuration in activemq.xml:
    <policyEntry topic=">" enableExpiry="false">
        <pendingMessageLimitStrategy>
            <constantPendingMessageLimitStrategy limit="20000"/>
        </pendingMessageLimitStrategy>
    </policyEntry>
Also adds:
- TopicSubscriptionEnableExpiryTest: 11 correctness/propagation/integration tests
- TopicSubscriptionEnableExpiryThroughputTest: throughput comparison test

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/TopicSubscription.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/TopicSubscription.java
@@ -25,7 +25,6 @@ import org.apache.activemq.broker.region.cursors.VMPendingMessageCursor;
 import org.apache.activemq.broker.region.policy.MessageEvictionStrategy;
 import org.apache.activemq.broker.region.policy.OldestMessageEvictionStrategy;
 import org.apache.activemq.command.*;
-import org.apache.activemq.management.MessageFlowStats;
 import org.apache.activemq.thread.Scheduler;
 import org.apache.activemq.transaction.Synchronization;
 import org.apache.activemq.transport.TransmitCallback;
@@ -166,7 +165,7 @@ public class TopicSubscription extends AbstractSubscription {
                         if (maximumPendingMessages > 0 && maximumPendingMessages < max) {
                             max = maximumPendingMessages;
                         }
-                        if (!matched.isEmpty() && matched.size() > max) {
+                        if (messageEvictionStrategy.isExpiryCheckEnabled() && !matched.isEmpty() && matched.size() > max) {
                             removeExpiredMessages();
                         }
                         // lets discard old messages as we are a slow consumer
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/MessageEvictionStrategy.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/MessageEvictionStrategy.java
@@ -42,4 +42,18 @@ public interface MessageEvictionStrategy {
      */
     int getEvictExpiredMessagesHighWatermark();
 
+    /**
+     * Returns whether the eager expired-message scan is enabled.
+     * <p>
+     * When {@code false}, the O(n) scan inside
+     * {@link org.apache.activemq.broker.region.TopicSubscription#add} is skipped entirely.
+     * Set to {@code false} when messages carry no TTL, or when the scan cost outweighs
+     * the benefit of eagerly evicting expired messages from slow-consumer buffers.
+     * <p>
+     * See {@link MessageEvictionStrategySupport} for the default implementation that returns {@code true}.
+     *
+     * @return {@code true} if the expiry scan is enabled (default), {@code false} if skipped
+     */
+    boolean isExpiryCheckEnabled();
+
 }
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/MessageEvictionStrategySupport.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/MessageEvictionStrategySupport.java
@@ -24,6 +24,7 @@ package org.apache.activemq.broker.region.policy;
 public abstract class MessageEvictionStrategySupport implements MessageEvictionStrategy {
 
     private int evictExpiredMessagesHighWatermark = 1000;
+    private boolean expiryCheckEnabled = true;
 
     public int getEvictExpiredMessagesHighWatermark() {
         return evictExpiredMessagesHighWatermark;
@@ -35,6 +36,27 @@ public abstract class MessageEvictionStrategySupport implements MessageEvictionS
     public void setEvictExpiredMessagesHighWatermark(int evictExpiredMessagesHighWaterMark) {
         this.evictExpiredMessagesHighWatermark = evictExpiredMessagesHighWaterMark;
     }
-    
-    
+
+    @Override
+    public boolean isExpiryCheckEnabled() {
+        return expiryCheckEnabled;
+    }
+
+    /**
+     * Controls whether the broker performs an eager expired-message scan when a
+     * non-durable topic subscription's pending slow-consumer buffer exceeds
+     * {@link #getEvictExpiredMessagesHighWatermark()}.
+     * <p>
+     * Set to {@code false} when messages carry no TTL, or when the O(n) scan cost
+     * outweighs the benefit of eagerly evicting expired messages from slow-consumer
+     * buffers. When messages have no TTL, every scan iterates the full buffer without
+     * removing anything, adding latency to every enqueue once the buffer exceeds the
+     * high-water mark.
+     *
+     * @param expiryCheckEnabled {@code false} to skip the scan; {@code true} to enable it (default)
+     */
+    public void setExpiryCheckEnabled(boolean expiryCheckEnabled) {
+        this.expiryCheckEnabled = expiryCheckEnabled;
+    }
+
 }
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/broker/region/TopicSubscriptionEnableExpiryTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/broker/region/TopicSubscriptionEnableExpiryTest.java
@@ -1,0 +1,408 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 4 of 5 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/PolicyEntry.java
```diff
@@ -119,6 +119,8 @@ public class PolicyEntry extends DestinationMapEntry {
 
     private MessageInterceptorStrategy messageInterceptorStrategy = null;
 
+    private boolean enableExpiry = true;
```
mattrpav: Currently, setting the _expireMessagesPeriod_ to zero disables expiry checking at the policyEntry level.

If the desire is to have a way to disable exipry _only_ for the evictionStrategy, then the option should be added to that interface and impls

activemq-broker/src/main/java/org/apache/activemq/broker/region/TopicSubscription.java
```diff
@@ -55,6 +55,7 @@ public class TopicSubscription extends AbstractSubscription {
     private int maximumPendingMessages = -1;
     private MessageEvictionStrategy messageEvictionStrategy = new OldestMessageEvictionStrategy();
     private final AtomicInteger discarded = new AtomicInteger();
+    private boolean enableExpiry = true;
```
mattrpav: This looks like a good addition to TopicSubscription so it may optimize dispatching.

Nit: please rename 'enableExpiry' to 'expiryEnabled' so we get the isExpiryEnabled() linguistics.

mattrpav (changes requested):
I went ahead and created a GH issue for this: https://github.com/apache/activemq/issues/2020.

Please reference [#2020] as prefix to commit message

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/PolicyEntry.java
mwashburn2: Perhaps I should rename the variable expiryEnabledOnEnqueue? I'm just trying to short circuit the O(N) operation on enqueue. Is there a place i'm supposed to update an XSD or the Documentation for this new config?

mwashburn2: i see now the documentation is auto built

mattrpav: The XSD is automatically generated by the build.

I think a config flag 'expiryEnabled' to MessageEvictionStrategy is the best place. The section of code that this applies is within the guard that essentially tells the topic 'go do message eviction strategy stuff'

Also, the subscription automatically get a handle to the MessageEvictionStrategy, so no need to add anything or change signature on the TopicSubscription itself.

```
if (maximumPendingMessages > 0) {
...
```

Updating the doc is always welcome! 

website repo: https://github.com/apache/activemq-website
page ref: https://activemq.apache.org/components/classic/documentation/slow-consumer-handling

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/PolicyEntry.java
```diff
@@ -704,6 +705,7 @@ public void setEnableAudit(boolean enableAudit) {
         this.enableAudit = enableAudit;
     }
 
+
```
mattrpav: please remove blank space

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/PolicyEntry.java
```diff
@@ -119,6 +119,7 @@ public class PolicyEntry extends DestinationMapEntry {
 
     private MessageInterceptorStrategy messageInterceptorStrategy = null;
 
+
```
mattrpav: please remove blank space

activemq-broker/src/main/java/org/apache/activemq/broker/region/TopicSubscription.java
```diff
@@ -530,6 +529,7 @@ public void setMaximumPendingMessages(int maximumPendingMessages) {
         this.maximumPendingMessages = maximumPendingMessages;
     }
 
+
```
mattrpav: please remove blank space

mwashburn2: addressed all 3

mwashburn2:
see also https://github.com/apache/activemq-website/pull/178

cshannon:
I think a simpler approach here that avoids turning off the expiration check might be to just add a short circuit to the expiration loop. Right now the loop runs through all the messages which is the performance problem (as described) but I think we could add a short circuit that just aborts the loop as soon as the first message is seen that isn't expired.

This approach should work in the vast majority of use cases because most of the time messages on the same topic are going to be very similar (ie the same producer is setting the same TTL) so if the current message is not expired it is pretty unlikely any of the other messages are expired so there is no reason to check them. This of course also works well for the use case of never setting a TTL (always aborts on the first message) while still working if there's a bunch of expired messages in a row.

The downside is this would miss expiration on use cases where data is mixed (some messages have TTL and some don't or different TTL times) but that is not that common and there is expiration handling in other places in the broker as well such as on dispatch so I don't think this is a big downside.

@mattrpav - What do you think?

mattrpav:
@cshannon Interesting! I like how your approach sort of 'auto adjusts' for most workflows.

I think ultimately a ExpiryStrategy/ExpiryPolicy interface would be best to encapsulate different desired behaviors across policy entries and destination types.

For workflows that specifically _do not_ to do any expiry checking, an unequivocal flag (like @mwashburn2 has proposed) to disable entirely is probably useful at the very least. 

Also, for hot-spot performance testing with Virtual Threads we'll want to be able to toggle choke points that enter some sort of synchronization or 0(n) processing (think: parameterized unit tests to toggle cache, producerAudit, expiry, etc).

mwashburn2:
In testing with with 3 producers and 50 slow consumers:

Before Change:
15:11:44.062 [main] INFO  c.m.c.j.s.SlowConsumerSendBlockingTest - producer stats (last 1.00s): producer-0={rate=0/s, maxSendLatencyUs=0} producer-1={rate=18/s, maxSendLatencyUs=1760023} producer-2={rate=36/s, maxSendLatencyUs=1772421}  delivered=2250

After Change:
15:04:08.813 [main] INFO  c.m.c.j.s.SlowConsumerSendBlockingTest - producer stats (last 1.00s): producer-0={rate=119/s, maxSendLatencyUs=190856} producer-1={rate=128/s, maxSendLatencyUs=232260} producer-2={rate=36/s, maxSendLatencyUs=573874}  delivered=11550
 
 



This graph shows the number of times in our production environment over the last 14 days a producer was blocked enqueuing. We see it dropped to almost 0 when I released this on May 31st. 
 
image

There was no material change in the number of pending queues (as expected). 

image

cshannon:
> @cshannon Interesting! I like how your approach sort of 'auto adjusts' for most workflows.
> 
> I think ultimately a ExpiryStrategy/ExpiryPolicy interface would be best to encapsulate different desired behaviors across policy entries and destination types.
> 
> For workflows that specifically _do not_ to do any expiry checking, an unequivocal flag (like @mwashburn2 has proposed) to disable entirely is probably useful at the very least.
> 
> Also, for hot-spot performance testing with Virtual Threads we'll want to be able to toggle choke points that enter some sort of synchronization or 0(n) processing (think: parameterized unit tests to toggle cache, producerAudit, expiry, etc).

Adding something to turn if off entirely is fine but I think both things should be done to help out the default case if you want to do that. Even with an expiration policy to turn it off, most people are not going to know to do that so I think the strategy I proposed makes sense for the default case and could be added easily. With a fast abort like I proposed, there's likely not going to be much need for most people to turn it off as they won't notice it being slow. Also there's obviously the situation where people do expect expired messages but rarely, and would prefer to not have to pay the performance penalty if the expiration isn't common.

cshannon:
Also to be clear, the change I'm suggesting can be done (and really should be done) in another issue as it's not related to disabling it entirely, you don't need to add that now.

activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/MessageEvictionStrategy.java
```diff
@@ -42,4 +42,18 @@ public interface MessageEvictionStrategy {
...
+     * When {@code false}, the O(n) scan inside
+     * {@link org.apache.activemq.broker.region.TopicSubscription#add} is skipped entirely.
+     * Set to {@code false} when messages carry no TTL, or when the scan cost outweighs
+     * the benefit of eagerly evicting expired messages from slow-consumer buffers.
+     * <p>
+     * See {@link MessageEvictionStrategySupport} for the default implementation that returns {@code true}.
+     *
+     * @return {@code true} if the expiry scan is enabled (default), {@code false} if skipped
+     */
+    boolean isExpiryScanEnabled();
```
mattrpav: Last minor nit to align the terminology with other areas of the code base:

Change to: 
```
isExpiryCheckEnabled()
```

mattrpav:
Checklist for merge:

 - [x] Tidy up the minor field name nit I noted above
 - [x] Rebase commits down to 1
 - [x] Update the commit comment to include the issue id for auto-linking

Example:
``` 
[#2020] Add enableExpiry option to PolicyEntry to skip O(n) expiry scan
```

@mwashburn2 once you get a chance to run down these changes, I'll merge and back-port to 5.19.x. Thanks!

mattrpav (dismissed):
LGTM

mwashburn2:
updated the docs as well 

https://github.com/apache/activemq-website/pull/178

cshannon (changes requested):
The policy changes seem fine, the only issue I see is these are new tests and they are using JUnit 3 style and extending TestCase and that should that should be switched to JUnit 4 style tests.

mwashburn2:
> The policy changes seem fine, the only issue I see is these are new tests and they are using JUnit 3 style and extending TestCase and that should that should be switched to JUnit 4 style tests.

this has been addressed. Thanks

mattrpav (approved):
LGTM

mattrpav (approved):
LGTM

cshannon (approved):

PR #2025, opened 2026-05-19
Harden web console and Jolokia access by default

jbonofre:
jetty.xml:
- Add Referrer-Policy and Permissions-Policy response headers; provide commented-out Strict-Transport-Security for HTTPS deployments.
- Wrap the existing HandlerCollection in an InetAccessHandler and restrict access to loopback (127.0.0.1, ::1) by default. Provide commented-out templates for RFC1918 includes and exclude rules.
- Tie Server.start() to the loopback include beans via depends-on so the allow-list is fully populated before the server starts.
- Add a commented-out ForwardedRequestCustomizer for reverse-proxy setups, with a warning about header spoofing when the proxy does not strip inbound X-Forwarded-* headers.

jolokia-access.xml:
- Restrict Jolokia to HTTP POST to kill GET-based CSRF/SSRF vectors.
- Deny destructive and privileged broker operations (terminateJVM, stop, restart, add/removeConnector, add/removeQueue/Topic, durable subscriber lifecycle, runtime limit setters, reloadLog4jProperties).
- Deny destination message-data mutation (purge, remove/copy/move messages, sendTextMessage*, pause/resume); browse* remains allowed.
- Deny durable subscriber destroy/setSelector and JobScheduler removeJob*.
- Deny NetworkConnector Password/RemotePassword attributes and credential setters.
- Deny known JMX RCE / introspection surfaces: javax.management.loading MLet, JMImplementation, java.util.logging, java.lang:type=Memory and ClassLoading, plus java.lang:type=Runtime SystemProperties and InputArguments.

### DIFF
--- a/assembly/src/release/conf/jetty.xml
+++ b/assembly/src/release/conf/jetty.xml
@@ -21,6 +21,31 @@
 
     <bean id="httpConfig" class="org.eclipse.jetty.server.HttpConfiguration">
         <property name="sendServerVersion" value="false"/>
+        <!--
+            When the web console sits behind a reverse proxy or load balancer,
+            uncomment the customizer below so Jetty honors X-Forwarded-* (or
+            RFC 7239 Forwarded) headers when populating request.getRemoteAddr(),
+            request.getScheme(), etc. This is what gets the real client IP into
+            access logs and audit trails.
+
+            WARNING: ForwardedRequestCustomizer trusts these headers from any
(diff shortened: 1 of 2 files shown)


cshannon (approved):
@jbonofre - nice job! I think this looks very good, we can keep hardening if needed but this is looking good to me for now

mattrpav (approved):
LGTM

PR #2026, opened 2026-05-19
Remove "java.lang" package as a default allowed serializable package

cshannon:
Many classes in the java.lang package should not ever need to be serialized so this commit removes the default package and instead includes an allow list of classes that are ok to serialize that are part of the package. Users have the option to restore the previous behavior by appending "java.lang" back to the serialized packages property if desired.

The main goal of this is to restrict the allowed list to help prevent any future issues that could be uncovered by allowing unnecessary types to be serialized.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/util/ClassLoadingAwareObjectInputStream.java
+++ b/activemq-client/src/main/java/org/apache/activemq/util/ClassLoadingAwareObjectInputStream.java
@@ -32,15 +32,21 @@ public class ClassLoadingAwareObjectInputStream extends ObjectInputStream {
     private static final ClassLoader FALLBACK_CLASS_LOADER =
         ClassLoadingAwareObjectInputStream.class.getClassLoader();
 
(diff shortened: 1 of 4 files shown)


mattrpav (approved):
LGTM

activemq-client/src/main/java/org/apache/activemq/util/ClassLoadingAwareObjectInputStream.java
```diff
@@ -98,7 +104,7 @@ private boolean trustAllPackages() {
     }
 
     private void checkSecurity(Class clazz) throws ClassNotFoundException {
-        if (trustAllPackages() || clazz.isPrimitive()) {
+        if (trustAllPackages() || clazz.isPrimitive() || ALLOWED_JDK_TYPES.contains(clazz)) {
```
mattrpav, line 107: This is a good change to permanently support these core types

jbonofre (approved):

PR #2029, opened 2026-05-20
Harden default broker and web console configuration

jbonofre:
Reduce the attack surface of the sample broker shipped in the binary assembly:

* activemq.xml: enable only the openwire TCP transportConnector by default; amqp, stomp, mqtt and ws are commented out with a note pointing to the SSL-secured variants.
* activemq.xml: add a commented <plugins> block wiring JAAS authentication, destination-level authorization and broker-side audit logging, with a prominent reminder to rotate the default admin=admin credentials and an ACTIVEMQ_OPTS hint for restricting ObjectMessage deserialization.
* jetty.xml: suppress X-Powered-By and Date response headers in addition to the already-disabled Server header.
* jetty.xml: add Referrer-Policy, Permissions-Policy, Cross-Origin-Opener-Policy and Cross-Origin-Resource-Policy response headers alongside the existing CSP/XFO/XSS rules.

### DIFF
--- a/assembly/src/release/conf/activemq.xml
+++ b/assembly/src/release/conf/activemq.xml
@@ -55,6 +55,50 @@
 
 
         <!--
+            Security hardening (disabled by default so the sample broker starts
+            out-of-the-box). Uncomment the <plugins> block below to enforce:
+              * JAAS authentication using the 'activemq' realm in conf/login.config
+              * Destination-level authorization based on conf/groups.properties
+              * Broker-side audit logging of connections, producers and consumers
+
+            BEFORE ENABLING IN PRODUCTION:
+              1. ROTATE the credentials in conf/users.properties. The default
+                 'admin=admin' user is well-known and MUST NOT be used as-is.
+              2. Tighten the authorizationMap below to match the destinations
+                 you actually use; the defaults grant access only to 'admins'.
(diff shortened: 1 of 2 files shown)


assembly/src/release/conf/activemq.xml
```diff
@@ -105,10 +149,15 @@
         <transportConnectors>
             <!-- DOS protection, limit concurrent connections to 1000 and frame size to 100MB -->
             <transportConnector name="openwire" uri="tcp://0.0.0.0:61616?maximumConnections=1000&amp;wireFormat.maxFrameSize=104857600"/>
```
cshannon, line 158: This is plaintext by default now, we may want to add a comment at least mentioning it's recommended to use SSL here and only use TCP for initial testing.

cshannon (dismissed):

assembly/src/release/conf/activemq.xml
jbonofre: That's a good point. Let me add a comment there.

jbonofre: I added a comment.

cshannon (dismissed):

mattrpav (dismissed):
LGTM

cshannon (approved):

PR #2034, opened 2026-05-21
[#] SECURITY.md updates

mattrpav:


SECURITY.md
```diff
@@ -70,3 +72,13 @@ Layers of security provide valuable options to prevent attacks, and to provide a
 6. Add allow/deny lists to transport connectors to limit IP addresses
 
 7. [Done] VM Transport creation blocks the XBean factory by default
+
+## Security vs Features
+
+AI code scanning tools often mistaken designed features as a security issue.
```
cshannon:
```suggestion
AI code scanning tools often mistaken designed features as a security issue. It is the responsibility of the reporter to review AI output and verify if it's a real issue. There has been a large number of invalid submissions that could be avoided by simply reviewing the JMS spec and the features of the broker itself.

Some of the most common reported examples:
```

jbonofre (dismissed):

SECURITY.md
```diff
@@ -70,3 +72,13 @@ Layers of security provide valuable options to prevent attacks, and to provide a
...
+
+1. JMS Selectors - An optional query parameter designed to filter messages on a queue or topic that is not security related
```
cshannon:
```suggestion
1. JMS Selectors - An optional query parameter designed to filter messages on a queue or topic that is not security related. It is used by clients to consume a subset of messages on the destination instead of all messages. However, if a client is authorized for a destination it is always free to consume all the messages if it chooses so by simply not setting the selector. Therefore any reports showing issues with selectors allowing the consumption of extra messages would be considered a bug and not a security issue as long as it doesn't escape the destination the client is authorized for.
```

SECURITY.md
```diff
@@ -70,3 +72,13 @@ Layers of security provide valuable options to prevent attacks, and to provide a
...
+
+2. ClientId - A non-secret unique identifier used to provide once-and-only-once delivery that are designed to be used between connections and be deleted
```
cshannon:
```suggestion
2. ClientId - A non-secret unique identifier used to provide once-and-only-once delivery that are designed to be used between connections and be deleted. The JMS spec specifically allows any authorized connection to use the same clientid as long as it isn't currently in use. Some protocols, such as MQTT, also allow link stealing and taking over if in use.

3. Durable Subscriptions - The JMS spec allows authorized connections to connect to any existing durable subscription (combination of client id and subscription name) as long as it is offline. Authorized clients are allowed to delete the durable subscriptions as well even if they didn't create it.
```

SECURITY.md
```diff
@@ -70,3 +72,13 @@ Layers of security provide valuable options to prevent attacks, and to provide a
...
+
+3. BlobMessages - Blob message support is a side-channel for moving large messages with the JMS API by routing the large message through a different endpoint such as http, sftp or scp. Clients using BlobMessages are responsible for validating the authenticity and validity of the uri provided by the broker. ActiveMQ recommends using SSL secured transports, with two-way SSL as the most preferred.
```
cshannon:
```suggestion
4. BlobMessages - Blob message support is a side-channel for moving large messages with the JMS API by routing the large message through a different endpoint such as http, sftp or scp. Clients using BlobMessages are responsible for validating the authenticity and validity of the uri provided by the received message before taking any action such as downloading or deleting the file. ActiveMQ recommends using SSL secured transports, with two-way SSL as the most preferred.
```

cshannon:
@mattrpav - I think one more comment to add to the new section is to mention that we don't consider vulnerability reports that require no authentication or authorization to be valid. We constantly get reports with "out of the box" defaults meant for developers. While we are changing that, it's still good to point out that if a security issue requires authentication to be turned off then that is not valid because there should be no expectation of protection if a user doesn't turn on proper authentication and authorization controls.

cshannon:
I'm going to go ahead and merge what we have for now and we can make more updates going forward as i'm sure there will be plenty more changes as we go.

cshannon (approved):

PR #2044, opened 2026-05-26
Bump default JVM heap and add OOM diagnostics

jbonofre:
- Raise the shipped default heap from `-Xms64M -Xmx1G` to `-Xms512M -Xmx2G` in `assembly/src/release/bin/setenv`. The previous default is undersized for a production broker; the new value is still modest and remains overridable via `ACTIVEMQ_OPTS_MEMORY`.
- Add a new `ACTIVEMQ_OOM_OPTS` variable defaulting to `-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=$ACTIVEMQ_DATA -XX:+ExitOnOutOfMemoryError` so that an `OutOfMemoryError` produces a heap dump under the broker data directory and the JVM terminates rather than continuing in a degraded state.
- Mirror the same `ACTIVEMQ_OOM_OPTS` fallback inside `assembly/src/release/bin/activemq` so the diagnostics still apply if `setenv` is absent.

Both new variables remain overridable from the environment or from any of the higher-priority configuration files (`/etc/default/activemq`, `~/.activemqrc`, `bin/setenv`).

### DIFF
--- a/assembly/src/release/bin/setenv
+++ b/assembly/src/release/bin/setenv
@@ -31,7 +31,7 @@
 
 # Set jvm memory configuration (minimal/maximum amount of memory)
 if [ -z "$ACTIVEMQ_OPTS_MEMORY" ] ; then
-    ACTIVEMQ_OPTS_MEMORY="-Xms512M -Xmx2G"
+    ACTIVEMQ_OPTS_MEMORY="-Xms64M -Xmx2G"
 fi
 
 # Terminate the JVM on OutOfMemoryError so the broker is not left running in a degraded state.
--- a/assembly/src/release/bin/setenv
+++ b/assembly/src/release/bin/setenv
@@ -31,10 +31,21 @@
 
 # Set jvm memory configuration (minimal/maximum amount of memory)
 if [ -z "$ACTIVEMQ_OPTS_MEMORY" ] ; then
-    ACTIVEMQ_OPTS_MEMORY="-Xms64M -Xmx1G"
+    ACTIVEMQ_OPTS_MEMORY="-Xms512M -Xmx2G"
 fi
+
+# Terminate the JVM on OutOfMemoryError so the broker is not left running in a degraded state.
+# Heap dump on OOM is disabled by default to avoid unexpected disk usage (especially with large
+# heap sizes). Uncomment ACTIVEMQ_OOM_DUMP_OPTS to enable heap dumps; -XX:HeapDumpGzipLevel=6
+# requires JDK 17+.
+if [ -z "$ACTIVEMQ_OOM_OPTS" ] ; then
+    ACTIVEMQ_OOM_OPTS="-XX:+ExitOnOutOfMemoryError"
+    #ACTIVEMQ_OOM_DUMP_OPTS="-XX:+HeapDumpOnOutOfMemoryError -XX:HeapDumpPath=$ACTIVEMQ_DATA -XX:HeapDumpGzipLevel=6"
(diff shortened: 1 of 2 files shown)


jbonofre:
@cshannon I propose to update the default configuration like this.

cshannon:
Hmm, I think this might need more discussion. I think the 2GB default is good but I don't know about the OOM heap dump as that will cause unexpected output to a file system for users. I think that needs to be configurable on set up maybe. We could bring it up for discussion on the dev list.

jbonofre:
@cshannon yes, I can add a optional flag to setup OOM. I'm just sharing my experience: I have to ask more of the ActiveMQ users to enable it it to investigate 😄 So having a way to enable it by default would be good (or at least documented).

cshannon:
> @cshannon yes, I can add a optional flag to setup OOM. I'm just sharing my experience: I have to ask more of the ActiveMQ users to enable it it to investigate 😄 So having a way to enable it by default would be good (or at least documented).

That sounds good, I personally set that flag myself so I am fine with it but I figure someone might not want the extra output.
It can be an issue if the broker gets into an OOM restart loop and has the potential to fill up a partition so users should be aware it's on or at least opt into it.

mattrpav:
+1 to have a preconfigured but commented out set of options. Having it on by default can quickly lead to disk space usage issues -- there is also a 'where to write the files' problem

jbonofre:
@cshannon @mattrpav what are your thoughts on this PR? I think this default configuration is a good improvement.

cshannon:
@jbonofre - I think the memory updates make sense and I agree with those. For the OOM settings to dump to disk I think we talked about making that optional with a flag to turn on. I don't think it should be on by default because of the unexpected disk usage. Some people might configure large heap sizes.

Also, I would recommend adding compression: `-XX:HeapDumpGzipLevel=6` as that was added in JDK 17 (as long as we are not backporting this)

jbonofre:
@cshannon thanks! It makes sense. Let me update the PR accordingly.

jbonofre:
@cshannon Thanks for the feedback, I've updated the PR:

- Heap dump on OOM is now **disabled by default**. Users can opt in by uncommenting the `ACTIVEMQ_OOM_DUMP_OPTS` lines in `bin/setenv` (or `bin/activemq` for the no-setenv fallback). `-XX:+ExitOnOutOfMemoryError` remains on by default since it has no disk impact.
- Added `-XX:HeapDumpGzipLevel=6` to the opt-in template (requires JDK 17+).

assembly/src/release/bin/setenv
```diff
@@ -31,10 +31,21 @@
 
 # Set jvm memory configuration (minimal/maximum amount of memory)
 if [ -z "$ACTIVEMQ_OPTS_MEMORY" ] ; then
-    ACTIVEMQ_OPTS_MEMORY="-Xms64M -Xmx1G"
+    ACTIVEMQ_OPTS_MEMORY="-Xms512M -Xmx2G"
```
mattrpav: I like the idea of setting the initial small, like 64MB. If the default out-of-the-box says '512MB' this will be inaccurately construed as 'ActiveMQ needs a lot of memory'.

jbonofre: That works for me.

jbonofre: Let me do the change.

jbonofre: Done — reverted to `-Xms64M` while keeping `-Xmx2G`. The JVM will start small and grow on demand, avoiding the impression that ActiveMQ has a 512MB baseline requirement.

cshannon (approved):

mattrpav (approved):
LGTM

PR #2065, opened 2026-06-01
[6.2.x] Update Stomp transports with improved validation (#2064)

cshannon:
This update makes the following changes to improve validation for the Stomp transport:

* Verifies that the first frame seen by the server is either a CONNECT (or FRAME) frame.
* Verifies that a duplicate CONNECT (or FRAME) frame is not received.
* Adds validation to make sure a content-length header that is set is not negative.
* Adds a new server mode (default true) to the Stomp wireformat to handle the validation differences between clients and servers. Client mode is only used for testing (currently). Also adds the option to configure using the StompWireFormatFactory in case there is a future use case.
* Centralizes the state tracking for frame size validation and for the new validation checks inside StompWireFormat so that it is shared by NIO, non-NIO and WS transports.
* Adds tests to verify everything for the NIO transports, non-NIO transprots and WS transports.

If any of these new validation checks throw a protocol error then it is marked as a fatal exception, an error is sent to the client and connection closed. Both NIO and non-NIO will stop parsing the rest of the frame on error, but only NIO transport errors will stop reading the frame from the socket buffer because non NIO requires reading the entire frame into a buffer first to validate.

(cherry picked from commit 1493db95b5918d4f4a305fd1df8155f57c38850b)
Backport of #2064

### DIFF
--- a/activemq-http/src/test/java/org/apache/activemq/transport/ws/StompWSTransportTest.java
+++ b/activemq-http/src/test/java/org/apache/activemq/transport/ws/StompWSTransportTest.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 import org.apache.activemq.transport.stomp.Stomp;
 import org.apache.activemq.transport.stomp.StompFrame;
 import org.apache.activemq.util.Wait;
+import org.apache.activemq.util.Wait.Condition;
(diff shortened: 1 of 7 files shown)


cshannon:
This backport also includes #2067

PR #2077, opened 2026-06-05
Add validation for LDAP network connector URIs

cshannon:
Validate the URIs provided by LDAP searches do not contain any protocols from the denied list.


mattrpav (approved):
LGTM. Good improvement to move the validator to a util class

PR #2080, opened 2026-06-05
Add validation for WireFormatInfo

cshannon:
The max number of properties in the map for the WireFormatInfo command was lowered to 64 and a new max value buffer size was added to validate buffers created during unmarshalling of the properties map. The max buffer allowed has been set to 512 bytes.

This also adds a maxDepth check for handling nested collections inside the map. It is set to 0 which will block list/map from being used as a value for a property.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
+++ b/activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
@@ -67,15 +67,11 @@ public final class MarshallingSupport {
     }
 
     public static Map<String, Object> unmarshalPrimitiveMap(DataInputStream in) throws IOException {
-        return unmarshalPrimitiveMap(in, Integer.MAX_VALUE);
+        return unmarshalPrimitiveMap(in, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE);
     }
 
-    public static Map<String, Object> unmarshalPrimitiveMap(DataInputStream in, boolean force) throws IOException {
-        return unmarshalPrimitiveMap(in, Integer.MAX_VALUE, force);
-    }
-
-    public static Map<String, Object> unmarshalPrimitiveMap(DataInputStream in, int maxPropertySize) throws IOException {
-        return unmarshalPrimitiveMap(in, maxPropertySize, false);
+    public static Map<String, Object> unmarshalPrimitiveMap(DataInputStream in, int maxPropertySize, int maxBufferSize, int maxDepth) throws IOException {
+        return unmarshalPrimitiveMap(in, false, maxPropertySize, maxBufferSize, maxDepth, 0);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
```diff
@@ -350,7 +358,11 @@ public static int writeUTFBytesToBuffer(String str, long count,
     }
 
     public static String readUTF8(DataInput dataIn) throws IOException {
-        int utflen = dataIn.readInt();
+        return readUTF8(dataIn, Integer.MAX_VALUE);
```
tabish121, line 361: This one is a bit dangerous so should check on who is calling it.  I might actually pass a value of `dataIn.available() ` unless there is a case where this can be used on blocking reads like a read from a socket input stream.

activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
```diff
@@ -158,10 +164,11 @@ public static void marshalPrimitive(DataOutputStream out, Object value) throws I
     }
 
     public static Object unmarshalPrimitive(DataInputStream in) throws IOException {
-        return unmarshalPrimitive(in, false);
+        return unmarshalPrimitive(in, false, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE, 0);
```
tabish121, line 167: Another case where I'd consider if the passed size limit could be the `stream.available() `

activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
```diff
@@ -67,15 +67,11 @@ public static void marshalPrimitiveMap(Map<String, Object> map, DataOutputStream
     }
 
     public static Map<String, Object> unmarshalPrimitiveMap(DataInputStream in) throws IOException {
-        return unmarshalPrimitiveMap(in, Integer.MAX_VALUE);
+        return unmarshalPrimitiveMap(in, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE);
```
tabish121, line 70: If possible consider capping the max size from these methods to `in.available()` to that if the encoded size is bogus you will fail early vs large allocation attempt. Unsure if this is possible if these are read via a socket input stream which is expected to block on reads waiting more bytes

activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
cshannon: That is a good point, I agree this could be looked into for improvement as well. This is a utility class and method so I would need to look where else it could be used. 

This particular method wasn't touched and is used at all by WireFormatInfo so I don't think it needs to be changed now. It could be used in the future by various use cases, also there's a chance (and likely) the stream could be a compressed stream which would impact the available() call as well. So I can look into it as a separate follow on.

activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
cshannon: [Applies here too](https://github.com/apache/activemq/pull/2080#discussion_r3365192557)

activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
cshannon: Same as other 2 comments to close the loop, I'll investigate as a follow on and look at other places using the util.

tabish121 (approved):

PR #2083, opened 2026-06-06
Ensure at most one BrokerInfo command is received

cshannon:
There should only ever be at most one BrokerInfo command received on a connection so we should throw an exception and close the connection if a second is receveived as that is a protocol error.


tabish121 (approved):
Good catch, I was worried something like this might exist

mattrpav (approved):
LGTM

PR #2088, opened 2026-06-08
Limit WireFormatInfo platform details to a reasonable length

cshannon:
Now that #2080 limits value sizes for properties inside of WireFormatInfo we should add a check to make sure platform details do not go past that limit. Limit has been set to 512 bytes as that should be more than enough as the details are usually less than half that. This is not turned of by default but if someone is using it this will prevent potential future problems for too large a value.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionMetaData.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionMetaData.java
@@ -22,6 +22,7 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import jakarta.jms.ConnectionMetaData;
+import org.apache.activemq.command.WireFormatInfo;
 
 /**
  * A <CODE>ConnectionMetaData</CODE> object provides information describing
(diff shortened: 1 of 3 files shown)


activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionMetaData.java
```diff
@@ -175,8 +179,10 @@ private static String getPlatformDetails() {
             platformInfo.append(", ");
             platformInfo.append(System.getProperty("os.arch"));
 
-            details = platformInfo.toString();
-        } catch (Throwable e) {
+            // truncate to the max allowed length if too long
+            details = platformInfo.length() > PLATFORM_DETAILS_MAX_LENGTH ?
```
tabish121, line 185: This cuts off the details without any indication they were cut off, which is probably fine. I have seen some implementation do this by inserting an ellipsis after a given point and then appending a small bit trailing value such that you get a sort of visual indication that it was clipped.  Probably not needed here so just making sure you thought about that.

tabish121 (dismissed):

mattrpav:
Some practical limits from a quick search show 256 is right up at the practical usage. 384 or 512 may be slightly better to avoid concatenation for known real-world data values.

Note: Cloud provider kernel version numbers can get long.

| System Property | Typical Max Length (Characters) | Real-World Long Examaple |
|------|-----|-----|
| java.vm.version | 45–60 | 25.0.0-internal-developer_2026_03_2_10_15-b01 |
| java.vendor | 25–40 | International Business Machines Corp.| 
| java.version | 10–20 | 1.8.0_412-20240417-b01 |
| os.arch | 5–10 | aarch64 or x86_64 |
| os.name | 15–25 | Red Hat Enterprise Linux | 
| os.version | 15–30 | 5.4.0-105-generic or 6.1.0-21-amd64 |

cshannon:
@mattrpav - thanks for pointing that out, i can just bump to 512 which is still not that big. 512 is our default max so I can just make it the same I suppose

activemq-client/src/main/java/org/apache/activemq/command/WireFormatInfo.java
```diff
@@ -42,7 +42,7 @@ public class WireFormatInfo implements Command, MarshallAware {
     // Max number of properties allowed in the map is 64
     static final int MAX_PROPERTY_SIZE = 64;
     // Used to validate property values that allocate buffers, limit to 512 bytes
-    static final int MAX_PROPERTY_BUFFER_SIZE = 512;
+    public static final int MAX_PROPERTY_BUFFER_SIZE = 512;
```
tabish121, line 45: Good catch I forgot that you made the default for WireFormatInfo 512 from the original 256

tabish121 (approved):

mattrpav (approved):
LGTM

activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionMetaData.java
cshannon: Yeah I think it's fine here, it's off by default and its now 512 and shouldn't really be hit but we could improve it later if needed. This is more of just a fail safe. I doubt anyone even turns this setting on to be honest.

PR #2097, opened 2026-06-08
Improve destination type configuration mismatch visibility

StanDarJeep:
Adds a server-side log hint when an authorization failure may be caused by a queue/topic type mismatch in the broker configuration. When a user is denied access to a destination but would have access to the same destination name under the alternate type (e.g., queue://foo would be allowed but client accesses topic://foo), the broker now logs a warning suggesting the operator verify the destination type specified in the configuration. This is a misconfiguration that can be difficult to diagnose from only the "not authorized" error message. The hint is only written to the broker log; the client-facing error message is unchanged and no extra details are exposed.

Example log output when the authorization failure occurs (the first line is the new log):

```
2026-06-03 16:33:45,507 | WARN  | Possible destination type mismatch: user 'test-user' is not authorized to write topic 'test-queue', but a queue authorization entry for 'test-queue' would grant access. Verify the destination type in the broker configuration. | org.apache.activemq.security.AuthorizationBroker | ActiveMQ Transport: tcp:///127.0.0.1:60031@61613
2026-06-03 16:33:45,508 | WARN  | Async error occurred: User test-user is not authorized to write to: topic://test-queue | org.apache.activemq.broker.TransportConnection.Service | ActiveMQ Transport: tcp:///127.0.0.1:60031@61613
2026-06-03 16:33:45,508 | WARN  | Exception occurred while processing a command: java.lang.SecurityException: User test-user is not authorized to write to: topic://test-queue | org.apache.activemq.transport.stomp.ProtocolConverter | ActiveMQ BrokerService[localhost] Task-1
```

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/security/AuthorizationBroker.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/security/AuthorizationBroker.java
@@ -35,6 +35,8 @@ import org.apache.activemq.command.ConsumerInfo;
 import org.apache.activemq.command.DestinationInfo;
 import org.apache.activemq.command.Message;
 import org.apache.activemq.command.ProducerInfo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Verifies if a authenticated user can do an operation against the broker using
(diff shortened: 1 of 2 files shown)


mattrpav:
I'm going to deny and close this request for the following reasons:

1. It is a security information leak to share what resources a user may have access to (or if a provided user has any access)
2. It is semantically impossible that this would be useful, b/c may permissions are granted as wildcard patterns.

edit: If there is a use case or an issue around discovering access, please open a ticket first and we can discuss possible approaches.

PR #2107, opened 2026-06-11
Remove extra buffer copy in StompNIOSSLTransport

cshannon:
Optimizes the StompNIOSSLTransport by removing the unnecessary extra byte array allocation and buffer copy when processing frames


tabish121 (approved):
Makes sense to go straight to the ByteBuffer, so long as the tests pass

mattrpav (approved):
LGTM. Good cleanup

cshannon:
As an aside, MQTT also has the same issue but is not as trivial to fix and has more cleanup to replace it including tests. I figure not as many people use MQTT but it's something that could be done in another PR

I'll let the tests pass and if no issue i'll merge.

PR #2113, opened 2026-06-14
Replace active temp dests map with set

cshannon:
For some reason ActiveMQConnection was using a map instead of a set to store the destinations, and just stored the exact same thing as the key and value. Furthermore, when checking if the map contained the destination a call was being made to containsValue() which requires iterating over the entire map.

This commit replaced the Map with a Set which simplifies things and makes the contains() check constant. Also, the scope of the set was changed from public to package because it makes no sense to have the scope as public and should be limited to only classes in the same package.


cshannon:
The flaky test that keeps failing will be fixed in #2112

mattrpav (approved):
LGTM

PR #2118, opened 2026-06-15
Provide a flexible filter style input stream that limits read amounts

tabish121:
This commit supplies a utility type that acts as a filter input stream style stream wrapper than can be configured with an available bytes window which decreases as bytes are read from the stream. This allows a transport or other utility to limit what can be read based on a fixed amount such as a max frame size option.


cshannon (approved):
LGTM, nice job with the unit tests!

mattrpav (approved):
LGTM. Nice add!

PR #2122, opened 2026-06-15
Add flag to optionally enable temp destination stealing

cshannon:
This adds a flag to allow other connections to create subscriptions on a temp destination, the default has been set to off. Setting to true enables some use cases to work such as failover but most of the time it is not needed and should be disabled to be JMS client spec compatible.

The java client also already checked this, this now adds a broker side check as well.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
@@ -200,6 +200,8 @@ public class BrokerService implements Service {
     private boolean useVirtualTopics = true;
     private boolean useMirroredQueues = false;
     private boolean useTempMirroredQueues = true;
+    private boolean allowTempDestinationStealing = false;
+
     /**
      * Whether or not virtual destination subscriptions should cause network demand
      */
@@ -1936,6 +1938,14 @@ public class BrokerService implements Service {
 
     public void setUseTempMirroredQueues(boolean useTempMirroredQueues) {
         this.useTempMirroredQueues = useTempMirroredQueues;
+    }
+
+    public boolean isAllowTempDestinationStealing() {
+        return allowTempDestinationStealing;
+    }
+
+    public void setAllowTempDestinationStealing(boolean allowTempDestinationStealing) {
+        this.allowTempDestinationStealing = allowTempDestinationStealing;
     }
 
     public synchronized JobSchedulerStore getJobSchedulerStore() {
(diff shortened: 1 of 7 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/DestinationFactoryImpl.java
```diff
@@ -88,8 +91,11 @@ public Destination createDestination(ConnectionContext context, ActiveMQDestinat
                 return queue;
             }
         } else if (destination.isTemporary()) {
-            
-            Topic topic = new Topic(brokerService, destination, null, destinationStatistics, taskRunnerFactory);
+            final ActiveMQTempDestination tempDest = (ActiveMQTempDestination)destination;
+            if (tempDest.getConnectionId() == null) {
+                throw new IllegalArgumentException("Temporary topic must have a connectionId");
+            }
+            Topic topic = new TempTopic(brokerService, destination, null, destinationStatistics, taskRunnerFactory);
```
mattrpav, line 98: Good fix! A proper TempTopic v Topic here.

activemq-broker/src/main/java/org/apache/activemq/broker/BrokerService.java
```diff
@@ -200,6 +200,8 @@ public class BrokerService implements Service {
     private boolean useVirtualTopics = true;
     private boolean useMirroredQueues = false;
     private boolean useTempMirroredQueues = true;
+    private boolean allowTempDestinationStealing = false;
```
mattrpav: I think this would be better in the TempDestinationPolicy vs broker level. Overall, moving configurations out of broker-wide.

cshannon: This doesn't make sense, policies get set on a destination hierarchy and temporary destinations by definition don't fit into any policy. There is no hierarchy with temporary destinations as the names are just client generated, this needs to be at the broker level.

mattrpav: There is a PolicyEntry for setting options for temp queue and/or temp topic that covers all temp queues or topic destinations in one shot. 

All other destination behaviors are there like memory, eviction policy, etc. feels like it’s a better fit overall

cshannon: Ah yes, I completely forgot there is even a flag for policy entries for temp. I can move it, it likely doesn't matter much either way as most people are unlikely to change the default given the java client already blocks it.

cshannon:
@mattrpav - setting was moved into the policy entry and is now applied for temp destinations

mattrpav (approved):
LGTM

mattrpav:
Side note: Probably should update the test for dynamic config update for PolicyEntry anytime we make a PolicyEntry change, but I don't think this one is a big deal and we can update PolicyEntry test in a sweep of updates at a later time.

PR #2123, opened 2026-06-16
Preserve ${...} placeholders when removing/modifying networkConnectors in RuntimeConfigurationPlugin

jbonofre:
Fix `NetworkConnectorProcessor.configMatch` so it resolves `${...}` placeholders on the DTO side before comparing against the live `NetworkConnector` (whose attributes are already resolved by Spring). Without this, a networkConnector declared with placeholders could never be matched for removal or modification by the RuntimeConfigurationPlugin.

Adds a test covering `uri`, `userName`, and `password` placeholders across removal.


cshannon:
At first glance this looks ok to me, i will do a couple quick tests though. Thoughts @mattrpav ? I don't use the runtime configuration plugin a ton

cshannon (approved):

Issue #2132, opened 2026-06-22
Modernize SSL support to include per-connector SSLContext, pkcs12 default, keyalias support.


mattrpav:
- [x] Migrate from ThreadLocal by default
- [x] Default to prefer pkcs12 keystore type
- [x] Support automatic JKS keystore type fallback
- [x] Support keyAlias
- [x] Support SSLContext per-transportConnector, per-networkConnector with broker-wide fallback
- [x] Prepare for reload support

ref: [AMQ-9753](https://issues.apache.org/jira/browse/AMQ-9753)
ref: [AMQ-9694](https://issues.apache.org/jira/browse/AMQ-9694)

PR #2133, opened 2026-06-22
[#2132] Migrate SSL support away from ThreadLocal, support key alias and default to pkcs12

mattrpav:
All wiring of transports now takes SslContext as a parameter, instead of 'magically' showing up as a ThreadLocal when needed.

This unlocks a lot of capabilities, including per-transportConnector SSLContext, per-networkConnector SSLContext and eventually auto-reloading of SSLContext.

 - [x] DefaultSslContext (remove active usage of ThreadContextSslContext)
 - [x] Default to stronger and now JDK-default keystore type: 'pkcs12'
 - [x] Add support for keyAlias
 - [x] Add per-connector sslContect
 - [x] Add per-networkConnector sslContext
 - [x] Add ComptibleSslContext to bridge Spring for a non-ThreadLocalSslContext
 - [x] Retain a ThreadContextSslContext class in case someone extended some part of the broker or client (add'l transport connector, etc) and needs a way to transition
 - [x] Unit test refactors

Merge plan: Squash all commits to one for v6.4.0 only (no backport)

### DIFF
--- a/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpNioSslTransportFactory.java
+++ b/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpNioSslTransportFactory.java
@@ -30,7 +30,6 @@ import javax.net.ssl.SSLEngine;
 
 import org.apache.activemq.broker.SslContext;
 import org.apache.activemq.transport.Transport;
-import org.apache.activemq.transport.TransportServer;
 import org.apache.activemq.transport.tcp.TcpTransport;
 import org.apache.activemq.transport.tcp.TcpTransport.InitBuffer;
 import org.apache.activemq.transport.tcp.TcpTransportServer;
@@ -38,10 +37,14 @@ import org.apache.activemq.wireformat.WireFormat;
 
 public class AmqpNioSslTransportFactory extends AmqpNioTransportFactory {
 
-    protected SSLContext context;
-
     @Override
     protected TcpTransportServer createTcpTransportServer(URI location, ServerSocketFactory serverSocketFactory) throws IOException, URISyntaxException {
+        return createTcpTransportServer(location, serverSocketFactory, null);
+    }
+
+    @Override
+    protected TcpTransportServer createTcpTransportServer(URI location, ServerSocketFactory serverSocketFactory, SslContext sslContext) throws IOException, URISyntaxException {
+        final SSLContext context = toSSLContext(sslContext);
         return new TcpTransportServer(this, location, serverSocketFactory) {
             @Override
             protected Transport createTransport(Socket socket, WireFormat format) throws IOException {
@@ -69,17 +72,5 @@ public class AmqpNioSslTransportFactory extends AmqpNioTransportFactory {
             SSLEngine engine, InitBuffer initBuffer, ByteBuffer inputBuffer)
             throws IOException {
         return new AmqpNioSslTransport(wireFormat, socket, engine, initBuffer, inputBuffer);
-    }
-
-    @Override
-    public TransportServer doBind(URI location) throws IOException {
-        if (SslContext.getCurrentSslContext() != null) {
-            try {
-                context = SslContext.getCurrentSslContext().getSSLContext();
-            } catch (Exception e) {
-                throw new IOException(e);
-            }
-        }
-        return super.doBind(location);
     }
 }
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnector.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnector.java
@@ -22,6 +22,7 @@ import java.net.URISyntaxException;
 import java.util.ArrayList;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Optional;
 import java.util.StringTokenizer;
 import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.atomic.AtomicBoolean;
@@ -82,6 +83,7 @@ public class TransportConnector implements Connector, BrokerServiceAware {
     private boolean warnOnRemoteClose = false;
     private boolean displayStackTrace = false;
     private boolean autoStart = true;
+    private SslContext sslContext;
 
     LinkedList<String> peerBrokers = new LinkedList<String>();
     private AtomicBoolean started = new AtomicBoolean(false);
@@ -337,7 +339,8 @@ public class TransportConnector implements Connector, BrokerServiceAware {
             throw new IllegalArgumentException(
                     "You must specify the brokerService property. Maybe this connector should be added to a broker?");
         }
-        return TransportFactorySupport.bind(brokerService, uri);
+        return TransportFactorySupport.bind(brokerService, uri,
+                Optional.ofNullable(sslContext).orElse(brokerService.getSslContext()));
     }
 
     public DiscoveryAgent getDiscoveryAgent() throws IOException {
@@ -702,6 +705,14 @@ public class TransportConnector implements Connector, BrokerServiceAware {
     @Override
     public boolean isAutoStart() {
         return autoStart;
+    }
+
+    public SslContext getSslContext() {
+        return sslContext;
+    }
+
+    public void setSslContext(SslContext sslContext) {
+        this.sslContext = sslContext;
     }
 
     @Override
--- a/activemq-broker/src/main/java/org/apache/activemq/transport/auto/nio/AutoNioSslTransportFactory.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/transport/auto/nio/AutoNioSslTransportFactory.java
@@ -26,6 +26,7 @@ import java.util.Map;
 import java.util.Set;
 
 import javax.net.ServerSocketFactory;
+import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLEngine;
 
 import org.apache.activemq.broker.BrokerService;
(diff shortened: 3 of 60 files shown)


cshannon:
This is a pretty major refactor and needs to be reviewed carefully before merging. I assume this will not be backported and only be targeted for 6.3.0 due to the major changes.

@mattrpav - are there breaking API changes here? a quick look showed old methods delegating to the new ones so I don't think so but we don't want to break anyone who upgrades

mattrpav:
@cshannon correct, this is only for 6.3.0. 

The original SslContext is renamed ThreadLocalSslContext and remains if anyone needs previous behavior exactly.

The ComaptibleSslContext provides same method signature as the ThreadLocalSslContexr for historical compatibility for wiring of arrays v lists, but removes the ThreadLocal usage.

The DefaultThreadLocalContext becomes the new default.

This change removes all ThreadLocalSslContext usage and updates transportConnectors and networkConnectors to support having independent SSLContexts as needed. They all fall back to the broker-wide one as before, if an independent sslcontext is not configured.

activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnector.java
```diff
@@ -337,7 +338,8 @@ protected TransportServer createTransportServer() throws IOException, URISyntaxE
             throw new IllegalArgumentException(
                     "You must specify the brokerService property. Maybe this connector should be added to a broker?");
         }
-        return TransportFactorySupport.bind(brokerService, uri);
+        SslContext ctx = sslContext != null ? sslContext : brokerService.getSslContext();
```
cshannon: These comments fall into the "nit" category and aren't required but I was just thinking this might be a good time to use Optional. This is not on a hot path so it won't create a lot of extra objects to clean up so it might make it a bit cleaner (this comment applies for all the places looking up the ssl context is used with a ternary operator.)  

```java
return TransportFactorySupport.bind(brokerService, uri,
    Optional.ofNullable(sslContext).orElseGet(() -> brokerService.getSslContext()));
```
or 
```java
return TransportFactorySupport.bind(brokerService, uri,
                Optional.ofNullable(sslContext).orElse(brokerService.getSslContext()));
```

Using a helper method:

```java
public SslContext getAppliedSslContext() {
   return sslContext != null ? sslContext : brokerService.getSslContext();
}
```

activemq-client/src/main/java/org/apache/activemq/broker/CompatibleSslContext.java
```diff
@@ -0,0 +1,155 @@
...
+    public String getProvider() {
+        return provider;
+    }
+
+    public void setProvider(String provider) {
+        this.provider = provider;
+    }
+
+    @Override
+    public SSLContext getSSLContext() throws NoSuchProviderException, NoSuchAlgorithmException, KeyManagementException {
```
cshannon, line 134: I need to look at this more but this likely needs to be tweaked because sslContext isn't volatile

cshannon:
I did a quick browse of this and it makes sense to get rid of the thread local if possible but I need to do a deeper dive and try it out. I think there might be a couple things to tweak but overall I think this makes sense. The less thread locals we have the better and as you pointed out it gives us more flexibility in the future.

activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnector.java
mattrpav: This completed w/ the last commit that has changes based on review feedback, and includes the volatile update.

activemq-client/src/main/java/org/apache/activemq/broker/CompatibleSslContext.java
mattrpav: Completed in last commit containing review feedback changes.

activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpNioSslTransportFactory.java
```diff
@@ -73,13 +80,33 @@ public TcpTransport createTransport(WireFormat wireFormat, Socket socket,
...
             } catch (Exception e) {
                 throw new IOException(e);
             }
         }
-        return super.doBind(location);
+        try {
+            Map<String, String> options = new HashMap<String, String>(URISupport.parseParameters(location));
+
+            ServerSocketFactory serverSocketFactory = createServerSocketFactory();
+            TcpTransportServer server = createTcpTransportServer(location, serverSocketFactory, context);
```
cshannon: It looks like you copied this entire block of code from super.doBind() for this one line, is it possible to refactor the parent so we don't have duplicate code?

mattrpav: Yep, good catch

activemq-broker/src/main/java/org/apache/activemq/transport/auto/nio/AutoNioSslTransportFactory.java
```diff
@@ -62,6 +63,12 @@ public void setBrokerService(BrokerService brokerService) {
 
     @Override
     protected AutoNIOSSLTransportServer createTcpTransportServer(URI location, ServerSocketFactory serverSocketFactory) throws IOException, URISyntaxException {
+        return createTcpTransportServer(location, serverSocketFactory, null);
+    }
+
+    @Override
+    protected AutoNIOSSLTransportServer createTcpTransportServer(URI location, ServerSocketFactory serverSocketFactory, SslContext sslContext) throws IOException, URISyntaxException {
+        final SSLContext context = toSSLContext(sslContext);
```
cshannon: I don't think this works, is there a test for the auto transport ssl using a specific context that overrides the broker context?

The issue i see is below it initializes with the SSLEngine, but the context is used to create the engine. The context is not set on the transport until after creation, so it won't be used to generate the SSLEngine.

I'm not sure if this is a pre-existing issue or if it used to work with thread local.

activemq-client/src/main/java/org/apache/activemq/transport/failover/FailoverTransportFactory.java
```diff
@@ -43,6 +45,18 @@ public Transport doConnect(URI location) throws IOException {
         }
     }
 
+    @Override
+    public Transport doConnect(URI location, SslContext sslContext) throws Exception {
```
cshannon: i think you should be able to remove the copy and paste and update the doConnect(location) method to just delegate to this one now with doConnect(location, null)

activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransportFactory.java
```diff
@@ -121,8 +149,72 @@ protected boolean isUseInactivityMonitor(Transport transport) {
...
+        return doConnect(location, (SslContext) null);
+    }
+
+    /**
+     * Connects a TCP based transport. The given {@link SslContext} is handed
+     * to {@link #createSocketFactory(SslContext)} so SSL capable subclasses
+     * can derive their socket factory from it; plain TCP ignores it.
+     */
+    @Override
+    public Transport doConnect(URI location, SslContext sslContext) throws Exception {
```
cshannon: Is there any way to share this logic with TransportFactory?

This is a [copy](https://github.com/apache/activemq/pull/2133/changes#diff-e046a9bfc52ea7eb6ec812e84805d8c33a8ecd1a3ac14e68eacd61102ae8b3bcR155) and we are at risk of getting out of sync by only updating one or the other in the future

activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransportFactory.java
```diff
@@ -121,8 +149,72 @@ protected boolean isUseInactivityMonitor(Transport transport) {
...
+        }
+    }
+
+    @Override
+    public Transport doCompositeConnect(URI location) throws Exception {
+        return doCompositeConnect(location, (SslContext) null);
+    }
+
+    @Override
+    public Transport doCompositeConnect(URI location, SslContext sslContext) throws Exception {
```
cshannon: Same comment here

cshannon (changes requested):
I went through a lot of it but I stopped after I saw there were still several spots where logic could be consolidated and get rid of copy and paste, so that is my main feedback, to reactor to consolidate the logic around connection into one spot, unless there's something that makes it not possible

activemq-client/src/main/java/org/apache/activemq/transport/failover/FailoverTransportFactory.java
mattrpav: Two problems at this specific location:

1. The other doConnect methods throw Exception and not IOException
2. There are two doConnect method with 2-params.

I'm good with dropping the dupe code, but we'd need to be ok w/ wrapping the exception and casting (SslContext)null.

```
  try {
      return doConnect(location, (SslContext) null);
  } catch (Exception e) {
      throw new IOException(e);
  }
```

cshannon: So my next question is why is it different with the exception handling, I didn't notice that but shouldn't all the new methods being throw IOException and not Exception so it's consistent?

cshannon (approved):
LGTM now, assuming CI runs and passes. It's a big update so I expect we may have missed something minor but if we find a tweak to make later we can push a follow on PR.

cshannon (changes requested):
@mattrpav - CI found a real failure, the `DiscoveryTransportBrokerTest` failed

It took me a while to track it down as it was tricky but it turns out during the recent refactor an override got missed, causing the `DiscoveryTransportFactory` to create an instance of `FailoverTransport` instead of `DiscoveryTransport` (that then contained the FailoverTransport).

The fix is pretty simple, I created a quick PR: https://github.com/mattrpav/activemq/pull/8

cshannon (approved):

PR #2136, opened 2026-06-22
Improve the broker's handling of message corruption

cshannon:
This PR improves the broker' handling of messages that are corrupt and can't be read, such as when unmarshaling the properties or body. Currently if there is an error an IOException is triggered and can lead to a client connection be closed. Furthermore for queues messages can be stuck and no new messages can be delivered.

To improve things the following changes have been made:

1. The MarshallingSupport utility that is used to unmarshal message properties and bodies has improve validation to check for errors such is incorrectly encoded size values or a large stack depth of nested objects (limited to 127). This prevents problems such as large buffers being allocated by mistake.
2. The broker will now handle message format errors and there are two main spots errors had to be handled:
    * The first spot is when messages are published and a topic or queue try to match to consumers. The properties (or body for xpath) could be unmarshaled for for selector matching which could trigger an error before the message has been added to the subscription. The previous behavior just skips the message and continues. For topics this is makes sense because each sub is independently handled and will just ack/remove for an unmatching sub and it won't block any other sub. However for the queue case, we can end up in a loop and be "stuck" and need to DLQ.
    * The second spot is later during actual dispatch attempt to clients from the subscriptions if the message made it. The primary cause of failure here is an error tying to convert for another protocol the headers/body. Currently this just triggers an IOException and closes the consumer but doesn't deal with the message. This PR will fix this so that errors that happen that we know are due to a corrupt message will be acked and DLQ'd so they get removed by the queue or durable
  
3. The Stomp protocol converter was fixed to not auto ack or track acks until the message has been converted.
4. AMQP no longer swallows message format errors and will throw so the errors can be handled by the TransportConnection.
5. Any compressed streams that are passed to MarshallingSupport now are wrapped using the new FrameSizeLimitedFilterInputStream class from #2118 so the utility doesn't break during inflation. As noted below a future PR will handle limiting the inflation amount.

All of these changes allow the broker to deal with corrupt messages and remove them (and possibly DLQ) vs causing the connections to close or the messages to block consumers forever in the queue case.

**Note: There will be a follow on PR after this is merged to handle compressed message validation as well that builds on this.**

### DIFF
--- a/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/protocol/AmqpSender.java
+++ b/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/protocol/AmqpSender.java
@@ -22,6 +22,7 @@ import java.io.IOException;
 import java.util.LinkedList;
 import java.util.concurrent.atomic.AtomicInteger;
 
+import org.apache.activemq.ActiveMQMessageFormatException;
 import org.apache.activemq.broker.region.AbstractSubscription;
 import org.apache.activemq.command.ActiveMQDestination;
 import org.apache.activemq.command.ActiveMQMessage;
@@ -41,6 +42,7 @@ import org.apache.activemq.transport.amqp.AmqpProtocolConverter;
 import org.apache.activemq.transport.amqp.message.AutoOutboundTransformer;
 import org.apache.activemq.transport.amqp.message.EncodedMessage;
 import org.apache.activemq.transport.amqp.message.OutboundTransformer;
+import org.apache.activemq.util.ExceptionUtils;
 import org.apache.qpid.proton.amqp.messaging.Accepted;
 import org.apache.qpid.proton.amqp.messaging.Modified;
 import org.apache.qpid.proton.amqp.messaging.Outcome;
@@ -491,7 +493,16 @@ public class AmqpSender extends AmqpAbstractLink<Sender> {
                     }
                 }
             } catch (Exception e) {
-                LOG.warn("Error detected while flushing outbound messages: {}", e.getMessage());
+                // Check if there is a format error trying to convert the message. This error means the
+                // message can't be converted (corruption, etc). This will wrap and throw the message
+                // so it can be handled by the transport
+                ActiveMQMessageFormatException formatError = ExceptionUtils.createMessageFormatException(e);
+                if (formatError != null) {
+                    LOG.warn("Message conversion error while flushing outbound messages: {}", e.getMessage(), e);
+                    throw e;
+                } else {
+                    LOG.warn("Error detected while flushing outbound messages: {}", e.getMessage());
+                }
             }
         }
     }
--- a/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/protocol/AmqpSender.java
+++ b/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/protocol/AmqpSender.java
@@ -499,7 +499,7 @@ public class AmqpSender extends AmqpAbstractLink<Sender> {
                 ActiveMQMessageFormatException formatError = ExceptionUtils.createMessageFormatException(e);
                 if (formatError != null) {
                     LOG.warn("Message conversion error while flushing outbound messages: {}", e.getMessage(), e);
-                    throw e;
+                    throw formatError;
                 } else {
                     LOG.warn("Error detected while flushing outbound messages: {}", e.getMessage());
                 }
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
@@ -31,6 +31,7 @@ import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Set;
 import java.util.concurrent.CancellationException;
 import java.util.concurrent.ConcurrentLinkedQueue;
@@ -98,6 +99,7 @@ import org.apache.activemq.transaction.Synchronization;
 import org.apache.activemq.usage.Usage;
 import org.apache.activemq.usage.UsageListener;
 import org.apache.activemq.util.BrokerSupport;
+import org.apache.activemq.ActiveMQMessageFormatException;
 import org.apache.activemq.util.ThreadPoolUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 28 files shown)


activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/protocol/AmqpSender.java
```diff
@@ -491,7 +493,16 @@ public void pumpOutbound() throws Exception {
...
                 }
             } catch (Exception e) {
-                LOG.warn("Error detected while flushing outbound messages: {}", e.getMessage());
+                // Check if there is a format error trying to convert the message. This error means the
+                // message can't be converted (corruption, etc). This will wrap and throw the message
+                // so it can be handled by the transport
+                ActiveMQMessageFormatException formatError = ExceptionUtils.createMessageFormatException(e);
+                if (formatError != null) {
+                    LOG.warn("Message conversion error while flushing outbound messages: {}", e.getMessage(), e);
+                    throw e;
```
mattrpav: Do you mean to throw formatError here?

activemq-broker/src/main/java/org/apache/activemq/broker/region/QueueSubscription.java
```diff
@@ -100,4 +100,27 @@ protected boolean isDropped(MessageReference node) {
...
+        try {
+            return super.matchesSelector(node, context);
+        } catch (JMSException e) {
+            // This may cause the headers to unmarshal which could throw ActiveMQUnmarshalException
+            // The body could also be unmarshaled if using XPATH and could trigger a
+            // MessageDataFormat exception which this also handles.
+            ActiveMQMessageFormatException error = ExceptionUtils.createMessageFormatException(e);
+            if (error != null) {
+                LOG.error("Message could not be read for selector evaluation: {}", e.getMessage(), e);
+                throw error;
```
mattrpav: probably name this formatError to match variable name above

activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/protocol/AmqpSender.java
cshannon: Yeah this should be fixed, it still works because TransportConnection will catch the original error and then re-parse and rebuild ActiveMQMessageFormatException, however that's a waste so it should throw the new type.

mattrpav (changes requested):

activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
```diff
@@ -67,7 +70,7 @@ public static void marshalPrimitiveMap(Map<String, Object> map, DataOutputStream
     }
 
     public static Map<String, Object> unmarshalPrimitiveMap(DataInputStream in) throws IOException {
-        return unmarshalPrimitiveMap(in, Integer.MAX_VALUE, Integer.MAX_VALUE, Integer.MAX_VALUE);
+        return unmarshalPrimitiveMap(in, Integer.MAX_VALUE, Integer.MAX_VALUE, Byte.MAX_VALUE);
```
mattrpav, line 86: Does this limit properties to 127 (aka Byte.MAX_VALUE). Clarity on this impact could lead to the need to have this be a configurable option v static limit

mattrpav (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -593,7 +597,8 @@ public void removeSubscription(ConnectionContext context, Subscription sub, long
                 }
                 // AMQ-5107: don't resend if the broker is shutting down
                 if (dispatchPendingList.hasRedeliveries() && (! this.brokerService.isStopping())) {
-                    doDispatch(new OrderedPendingList());
+                    messageFormatErrors = new LinkedHashMap<>();
```
mattrpav: Can map allocation be deferred? This would create a lot of allocation+gc for the _vast_ majority of the time there is never a messageFormat error

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -2363,7 +2399,20 @@ protected void assignGroup(Subscription subs, MessageGroupMap messageGroupOwners
     }
 
     protected void pageInMessages(boolean force, int maxPageSize) throws Exception {
-        doDispatch(doPageInForDispatch(force, true, maxPageSize));
+        Map<QueueMessageReference, ActiveMQMessageFormatException> messageFormatErrors = new LinkedHashMap<>();
```
mattrpav: see note on deferring map allocation

mattrpav (dismissed):

activemq-client/src/main/java/org/apache/activemq/util/MarshallingSupport.java
mattrpav: Recall that this is a recursive limit for the number of maps-of-maps or lists-of-lists

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
cshannon: Not easily, we need some way to capture it so the other option is to create an AtomicReference or something to hold it which is still new object collection

cshannon: Maybe i can return something different from the method i'll look at it

tabish121: Could you pass a BiConsumer<QueueMessageReference, ActiveMQMessageFormatException> that stores the errors in a stable object instance that can be interrogated post dispatch for the Map of error and if nothing it there just return an Map.of() ? Afterwards you can reset it for the next dispatch

cshannon: I am playing around with maybe returning something instead to make it easier vs having to mess with passing an object

cshannon: @tabish121 and @mattrpav - Ok i fixed it by using an instance map to capture an errors under the lock. if should almost always just be an empty and not cause any new allocations or GC. If for that page in run it finds some errors it copies to a new map under lock so we can handle it outside the lock.

tabish121 (approved):
I can't spot any obvious issues, after all the fixes that have already been applied it seems reasonable to me.

cshannon:
All tests passed, I am going to merge and wrap up my second follow on PR

Issue #2137, opened 2026-06-23
Unable to use MaxUncommittedCount across a network bridge


mattrpav:
All network-forwarded messages (carry a null transactionId) cause a NullPointerException in
TransactionBroker.verifyUncommittedCount() when maxUncommittedCount is set to a positive value.

This bug renders maxUncommittedCount unusable for any deployment using network connectors.

cshannon:
Can you elaborate what exactly you are actually broken and what you are fixing?

PR #2138, opened 2026-06-23
[#2137] Fix maxUncommittedCount usage across a network of brokers

mattrpav:
Prior to the fix, network-forwarded messages (which have a null transactionId) caused a NullPointerException in
TransactionBroker.verifyUncommittedCount() when maxUncommittedCount was set to a positive value.

### DIFF
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountNetworkTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountNetworkTest.java
@@ -1,0 +1,154 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.usecases;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertTrue;
+
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.concurrent.atomic.AtomicInteger;
+
+import jakarta.jms.JMSContext;
+import jakarta.jms.Message;
+
+import org.apache.activemq.ActiveMQConnectionFactory;
+import org.apache.activemq.broker.BrokerService;
+import org.apache.activemq.command.ActiveMQQueue;
+import org.apache.activemq.util.Wait;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.apache.activemq.test.annotations.ParallelTest;
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+
+/**
+ * Verifies that maxUncommittedCount does not break non-transactional
+ * message forwarding across network connectors.
+ *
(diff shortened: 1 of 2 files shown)


cshannon:
@mattrpav - Can you elaborate? Neither the issue nor the PR actually explains the problem which is not very useful if someone is looking at this and wants to know the point of this change.

mattrpav:
@cshannon before this change, any time a non-zero maxUncommittedCount value was used, error was thrown when using network connector to transfer messages. 

If you run the unit test from this PR on an unpatched tree, you’ll get the error when sending a number of non-transacted messages beyond the maxUncommittedCount

cshannon:
> @cshannon before this change, any time a non-zero maxUncommittedCount value was used, error was thrown when using network connector to transfer messages.
> 
> If you run the unit test from this PR on an unpatched tree, you’ll get the error when sending a number of non-transacted messages beyond the maxUncommittedCount

Thanks, I looked at the test and saw the javadoc which I think explains it well, I will add to the PR is the following:

>Prior to the fix, network-forwarded messages (which have a null transactionId) caused a NullPointerException in
TransactionBroker.verifyUncommittedCount() when maxUncommittedCount was set to a positive value.

Running the test is fine and yes I can review the change and figure out what is going on, but keep in mind we are documenting for the users as well. Someone who upgrades will get release notes and want to know what changed and why and isn't going to have a clue without a description.

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountNetworkTest.java
```diff
@@ -0,0 +1,154 @@
...
+        var brokerBUri = brokerB.getTransportConnectorByScheme("tcp").getPublishableConnectString();
+        var consumerFactory = new ActiveMQConnectionFactory(brokerBUri);
+        var producerFactory = new ActiveMQConnectionFactory(brokerAUri);
+
+        try (final var consumerContext = consumerFactory.createContext(transacted ? JMSContext.SESSION_TRANSACTED : JMSContext.AUTO_ACKNOWLEDGE);
+             final var consumer = consumerContext.createConsumer(consumerContext.createQueue(queueName));
+             final var producerContext = producerFactory.createContext(transacted ? JMSContext.SESSION_TRANSACTED : JMSContext.AUTO_ACKNOWLEDGE)) {
+
+            assertTrue("Remote consumer should come online",
+                    Wait.waitFor(() -> {
```
cshannon, line 122: Can you switch the sleep times? I think by default it will sleep for 1 second and we can probably speed that up to 10ms just so the tests can run faster. It actually probably makes sense to update the utility at some point to be a quicker default as usually 1 second is kind of slow for a retry.

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountNetworkTest.java
```diff
@@ -0,0 +1,154 @@
...
+                                  producerContext.createTextMessage("msg-" + i));
+            }
+
+            if (transacted) {
+                producerContext.commit();
+            }
+
+            final var count = new AtomicInteger(0);
+            assertTrue("All messages should arrive across the network connector",
+                    Wait.waitFor(() -> {
```
cshannon, line 140: Same with speeding up the retry for the wait here.

cshannon (changes requested):

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountNetworkTest.java
mattrpav: Done

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/MaxUncommittedCountNetworkTest.java
mattrpav: Done

mattrpav:
Shorter interval timing applied:

```
[INFO]  T E S T S
[INFO] ---org.apache.activemq.usecases.MaxUncommittedCountNetworkTest - 2.555 s
[INFO]    +-- [OK] testSendAcrossNetworkConnector[transacted=true] - 1.498 s
[INFO]    '-- [OK] testSendAcrossNetworkConnector[transacted=false] - 1.049 s
```

cshannon (approved):

PR #2139, opened 2026-06-23
Add support for maxInflateDataSize to limit uncompressed buffers

cshannon:
This change adds support for limiting the maximum inflation size of a message body when a message needs to be decompressed. This new setting will help prevent OOM errors from large buffers being allocated.

The primary concern this is addressing is that compressed messages may be smaller than maxFrameSize and the broker will accept them, but if an event triggers a decompression a huge buffer could be created and cause OOM.

Most of the time the broker can avoid decompression but sometimes it is necessary. A message will need to be decompressed primarily when consumers come online for other other protocols (such as stomp or amqp) because the message needs to be converted. Other more rare instances include xpath selectors or custom plugins.

The broker will have a new `maxInflateDataSize` config that is broker wide because the value isn't tied to a protocol and it will default to 100 MB. The default XML shipped with the broker for maxFrameSize is set to 10MB so this makes sense as it is 10x which will match the client default. Because clients are tied to a transport, it makes more sense to configure maxInflatedDataSize as a ratio of maxFrameSize. The default is 10x maxFrameSize if configured which should be enough of a buffer under normal circumstances. The ratio can be changed using `maxInflatedDataSizeRatio` on a connection.

The validation of maxInflatedDataSize is handled by checking the buffer size before allocation if known, or by taking advantage of FrameSizeLimitedFilterInputStream and wrapping InflaterInputStream using maxInflateDataSize as the limit so an exception will be thrown if the size is too large. Errors thrown will be handled by the same improved error handling added in #2136 so that messages can be removed and sent to the DLQ if they can't be dispatched.

### DIFF
--- a/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/message/AmqpMessageSupport.java
+++ b/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/message/AmqpMessageSupport.java
@@ -18,6 +18,7 @@ package org.apache.activemq.transport.amqp.message;
 
 import java.io.DataInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.ObjectOutputStream;
 import java.io.Serializable;
 import java.nio.charset.Charset;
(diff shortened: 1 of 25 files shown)


mattrpav (dismissed):
LGTM, pending tests completion.

tabish121 (dismissed):
Looks good

tabish121 (approved):

PR #2154, opened 2026-06-23
Improve handling of partial reads and EOF in Frame size limited filter

cshannon:
This update makes the following changes/improvements:

* If the wrapped stream is finished and returns -1 this is now returned correctly and availableBytes are not decremented.
* Partial reads now will be attempted up to the availableBytes limit
* If 0 is passed in for the length for a partial read then no attempt to read is done and 0 is returned, in line with the spec.


tabish121 (approved):
Changes seems reasonable for getting this a bit more aligned with a standard stream type

PR #2165, opened 2026-06-30
Prevent cursor from using more than 100% temp store

cshannon:
This fixes a bug in `FilePendingMessageCursor` that could cause the temporary store to exceed 100% usage. The cursor now properly checks the temporary usage limits when writing messages from memory to the temporary store. If the store is full, messages are now discarded and a discarded advisory will be fired (if enabled) and the messages will be DLQ'd if the `processNonPersistent` flag is true. This commit also fixes an issue with Topic subs possibly sending expired messages to the DLQ twice which was noticed while working on this fix.

While not a common scenario, previously it was possible to blow past the limits for the temporary store because FilePendingMessageCursor did not check if usage was 100%. This could happen if there was a small temporary store configured, or just simply a lot of destinations using temporary usage. In the worst case scenario this could mean filling up a partition and causing an IO exception and broker crash, best case it would still be bad because it uses up disk space meant for persistent messaging.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
@@ -588,7 +588,7 @@ public class FilePendingMessageCursor extends AbstractPendingMessageCursor imple
 
     private void discardMessage(MessageReference reference) {
         try {
-            LOG.debug("Discarding message {}", reference);
+            LOG.debug("Discarding message {} due to full temporary storage {}", reference, systemUsage.getTempUsage());
             ConnectionContext context = broker.getAdminConnectionContext();
             // This will handle advisories if advisoryForDiscardingMessages is enabled
             // as well as DLQ processing
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
@@ -17,10 +17,12 @@
 package org.apache.activemq.broker.region.cursors;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.Iterator;
+import java.util.LinkedHashMap;
 import java.util.LinkedList;
-import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 
@@ -30,6 +32,7 @@ import org.apache.activemq.broker.region.Destination;
 import org.apache.activemq.broker.region.IndirectMessageReference;
 import org.apache.activemq.broker.region.MessageReference;
 import org.apache.activemq.broker.region.QueueMessageReference;
+import org.apache.activemq.broker.region.Subscription;
 import org.apache.activemq.command.Message;
 import org.apache.activemq.openwire.OpenWireFormat;
 import org.apache.activemq.store.PList;
@@ -54,9 +57,10 @@ public class FilePendingMessageCursor extends AbstractPendingMessageCursor imple
     private static final AtomicLong NAME_COUNT = new AtomicLong();
 
     protected Broker broker;
+    private final Subscription sub;
     private final PListStore store;
     private final String name;
-    private PendingList memoryList;
+    private final PendingList memoryList;
     private PList diskList;
     private Iterator<MessageReference> iter;
     private Destination regionDestination;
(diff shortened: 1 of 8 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
```diff
@@ -516,12 +561,44 @@ public PList getDiskList() {
         return diskList;
     }
 
+    private void discardMessage(MessageReference reference, DiscardType discardType) {
+        if (discardType == DiscardType.EXPIRED) {
```
mattrpav, line 565: nit:

```
DiscardType.EXPIRED.equals(discardType)
``

File: activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java [hunk 2]
```diff
@@ -516,12 +561,44 @@ public PList getDiskList() {
...
-            context.setBroker(broker);
-            ((Destination)reference.getRegionDestination()).messageExpired(context, null, new IndirectMessageReference(reference.getMessage()));
+        try {
+            if (reference.isExpired() && broker.isExpired(reference)) {
+                ConnectionContext context = broker.getAdminConnectionContext();
+                ((Destination) reference.getRegionDestination()).messageExpired(context, sub,
+                        new IndirectMessageReference(reference.getMessage()));
+            }
+        } catch (Exception e) {
+            LOG.warn("Error discarding expired message {}", reference.getMessageId());
```
mattrpav, line 581: I think two log messages for one event is an anti-pattern of sorts.

Can we simply log the exception message and the exception here? the log appender format will determine if the stacktrace is spit out in the log

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
```diff
@@ -516,12 +561,44 @@ public PList getDiskList() {
...
+    private void discardMessage(MessageReference reference) {
+        try {
+            LOG.debug("Discarding message {}", reference);
+            ConnectionContext context = broker.getAdminConnectionContext();
+            // This will handle advisories if advisoryForDiscardingMessages is enabled
+            // as well as DLQ processing
+            ((Destination)reference.getRegionDestination()).messageDiscarded(context, sub, new IndirectMessageReference(reference.getMessage()));
+        } catch (Exception e) {
+            LOG.warn("Error discarding message {}", reference.getMessageId());
+            LOG.debug(e.getMessage(), e);
```
mattrpav, line 598: double log

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
```diff
@@ -516,12 +561,44 @@ public PList getDiskList() {
...
+        } finally {
+            // we can now drop the reference count
+            reference.decrementReferenceCount();
+        }
+    }
+
+    private void discardMessage(MessageReference reference) {
+        try {
+            LOG.debug("Discarding message {}", reference);
```
mattrpav: nit: I think the log should say _why_ it is discarding in the message -- temp space full, temp space disabled, expired, etc

mattrpav (changes requested):

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
cshannon: There's really no way to avoid double logging because the default log4j appender shipped with the broker doesn't configure formats to filter out exceptions. So logging a stack trace as part of warn or error would be super annoying. We need to log just the message as warn/error and then debug can log the full trace.

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
cshannon: See other comment

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
cshannon: This is actually not recommended.

== is generally recommended for enum values for 2 reasons:

1. You get null check safety
2. The code won't compile if you screw up and mix enum types, where as .equals() takes an object and will compile even if you used the wrong type that will never be equals.

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
mattrpav: No worries, we'll have to tackle logging optimization in 7.x

activemq-broker/src/main/java/org/apache/activemq/broker/region/cursors/FilePendingMessageCursor.java
cshannon: @mattrpav - I updated the log message

mattrpav (approved):
LGTM

PR #2166, opened 2026-07-01
Lower default maxFrameSize to 10 MB in xml config

cshannon:
The previous default was 100 MB but 10 MB makes more sense as a default, especially with maxInflatedDataSize defaulting to 100 MB.

This only adjusts the XML config for now, we could also change the java code default in another PR (right now there is no default it is just Long.MAX)


tabish121 (approved):
LGTM

PR #2167, opened 2026-07-01
Add better frame size validation for AMQP

cshannon:
Add a negative size check for AMQP NIO transports. This also adds a similar check to OpenWire. For OpenWire the same IllegalArgumentException is thrown but this improves the error message in the NIO transport.

### DIFF
--- a/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpFrameParser.java
+++ b/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpFrameParser.java
@@ -84,9 +84,9 @@ public class AmqpFrameParser {
             maxFrameSize = wireFormat.getMaxFrameSize();
         }
 
-        if (frameSize > maxFrameSize) {
+        if (Integer.toUnsignedLong(frameSize) > maxFrameSize) {
             throw IOExceptionSupport.createFrameSizeException(frameSize, maxFrameSize);
-        } else if (frameSize <= 0) {
+        } else if (Integer.compareUnsigned(frameSize, 8) < 0) {
             throw new AmqpProtocolException("Frame size value was invalid: " + frameSize);
         }
     }
--- a/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpFrameParser.java
+++ b/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpFrameParser.java
@@ -86,6 +86,8 @@ public class AmqpFrameParser {
 
         if (frameSize > maxFrameSize) {
             throw IOExceptionSupport.createFrameSizeException(frameSize, maxFrameSize);
+        } else if (frameSize <= 0) {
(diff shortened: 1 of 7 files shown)


activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/AmqpFrameParser.java
```diff
@@ -86,6 +86,8 @@ private void validateFrameSize(int frameSize) throws IOException {
 
         if (frameSize > maxFrameSize) {
             throw IOExceptionSupport.createFrameSizeException(frameSize, maxFrameSize);
+        } else if (frameSize <= 0) {
```
tabish121: This could be improved a bit as the lower limit on frame size is actually 8 bytes and the type of the size value is an unsigned int so you should actually be using `Integer.compareUnsigned()` if you wanted to be precise on what went wrong.

This is an example from a Qpid proton frame decoder:
```
        private void validateFrameSize() throws FrameDecodingException {
            if (Integer.compareUnsigned(frameSize, 8) < 0) {
               throw new FrameDecodingException(String.format(
                    "specified frame size %d smaller than minimum frame header size 8", frameSize));
            }

            if (Integer.toUnsignedLong(frameSize) > configuration.getInboundMaxFrameSize()) {
                throw new FrameDecodingException(String.format(
                    "specified frame size %s larger than maximum frame size %d",
                    Integer.toUnsignedString(frameSize), configuration.getInboundMaxFrameSize()));
            }
        }
```

cshannon: I updated it and also applied to the blocking tansport as well in AmqpWireFormat

tabish121 (dismissed):
The changes seem fine, unsure how well covered this is in the test suite though so not sure if you want to add any additional tests for these conditions.

cshannon:
> The changes seem fine, unsure how well covered this is in the test suite though so not sure if you want to add any additional tests for these conditions.

I didn't add anything because I didn't think it would be very easy to fake a negative value for AMQP with the client, unless you have any suggestions. I suppose just doing some mocking or calling the validation directly would be the only option otherwise.

tabish121:
> > The changes seem fine, unsure how well covered this is in the test suite though so not sure if you want to add any additional tests for these conditions.
> 
> I didn't add anything because I didn't think it would be very easy to fake a negative value for AMQP with the client, unless you have any suggestions. I suppose just doing some mocking or calling the validation directly would be the only option otherwise.

Why would you need a client when AmqpWireFormat 

> > The changes seem fine, unsure how well covered this is in the test suite though so not sure if you want to add any additional tests for these conditions.
> 
> I didn't add anything because I didn't think it would be very easy to fake a negative value for AMQP with the client, unless you have any suggestions. I suppose just doing some mocking or calling the validation directly would be the only option otherwise.

I don't think you'd need a client or proton to test it since the NIO bit reads from a DataInput and the AmqpFrameParser reads from a ByteBuffer so you can just simulate the values

tabish121 (approved):
LGTM

PR #2177, opened 2026-07-02
Fire message discarded advisory on format errors

cshannon:
In #2136 the broker was updated to handle corrupt messages and to send them to the DLQ and discarding them. This change further improves on that by sending the message discarded advisory, if it has been enabled on the destination. By default the discard advisory will be off.


mattrpav (approved):
LGTM, pending tests.

PR #2185, opened 2026-07-03
chore: Update Spring to 6.2.19

coheigea:


jbonofre:
It should be tackled by dependabot.

jbonofre (approved):

PR #2188, opened 2026-07-06
Fix TextMessage body thread safety

cshannon:
This fixes the rare null body issues seen on the broker when using **text** messages by using synchronization when marshaling or unmarshaling the body into a String to prevent a race condition from causing the body to end up null. This has been optimized to use volatiles and double checked locking to avoid synchronization unless needed.

For some background, this has been a long running issue that has come up over the years and there have been many proposed ways to fix it (see #1851 and #1659 for examples). Despite the fact that these messages are supposed to be copied and not used across multiple threads there are cases where that is not true and can cause issues (topic consumers on dispatch for example don't copy) so this issue still pops up from time to time.

After looking at all the options and how disruptive it would be, I think this is the best approach for now as a short term fix. I think long term we could look at better solutions. I think ideally on the server side it would be nice to use messages that only store the data in the marshaled format as the vast majority of the the time the body should not need to be unmarshaled anyways (exceptions include protocol conversion or logging). The client versions could still use mutable messages that can store either version.

**Notes on other message types and properties:**

1. This PR only attempts to fix the body for ActiveMQTextMessage because this is **not a problem for the body of other message types** (including Map, object message, etc). This issue is specific to ActiveMQTextMessage because it is the only message type that will only store either the unmarshaled data (text) or the bytes, but not both. Other message types do not clear the content after unmarshaling. This means that if multiple threads try and do a conversion on a message like Map or Object message, it may convert twice, but you won't get null. The other message types only clear unmarshaled content at certain points in the broker if reduceMemoryFootprint is true.
2. The properties map also changes state between unmarshalled and marshaled so it needs to be investigated more, but it looks like we may be ok for now. Just like the other non-text messages, when getting properties if the map is null it will unmarshall but it won't clear the original content. This is confirmed by the fact that no one has reported null properties.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQTextMessage.java
+++ b/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQTextMessage.java
@@ -51,13 +51,11 @@ public class ActiveMQTextMessage extends ActiveMQMessage implements TextMessage 
     @Override
     public Message copy() {
         ActiveMQTextMessage copy = new ActiveMQTextMessage();
-        copy(copy);
+        synchronized (this) {
+            super.copy(copy);
+            copy.text = text;
+        }
         return copy;
-    }
-
-    private synchronized void copy(ActiveMQTextMessage copy) {
-        super.copy(copy);
-        copy.text = text;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


cshannon:
@mattrpav - I might close this out and open up another PR, there's still issues here because of needing to read 2 variables. There's still race conditions where you could read two nulls outside the lock during mutation.

I'm going to prototype something with a ReadWriteLock intead to see what it looks like

cshannon:
I marked this as a draft as I might be reworking this

cshannon:
Using a Reentrant lock doesn't make sense because of the memory usage and the added complexity. I went ahead and reworked this a little bit to fix the race condition by ensuring the decision to sync or not is only checked after reading one variable and not two. For example, getText() only relies on whether text is null and will synchronize if it is. This eliminates the race condition of reading text and content and then trying to determine their state outside of a lock when another thread could be mutating both.

mattrpav (dismissed):
LGTM. Passed the contention test that detected the issue

jbonofre (approved):

Issue #2196, opened 2026-07-07
Queue consumer dispatch fix


mattrpav:
A queue consumer starvation may occur when consumers are slow or blocked and the maxPageSize is low in relative relation to the number of consumers and the consumers' prefetch.

PR #2197, opened 2026-07-07
[#2196] Improve queue consumer starvation under low configuration

mattrpav:
Run the test before the fix and observe:

```
2026-07-07 15:56:20,119 [me-limited test] - INFO  PriorityDispatchStarvationTest - Produced 200 messages (40 app, 160 other)
2026-07-07 15:56:30,154 [me-limited test] - INFO  PriorityDispatchStarvationTest - App consumer received 8 / 40 messages (drained=false)
2026-07-07 15:56:30,156 [me-limited test] - INFO  PriorityDispatchStarvationTest - Queue stats - queueSize: 192, enqueues: 200, dequeues: 8, inflight: 0, dispatched: 8
```

Test run after fix:
```
2026-07-07 15:59:33,076 [me-limited test] - INFO  PriorityDispatchStarvationTest - Produced 200 messages (40 app, 160 other)
2026-07-07 15:59:33,110 [ Session Task-1] - INFO  PriorityDispatchStarvationTest - App consumer received 10 / 40 messages
2026-07-07 15:59:33,115 [ Session Task-1] - INFO  PriorityDispatchStarvationTest - App consumer received 20 / 40 messages
2026-07-07 15:59:33,119 [ Session Task-1] - INFO  PriorityDispatchStarvationTest - App consumer received 30 / 40 messages
2026-07-07 15:59:33,123 [ Session Task-1] - INFO  PriorityDispatchStarvationTest - App consumer received 40 / 40 messages
2026-07-07 15:59:33,123 [me-limited test] - INFO  PriorityDispatchStarvationTest - App consumer received 40 / 40 messages (drained=true)
2026-07-07 15:59:33,124 [me-limited test] - INFO  PriorityDispatchStarvationTest - Queue stats - queueSize: 160, enqueues: 200, dequeues: 40, inflight: 0, dispatched: 40
```

edit: Tests passed, except for assembly on an address-in-use bind error. Re-running tests to get a green checkmark

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
@@ -2113,8 +2113,13 @@ public class Queue extends BaseDestination implements Task, UsageListener, Index
             pagedInPendingDispatchLock.readLock().unlock();
         }
         if (isLazyDispatch() && !force) {
-            // Only page in the minimum number of messages which can be
-            // dispatched immediately.
+            toPageIn = Math.min(toPageIn, getConsumerMessageCountBeforeFull());
+        }
+        if (pagedInPendingSize >= maxPageSize && !force) {
+            // When the pending dispatch list is full, only page in what
+            // consumers can actually accept. This prevents unbounded growth
+            // while still allowing consumers with available prefetch capacity
+            // to receive messages even when other consumers are full.
             toPageIn = Math.min(toPageIn, getConsumerMessageCountBeforeFull());
         }
 
@@ -2132,7 +2137,8 @@ public class Queue extends BaseDestination implements Task, UsageListener, Index
                     maxPageSize);
         }
 
-        if (toPageIn > 0 && (force || (haveRealConsumer() && pagedInPendingSize < maxPageSize))) {
+        if (toPageIn > 0 && (force || (haveRealConsumer() && (pagedInPendingSize < maxPageSize
+                || getConsumerMessageCountBeforeFull() > 0)))) {
             int count = 0;
             result = new ArrayList<QueueMessageReference>(toPageIn);
             messagesLock.writeLock().lock();
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/bugs/NetworkConsumerPriorityDispatchStarvationTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/bugs/NetworkConsumerPriorityDispatchStarvationTest.java
@@ -1,0 +1,216 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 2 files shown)


activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -2113,8 +2113,13 @@ private PendingList doPageInForDispatch(boolean force, boolean processExpired, i
             pagedInPendingDispatchLock.readLock().unlock();
         }
         if (isLazyDispatch() && !force) {
-            // Only page in the minimum number of messages which can be
-            // dispatched immediately.
+            toPageIn = Math.min(toPageIn, getConsumerMessageCountBeforeFull());
```
cshannon, line 2116: getConsumerMessageCountBeforeFull() is an expensive operation and has to iterate, it should be called one time and not 3 times and result saved

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
```diff
@@ -2113,8 +2113,13 @@ private PendingList doPageInForDispatch(boolean force, boolean processExpired, i
...
+        }
+        if (pagedInPendingSize >= maxPageSize && !force) {
```
cshannon, line 2118: Shouldn't this be an if/else so you aren't doing the same thing twice?

cshannon (changes requested):
I only took a quick look so far but I don't really understand why you don't just set [lazyDispatch](https://activemq.apache.org/components/classic/documentation/per-destination-policies) to true? The point of that flag is to do exactly this where it will page in what consumers can handle.

mattrpav:
> I only took a quick look so far but I don't really understand why you don't just set [lazyDispatch](https://activemq.apache.org/components/classic/documentation/per-destination-policies) to true? The point of that flag is to do exactly this where it will page in what consumers can handle.

LazyDispatch vs this change comes down to when force = true. I think you are right that lazy dispatch would handle the normal consumer, since it looks for force=false scenarios. I'll do a pass through cursor scenarios. iirc the force=true scenarios were the management use cases.

edit: lazyDispatch does not provide relief. The issue is this must also evaluate to true:

```
if (toPageIn > 0 && (force || (haveRealConsumer() && pagedInPendingSize < maxPageSize))) {
```

When the pageSize is super low and the pagedInPendingSize is larger, this will evaluate false.

That being said, I'm not convinced taking the penalty to calculate if consumers have space makes sense to put on everyone's queue, every time for this one scenario. I'm leaning towards the best approach is to not starve the queue with a small maxPageSize.

cshannon:
I also have concerns that this could page in a lot more into memory and cause a huge memory spike if you are using count before full. The default prefetch is 32k and obviously that should be tuned but it could cause a large increase in paged in messages if there are a lot of consumers even with a prefetch not that high

cshannon:
> I also have concerns that this could page in a lot more into memory and cause a huge memory spike if you are using count before full. The default prefetch is 32k and obviously that should be tuned but it could cause a large increase in paged in messages if there are a lot of consumers even with a prefetch not that high

This might be fine the more i think about it (after all we want to keep prefetch full), i just want to take a closer look as it has been a while since i've looked at all the logic with paging stuff in

mattrpav:
> I also have concerns that this could page in a lot more into memory and cause a huge memory spike if you are using count before full. The default prefetch is 32k and obviously that should be tuned but it could cause a large increase in paged in messages if there are a lot of consumers even with a prefetch not that high

Note: queuePrefetch default is 1,000, but yeah the overall concern is valid.

Its an odd scenario and it looks like it trips up on the somewhat side-ways logic.

memoryUsage: has space
sum of consumer prefetch: has space
maxPageSize: less than the number of paged in messages.

The conditional comparing current pending count vs pageSize is a context switch in the comparison logic. Smells like an apples-to-oranges type deal rotating around counts vs space in memory usage.

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
jbonofre: I think also `dispatchPendingList` trigger `maxPageSize` each cycle (related).

Issue #2198, opened 2026-07-07
Fix PendingSends counter for a rolled-back send


mattrpav:
The internal (not-exposed) pendingSends metric does not decrement on a rolled back transacted send

PR #2199, opened 2026-07-07
[#2198] Fix pending sends count on rolled back send

mattrpav:
This was uncovered while looking into a different issue. 

The problem here is that pending sends counter (internal, not exposed) does not decrement on a rolled back producer transaction.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
@@ -125,7 +125,7 @@ public class Queue extends BaseDestination implements Task, UsageListener, Index
     // this is guarded by pagedInPendingDispatchLock
     private final Map<QueueMessageReference, ActiveMQMessageFormatException> dispatchMessageFormatErrors = new LinkedHashMap<>();
     protected QueueDispatchPendingList dispatchPendingList = new QueueDispatchPendingList();
-    private AtomicInteger pendingSends = new AtomicInteger(0);
+    private final AtomicInteger pendingSends = new AtomicInteger(0);
     private MessageGroupMap messageGroupOwners;
     private DispatchPolicy dispatchPolicy = new RoundRobinDispatchPolicy();
     private MessageGroupMapFactory messageGroupMapFactory = new CachedMessageGroupMapFactory();
@@ -888,6 +888,7 @@ public class Queue extends BaseDestination implements Task, UsageListener, Index
             if (store != null && messageContext.message.isPersistent()) {
                 rollbackPendingCursorAdditions(messageContext.message.getMessageId());
             }
+            pendingSends.decrementAndGet();
             messageContext.message.decrementReferenceCount();
         }
     }
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/bugs/PendingSendsRollbackLeakTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/bugs/PendingSendsRollbackLeakTest.java
@@ -1,0 +1,183 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.activemq.bugs;
+
+import static org.junit.Assert.assertEquals;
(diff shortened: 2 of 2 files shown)


cshannon:
@mattrpav - I have asked this several times before, can you please start adding actual information to your PRs and issues? 

You added a bunch of info to the test itself which explains the problem, some of that needs to be in the description so people can figure out what the point of the PR is without having to look at the test.

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
cshannon:
```suggestion
    private final AtomicInteger pendingSends = new AtomicInteger(0);
```

activemq-unit-tests/src/test/java/org/apache/activemq/bugs/PendingSendsRollbackLeakTest.java
```diff
@@ -0,0 +1,182 @@
...
+ * On rollback, CursorAddSync.afterRollback() must also decrement it —
+ * without that, every rolled-back transactional send permanently inflates
+ * pendingSends.
+ *
+ * The leaked counter makes Queue.singlePendingSend() return false forever,
+ * which disables cursor caching via QueueStorePrefetch.canEnableCash().
+ * The queue is then stuck in store page-in mode, degrading throughput and
+ * amplifying duplicateFromStore events.
+ */
+public class PendingSendsRollbackLeakTest {
```
cshannon, line 60:
```suggestion
@Category(ParallelTest.class)
public class PendingSendsRollbackLeakTest {
```

cshannon (changes requested):
LGTM, this is a good catch, just a couple minor tweaks that I noticed that could be done

cshannon (approved):
LGTM to merge, assuming CI looks good. I'm always amazed when stuff like this has been hanging around for such a long time and no one noticed. Another example is #2178 that was broken for a long time

activemq-unit-tests/src/test/java/org/apache/activemq/bugs/PendingSendsRollbackLeakTest.java
```diff
@@ -0,0 +1,183 @@
...
+        assertTrue("pendingSends should return to 0 after commit",
+                Wait.waitFor(() -> getPendingSends(queue) == 0, 5000, 100));
+        assertTrue("Queue should have " + ROLLBACK_SEND_COUNT + " messages after commit",
+                Wait.waitFor(() -> queue.getDestinationStatistics().getMessages().getCount() == ROLLBACK_SEND_COUNT,
+                        5000, 100));
+    }
+
+    private int getPendingSends(Queue queue) {
+        try {
+            var f = Queue.class.getDeclaredField("pendingSends");
```
jbonofre, line 176: nit: instead of reflection, we could have added an accessor with `@VisibleForTesting` in `Queue` to avoid coupling the test.

jbonofre (approved):
I suspect that we could have double-decrement `pendingSends` in XA. The `singlePendingSend()` uses <= 1 so negative values can pass the test. Neither `fireAfterCommit()` and `fireAfterRollback()` has a guard preventing both from running on the same sync list.
That's OK, and we can always improve in a follow-up PR.

activemq-unit-tests/src/test/java/org/apache/activemq/bugs/PendingSendsRollbackLeakTest.java
cshannon: I almost mentioned this too, and I would have suggested this if we could use protected/package scope as an accessor but it's a different package so it would have to be public which I wasn't sure was a good idea, so i was ok with the reflection in this case

activemq-broker/src/main/java/org/apache/activemq/broker/region/Queue.java
cshannon, line 904: I just realized that checkUsage() can throw an exception. If that happens, the counter gets incremented but it would never be decremented so I think that needs to be fixed too. We might just need a separate try/catch to wrap pendingSends.increment if we need to increment before checkUsage. I haven't looked at the producerExchange.incrementSend() but i am wondering if that is a problem too with missing decrement on exception.

cshannon (changes requested):

cshannon:
> I suspect that we could have double-decrement `pendingSends` in XA. The `singlePendingSend()` uses <= 1 so negative values can pass the test. Neither `fireAfterCommit()` and `fireAfterRollback()` has a guard preventing both from running on the same sync list. That's OK, and we can always improve in a follow-up PR.

With the current design, I am not sure if there is a great way to guarantee we don't miss a decrement or double decrement using a simple counter like it is doing now. The problem is there is so many code paths for the decrement (async etc) and possible error scenarios it would be hard to ensure it's correct.

I think a future PR to refactor to just try and simplify and reduce the number of code paths where increment/decrement happen would help provide better guarantees it's correct. One way to handle the double decrement case would be to to track pendingSend message Ids in a Set instead of using a counter and just remove instead of decrement. However, that doesn't help missing a decrement and would cause a memory leak so not sure that makes sense either.

activemq-unit-tests/src/test/java/org/apache/activemq/bugs/PendingSendsRollbackLeakTest.java
mattrpav: I think a long getPendingSendCount() getter makes sense -- this would also provide observability if exposed via JMX and closes the last remaining gap in the message lifecycle.

WDYT?

Issue #2119, opened 2026-06-15
ActiveMQ does not start without Internet access


lichtin:
ActiveMQ (6.2.6 in this case) does not start if there is no Internet access.
This seems due to a missing entry in the "META-INF/spring.schemas" file. Upon adding the following line

   http\://www.springframework.org/schema/beans/spring-beans.xsd=org/springframework/beans/factory/xml/spring-beans.xsd

then ActiveMQ is able to start without reaching out to the Internet.

2026-06-07T10:00:27,414 | WARN  | =[org.apache.activemq.server]) | XmlBeanDefinitionReader    | k.util.xml.SimpleSaxErrorHandler   48 | org.springframework.__version_6.2.17 - 0.0.0 | Ignored XML validation warning
org.xml.sax.SAXParseException; lineNumber: 3; columnNumber: 100; schema_reference.4: Failed to read schema document 'http://www.springframework.org/schema/beans/spring-beans.xsd', because 1) could not find the document; 2) the document could not be read; 3) the root element of the document is not <xsd:schema>.
	at java.xml/com.sun.org.apache.xerces.internal.util.ErrorHandlerWrapper.createSAXParseException(ErrorHandlerWrapper.java:204)
	at java.xml/com.sun.org.apache.xerces.internal.util.ErrorHandlerWrapper.warning(ErrorHandlerWrapper.java:100)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLErrorReporter.reportError(XMLErrorReporter.java:392)
	at java.xml/com.sun.org.apache.xerces.internal.impl.XMLErrorReporter.reportError(XMLErrorReporter.java:306)
	at java.xml/com.sun.org.apache.xerces.internal.impl.xs.traversers.XSDHandler.reportSchemaErr(XSDHandler.java:4258)
	at java.xml/com.sun.org.apache.xerces.internal.impl.xs.traversers.XSDHandler.reportSchemaWarning(XSDHandler.java:4249)
	at java.xml/com.sun.org.apache.xerces.internal.impl.xs.traversers.XSDHandler.getSchemaDocument1(XSDHandler.java:2543)
	at java.xml/com.sun.org.apache.xerces.internal.impl.xs.traversers.XSDHandler.getSchemaDocument(XSDHandler.java:2239)
	at java.xml/com.sun.org.apache.xerces.internal.impl.xs.traversers.XSDHandler.parseSchema(XSDHandler.java:589)
[...]

PR #2201, opened 2026-07-08
[#2119] Update spring.schemas to include missing 'spring-beans.xsd' mapping

lichtin:


jbonofre:
@lichtin can you provide the rationale for this? I guess it's because you extend the `activemq.xml` with spring beans?

lichtin:
It's the fix for issue #2119 as discussed on the mailing list.
Without this change, the broker does not startup in an isolated environment.

jbonofre:
@lichtin ah yes, the "offline" support. Thanks!

jbonofre:
Please, create the PR based on `main` branch.

PR #2207, opened 2026-07-09
Add tests validating TextMessage body thread safety fix

cshannon:
@mattrpav authored this and I forgot to merge into #2188 so I'm pushing up a new PR with the commit

Validates the fix for the race between getText(), storeContentAndClear() and copy() on a shared ActiveMQTextMessage instance that could null the message body: getText() decoded content into text and cleared content while storeContentAndClear() observed the pre-clear content, skipped re-encoding, and cleared text — losing the body permanently. Triggered on the broker by XPath selector evaluation or JMX browse racing dispatch marshalling, with the network bridge copy() adding a third participant.

Four tests:
- sequential control: getText/storeContentAndClear round trips preserve the body
- two-way race: 20k iterations of concurrent getText vs storeContentAndClear; no iteration may lose the body
- three-way race including copy(): both the original and the copy must retain a recoverable body (pre-fix this loses the original's body within a few thousand iterations)
- deterministic lock-discipline pin: the fix serializes the text/content state transitions on the message instance, so storeContentAndClear() must block while another thread holds the message monitor; fails immediately on any regression to unsynchronized transitions

Verified failing (2 of 4) against the pre-fix base and passing after #2188

### DIFF
--- a/activemq-client/src/test/java/org/apache/activemq/command/ActiveMQTextMessageContentRaceTest.java
+++ b/activemq-client/src/test/java/org/apache/activemq/command/ActiveMQTextMessageContentRaceTest.java
@@ -1,0 +1,306 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


mattrpav (approved):
Approved

PR #2208, opened 2026-07-09
[6.2.x] feat(#1828): Expose NetworkConnector URI and local URI in JMX MBean (activemq-6.2.x)

jbonofre:
Backport of #1829 to the `activemq-6.2.x` branch.

## Summary

- Add `getUri()` and `getLocalUri()` to `NetworkConnectorViewMBean` so the network connector URI is visible via JMX.
- Add `abstract getUri()` to the `NetworkConnector` base class.
- Implement `getUri()` in `MulticastNetworkConnector` (delegates to `getRemoteURI()`).
- `DiscoveryNetworkConnector` and `LdapNetworkConnector` already had `getUri()` implementations and satisfy the new abstract contract without changes.
- Extend `JmxCreateNCTest` to assert the exposed URI values.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/NetworkConnectorView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/NetworkConnectorView.java
@@ -16,6 +16,8 @@
  */
 package org.apache.activemq.broker.jmx;
 
+import java.net.URISyntaxException;
+
 import org.apache.activemq.network.NetworkConnector;
 
 public class NetworkConnectorView implements NetworkConnectorViewMBean {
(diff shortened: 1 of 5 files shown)


mattrpav:
nit: Once v6.3.0 we should tighten up our back-port policies to not back-port new fields.

jbonofre:
@mattrpav it's not breaking change as it's new fields. I'm fine to close it, but adding fields or features that don't impact (breaking change) is reasonable to me. That said, as 6.3.0 is coming 😄 let me close this backport one.

mattrpav:
@jbonofre no need to close this, esp since 6.3.x has taken a while to get out. This is minor, but I think we need to clarify for going forward. Adding a metric to a network connector isn't a big deal, but a config flag to queue or kahadb could be. Changing bin/setenv OPTS variables, causes admins to have to re-work packaging bundles b/w point releases, etc. 

So I think 'adding a field' can be too reductive and over arching.

Keep this one for 6.2.x, I think we need clarify and possibly add a SEMVER.md for clarity all around

jbonofre:
@mattrpav agree. This PR is similar to the one you did (#2239 ). So, let revisit once 6.3.0 is out.

PR #2209, opened 2026-07-10
Minor updates to text message sync

cshannon:
This is a small follow on to #2188 to simply some of the sync by moving to the method level and also adding sync in a spot or 2 that got missed.

This commit also updates the race condition test to add all the other message types to show that they do not suffer from the same race condition with null bodies either.

As noted in #2188, only text message needs sync because the other message types do not null out the marshalled content when unmarshalling. Properties are also ok because unmarshalling properties do not clear the marshalled properties either (unless told to at specific spots in the broker when reduceMemoryFootPrint is true)


jbonofre (approved):
LGTM.

The test could be simplified imho, but that's allright.

Issue #1734, opened 2026-03-05
Migrate to Jetty 12


mattrpav:

grigoni:
Hi @mattrpav will this target jetty 12.0 or 12.1?

PR #2217, opened 2026-07-11
[#1734] Jetty 12.1 refactor

mattrpav:
Jetty 12.1 update #1734 

1. Web Socket updates
2. Jetty BOM alignment
3. New JettyServerBean that loads jetty-dist xml vs Spring bean wiring
4. Updated unit tests
5. Demos migrated to new Jetty config
6. Web Console loads
7. Recent 6.2.x security constraints migrated

Note: This is still currently aligned to EE9, and an EE11 conversion will follow-on this change.

Merge plan:
1. Squash commits (keep @grgrzybek's commits)

ref: AMQ-9359 (legacy JIRA issue)


cshannon:
@mattrpav - Wouldn't it make sense to go ahead and just bump to EE11 in this PR? it seems pretty straight forward and we need to do it anyways.

cshannon (changes requested):
At first glance this looks good, all the tests are passing. My main critique is I think we should just use ee11 in this PR, no real reason to do it separate as it's easy enough to do now as we discussed, because we are now bumping to Spring 7 in 6.3.0 due to 6.2.x being end of life and Spring 7 requiring Jakarta EE 11.

The other main thing is just ensuring that all the recent hardening stuff is still good. It looks ok to me but it would be good for @jbonofre to take a look as well as that was something he worked on. We don't want to introduce any regressions into the security/authorization layers.

Ultimately, this is a huge change and it's possible (probable) that something may end up not being quite right but I figure we can always do a patch release if someone in the community finds an issue.

mattrpav:
@cshannon the websocket layer gets significant overall in ee9 to ee11. I'm going to start on it, but keep the commits separate if it falls down or looks likes it going to take a few days. Worst case scenario we may consider a 6.3.0 EE9 and then 6.4.0 EE11. 

With the holiday/vacation schedules approaching, I think getting the PRs in for 6.3.0 this week would be good, so we can review and burn down to have v6.3.0 out at end of July, first of August.

edit: Scratch all that.. we got to get it in b/c Spring isn't releasing security patches for 6.2.x for consumption of OSS projects.

cshannon:
> @cshannon the websocket layer gets significant overall in ee9 to ee11. I'm going to start on it, but keep the commits separate if it falls down or looks likes it going to take a few days. Worst case scenario we may consider a 6.3.0 EE9 and then 6.4.0 EE11.
> 
> With the holiday/vacation schedules approaching, I think getting the PRs in for 6.3.0 this week would be good, so we can review and burn down to have v6.3.0 out at end of July, first of August.

@mattrpav - We have to upgrade to Spring 7, we have no choice., spring 6.2.x is now EOL. With the current AI models finding CVEs every 5 minutes we can't ship 6.3.0 with a version of spring that is already end of life. 

Spring 7.x officially bumps to Jakarta 11 so there is really nothing else we can do but upgrade. We could try and just stay at EE 9 or EE 10 and still bump spring and hope it doesn't break (it may not, depends on the apis used) but that is risky

cshannon:
I pushed up a PR to this branch with a jakarta ee 11 update that i tried out, i haven't run all the tests or anything. https://github.com/mattrpav/activemq/pull/5

Feel free to use it as a guide or not, etc.

cshannon (approved):
LGTM, Nice work!

Everything seems pretty good, tests are passing and the callback usage looks ok to me. I tested the console and things seems good (including security). 

Once we get this merged we can go ahead and make the switch to Spring 7.0.x. and fix anything broken with that.

PR #2219, opened 2026-07-13
Ensure composite temp dests are handled in AuthorizationBroker

cshannon:
For temporary destinations we need to iterate over the list of composite destinations and compute the union using the same strategy as DefaultAuthorizationMap.


mattrpav (approved):
LGTM

PR #2222, opened 2026-07-15
Upgrade to Spring 7.0.8

cshannon:
Spring 6.2.x is now EOL so this upgrades to Spring 7.0.x going forward. Spring 7 requires JakartaEE 11, which was upgraded in #1734

I looked this over and I think the only thing that needs to be done is updating the version and version range (which is used for OSGi). I ran through some of the spring tests and things seem to work without any issue but the full test suite needs to run.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -93,8 +93,8 @@
     <rome-version>2.1.0</rome-version>
     <shiro-version>2.2.1</shiro-version>
     <slf4j-version>2.0.18</slf4j-version>
-    <spring-version>6.2.19</spring-version>
-    <spring-version-range>[6,7)</spring-version-range>
+    <spring-version>7.0.8</spring-version>
+    <spring-version-range>[7,8)</spring-version-range>
     <taglibs-version>1.2.5</taglibs-version>
     <xpp3-version>1.1.4c</xpp3-version>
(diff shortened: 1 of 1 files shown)


cshannon:
@mattrpav and @jbonofre - It would be good if you both can look over the OSGi stuff to make sure it's ok with this change (plus all the changes in #1734). I think the version range for spring is enough for OSGi but i don't use it so i'm not the best person to test it out.

jbonofre:
@cshannon we would need the spring features in Karaf. In the meantime, I will implement a workaround. 

By the way, we have a dependabot PR about this update already (I commented on it while ago).

cshannon:
> @cshannon we would need the spring features in Karaf. In the meantime, I will implement a workaround. 
> 
> By the way, we have a dependabot PR about this update already (I commented on it while ago). 

Yes I am aware of the dependabot, I commented on the same PR. I opened this PR because I'm assuming there is more work to do here than a simple simple version change with the major version update

jbonofre:
@cshannon no worries, I will add commits in this one.

cshannon:
I realized we could just push to the dependabot branch if we need to make more changes, which I wasn't think about. I can close this and we can use that one instead which his #2102

Issue #1715, opened 2026-02-25
Implement Shared subscriptions


jeanouii:
`createSharedConsumer` and `createSharedDurableConsumer` to be implemented to pass TCK 3.1

PR #2225, opened 2026-07-17
WIP: [#1715] Jakarta Messaging 3.1 Shared Topic Support (Preview)

mattrpav:
- [x] Separate SharedTopicBrokerService allows for isolated testing
- [x] Separate SharedTopicConnectionFactory allows for isolated testing
- [x] OpenWire v13 w/ Shared Topic Subscription info and deliveryDelay accessors
- [x] Ability to migrate non-shared to shared (spec limitation work-around)

Jakarta Messaging specification calls out the Shared Topics are essentially a separate region from non-Shared topics. This works in our favor, as we can leverage *Region separation and implement this along side current topic support and not have to overhaul existing topic and subscription behavior.

Note: The SharedTopicBrokerService and SharedTopicConnectionFactory are intended to be  _transitional_ classes. This approach would allow us to release the feature as experimental/preview (if we chose) and keep the code path out of the stable current broker code. Once the implementation feels right, the last step would be to merge SharedTopicBrokerService -> BrokerService, etc.

### DIFF
--- a/activemq-broker/pom.xml
+++ b/activemq-broker/pom.xml
@@ -76,6 +76,11 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
       <groupId>org.apache.logging.log4j</groupId>
       <artifactId>log4j-core</artifactId>
       <scope>test</scope>
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/SharedTopicBrokerService.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/SharedTopicBrokerService.java
@@ -1,0 +1,167 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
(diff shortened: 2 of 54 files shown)


cshannon:
@mattrpav - Thanks for getting this started, I'll just make a couple points here.

This is going to be a lot of work so this isn't really suited for a single PR. The work here is going to need to be divided up and there will be some back and forth as we go. What I would recommend is the strategy we did in Apache Accumulo when we first started working on a major refactor of the project to make it more "elastic", and that is create a long running feature branch that we create PRs against.

One reason for the separate branch is we keep main clean until we are happy with the work. This is such a big feature that we might have to redo things and refactor a few times until we are happy with it. It's not possible to do it all in one big commit but it also doesn't make sense to merge work that is half done. During the initial work things are going to break, it's just a big refactor and that is fine becuase it will be isolated outside of main.

The other reason is we can avoid the extra work for the transition stuff. In fact, your PR directly illustrates this because you had to create classes like SharedTopicBrokerService which are throw away and I'd rather avoid. If we keep it separate we don't need to do that.  We can just update BrokerService directly as there is no need to worry about isolated testing because the feature branch is doing the isolation for us.

After creating the branch we need to periodically merge the main branch into it (note I said merge, we don't want to rebase) to keep things up to date. Once we thing it's good we can create a PR and merge the feature branch back into main and preserve the commit history.

So the basic flow would be:

1. Create a feature branch for shared sub support in the activemq repo
2. periodically merge main into it to keep it up to date and resolve conflicts
3. create PRs against the feature branch
4. Merge the feature branch back into main when we are done

cshannon:
I moved the above comment into a dev list discussion [here](https://lists.apache.org/thread/j1w3ytf11po6fxo5opns1tb4cgm2n3wl) which is a better spot for it I think

mattrpav:
Replaced by [#2229] targeted to activemq-wip-jakarta-3.1 branch

Issue #2230, opened 2026-07-18
KahaDB needs dedicatedIterator with dedicatedCursor


mattrpav:
When using a dedicated cursor with KahaDB and the current iterator that is shared with the queue cursor, messages may temporarily be skipped over or duplicate from store may occur.

This change adds a dedicated iterator to optionally not track last key changes.

PR #2231, opened 2026-07-18
[#2230] KahaDB dedicated cursor iterator

mattrpav:
Validation steps:

1. Checkout this branch 
2. Checkout first commit and build
3. Run the unit test and observe 2 test failures against 2 control tests
4. Checkout second commit and build
5. Run the unit test and observe 4 test pass

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
@@ -3848,12 +3848,26 @@ public abstract class MessageDatabase extends ServiceSupport implements BrokerSe
             }
         }
 
+        /**
+         * Iterate and record the last visited sequence keys
+         * (lastDefaultKey/lastHighKey/lastLowKey) so that a subsequent
+         * {@link #stoppedIterating()} advances the destination cursor past the
+         * visited messages.
+         */
         Iterator<Entry<Long, MessageKeys>> iterator(Transaction tx) throws IOException{
             return new MessageOrderIterator(tx,cursor,this);
         }
 
         Iterator<Entry<Long, MessageKeys>> iterator(Transaction tx, MessageOrderCursor m) throws IOException{
             return new MessageOrderIterator(tx,m,this);
+        }
+
+        Iterator<Entry<Long, MessageKeys>> iteratorIsolated(Transaction tx) throws IOException{
+            return iteratorIsolated(tx, new MessageOrderCursor());
+        }
+
+        Iterator<Entry<Long, MessageKeys>> iteratorIsolated(Transaction tx, MessageOrderCursor m) throws IOException{
+            return new IsolatedMessageOrderIterator(tx, m, this);
         }
 
         public byte lastGetPriority() {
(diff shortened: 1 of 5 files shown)


cshannon:
I only glanced at this and haven't reviewed it but it sounds like the name should probably be `IsolatedIterator` vs `DedicatedIterator` as a better description of what's going on

mattrpav:
> `IsolatedIterator`

'DedicatedItertor' was following the taxonomy of 'DedicatedCursor'. Are you suggesting a change to only the Iterator now, or go ahead and bite off changing both?

cshannon:
> > `IsolatedIterator`
> 
> 'DedicatedItertor' was following the taxonomy of 'DedicatedCursor'. Are you suggesting a change to only the Iterator now, or go ahead and bite off changing both?

I forgot about the DedicatedCursor, I would keep both the same so it's clear and if there is no breaking change you could rename both. Isolated i think better describes what's really going on, but if it's going to cause problems we can just keep it as Dedicated

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
```diff
@@ -3848,12 +3848,34 @@ MessageKeys put(Transaction tx, int priority, Long key, MessageKeys value) throw
...
+         * visited messages.
+         */
         Iterator<Entry<Long, MessageKeys>> iterator(Transaction tx) throws IOException{
-            return new MessageOrderIterator(tx,cursor,this);
+            return new MessageOrderIterator(tx,cursor,this,true);
         }
 
         Iterator<Entry<Long, MessageKeys>> iterator(Transaction tx, MessageOrderCursor m) throws IOException{
-            return new MessageOrderIterator(tx,m,this);
+            return new MessageOrderIterator(tx,m,this,true);
```
cshannon: I am trying to remember the original PR, but I feel like we should only need 2 methods and not 4.

Shouldn't we only ever need to pass in a new MessageOrderCursor when we want to be isolated? For some reason the recovery() method was updated to use new MessageOrderCursor() but i don't see why that was [changed](https://github.com/apache/activemq/pull/1377/changes#diff-804376ddbeaaa7b31ef291c209f364fde5f0b2ccbe4a20c137bebc0c4cf31990R689) in the original PR. That method already resets it [here](https://github.com/apache/activemq/pull/1377/changes#diff-804376ddbeaaa7b31ef291c209f364fde5f0b2ccbe4a20c137bebc0c4cf31990R687)

I would think we would only two variations of the iterator...one that tracks state (including the last used keys) and uses the MessageOrderCursor that is part of the definition and one that is completely isolated but maybe I am missing something.

cshannon (changes requested):

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
cshannon: Or maybe we update the recovery() method to not track the last keys either. Either way, it seems like we should either be isolated or not and the 4 methods includes a weird hybrid option of partial isolation

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
```diff
@@ -3893,8 +3915,12 @@ class MessageOrderIterator implements Iterator<Entry<Long, MessageKeys>>{
             final Iterator<Entry<Long, MessageKeys>>highIterator;
             final Iterator<Entry<Long, MessageKeys>>defaultIterator;
             final Iterator<Entry<Long, MessageKeys>>lowIterator;
+            // when false, next() leaves the shared lastDefaultKey/lastHighKey/lastLowKey
+            // bookmarks untouched so this iteration cannot affect the destination cursor
+            final boolean trackLastKeys;
 
-            MessageOrderIterator(Transaction tx, MessageOrderCursor m, MessageOrderIndex messageOrderIndex) throws IOException {
+            MessageOrderIterator(Transaction tx, MessageOrderCursor m, MessageOrderIndex messageOrderIndex, boolean trackLastKeys) throws IOException {
```
cshannon: Instead of using a single MessageOrderIterator with lots of args I am wondering if it makes sense to just have two impls. 

We could rename MessageOrderIterator to IsolatedMessageOrderIterator that does no tracking (uses a new MessageOrderIndex() and doesn't track keys), and then extend that class with MessageOrderIterator that tracks keys and uses the shared MessageOrderCursor.

This is just an idea I had, I'm not sure if it would make it better or worse without trying it compared to constructor orgs and a single impl

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
cshannon: I took a look closer and there are 2 spots that should be be using the isolated cursor, the [recover()](https://github.com/apache/activemq/blob/dce599f02be91890cdb7c26754d5f7221436fc8b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java#L652) method and also the [recoverExpired() ](https://github.com/apache/activemq/blob/main/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java#L1277) methods. Also, both methods have a reset call that can go away, for example: https://github.com/apache/activemq/blob/dce599f02be91890cdb7c26754d5f7221436fc8b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java#L1283

The reason it doesn't break is those methods are used exclusively for Topics and it doesn't interfere with the subscription recovery as each call to recoverNext() for a sub resets the cursor to the correct offset anyways.

Queues and durable subscriptions exclusively use the recoverNext() methods to keep track of state.

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
mattrpav: Latest commit 68eea73c71e581ad1c3804a3ed3ba1726276ce57 applies these changes

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/MessageDatabase.java
mattrpav: Change applied here https://github.com/apache/activemq/pull/2231/changes/77de343b09da1cda34a0a2ad6c522e302020c103

cshannon (approved):
LTGM assuming tests are all good

PR #2233, opened 2026-07-18
Add support for PageFile free page truncation

### LINKED ISSUE #2232: KahaDB index compaction
KahaDB contains two components, he journal and the index. The journal stores all of the data (mesages, acks, metadata, etc) as commands in the journal and the index is used to keep track of where certain things are for quick look up (such as being able to iterate over messages by order or by id). The goal of this issue is to address compacting the index (stored as page files) when large. Journal compaction is a separate issue that will be addressed separately. Inded compaction will be done in two phases, truncation and defrag.

### Index usage in KahaDB

The KahaDB index is stored using a PageFile, which is just a contiguous block of pages in a single file.  The page file is essentially a big linked list on disk, and writes will be written to as many pages as needed and flushed to disk. Pages can also be cached in memory for fast access.

When updates are written to the index and there isn't enough space, new pages are allocated as necessary which grows the file. When deletes happen pages are just marked as free and reclaimed into a list of free pages. Writing updates to the index first looks to re-use pages that are marked as free and will only allocate new space if needed. Under normal use cases this leads to a pretty stable PageFile size as the amount of pages necessary get allocated. 

### The need for compaction

Sometimes the PageFile can grow quite large. The most common scenario for this is a large backlog on a destination. A large backlog will cause a lot of extra writes to the index to track all the messages which causes the page file to grow quite large. Once the backlog is gone the pages will be marked as free and reclaimed to be re-used, but the page file itself will never shrink. This can lead to a huge PageFile with a lot of empty space that just isn't necessary. 

### Compaction via truncation (#2233) 

A simple way to help solve this problem is using a truncation compaction strategy. Truncation helps in this scenario by simply truncating the file and removing the excess free pages. This works well because all we need to do is remove the free pages from the tracking and then set the length of the file to the new size which is nearly instant time, so there is no noticeable performance impact.

While this strategy is very fast and easy to implement, the major and obvious limitation to this strategy is it only works if the free pages are at the end of the file. If the free pages are in the middle of the file his won't work so we can't shrink the file. This could happen, for example, if a new destination was created while there was a big backlog and it was written to the index. This means all the eventual free pages that would be reclaimed after the backlog is gone would be in the middle of the file so we could not truncate. 

### Compaction via defragmentation (Future plans)

The goal here is to address the shortcomings of truncation and there are a few ideas to explore. One way to reclaim space is to just re-write the entire page file and drop free pages but this would be pretty slow. Another idea is to implement defragmentation to move pages and reduce the empty space in the middle of the file so that truncation could work and there are a few ideas to try:

1. One option to explore is we re-write/move pages at the end of the file to early free pages. Eventually we will end up with enough free pages at the end of the file and we can truncate again. This is more complex because the page file forms lots of separate linked lists within the page file. We could convert the linked lists to a doubly linked lists so that when we move pages we can update the metadata of the page that points to the page so we don't break the list. This also means it would not be backwards compatible (you'd need to start with a fresh page file) because existing page files don't contain markers for the previous page in the list. 
2. Another option would be to still move pages to earlier in the file, but add another index that's a virtual page file and a look up table that would likely be stored in a second small page file. It would just be a map that we could use to find the real location of a page based off the stored pageId if it was moved. 
3. A third option worth exploring is to just rewrite stored destination metadata to earlier in the file and over time truncation would naturally work as backlogs are gone. The main cause of big files (if truncation is enabled) would be data that is permanent like destination data so moving it earlier int he page file could allow cleanup. Stored destination metadata just contains the "first" pages of each of its indexes in the linked list so you don't need to store previous pages (so no doubly linked list). One flaw is that this may not necessarily work as intended though if a destination has a lot of durable subscriptions and they get added later after a big backlog as that would add pages to the index that were not moved.

It's TBD which would work best.

cshannon:
This commit adds support for index compaction using a truncation strategy as described in #2232. See that issue for more information.

Compaction works by reclaiming space by truncating free pages at the end of the file. A large amount of free pages can be allocated during certain use cases (such as a large message backlogs) and this strategy helps by removing the excess space when possible.

The amount of space reclaimed is configurable using a min/max free page ratio. The min free page ratio defines the amount of free pages to leave and the max free page ratio defines the max free pages to allow before attemping to reclaim space.

The compaction check will be performed during the normal checkpoint cycle.

It's also worth noting this doesn't introduce any extra risk of data loss. By default the recovery file is on, and will be able to recover free pages if a failure happens during truncation. In the worst case scenario where the file is completely corrupted (which is the same risk as any other index write, there isn't anything special about truncation), the index corruption is detected and the file can be rebuilt form the journal.

### DIFF
--- a/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
+++ b/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
@@ -35,6 +35,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.TreeMap;
 import java.util.concurrent.CountDownLatch;
@@ -137,8 +138,9 @@ public class PageFile {
     private final AtomicLong nextFreePageId = new AtomicLong();
     private SequenceSet freeList = new SequenceSet();
 
-    private AtomicReference<SequenceSet> recoveredFreeList = new AtomicReference<SequenceSet>();
-    private AtomicReference<SequenceSet> trackingFreeDuringRecovery = new AtomicReference<SequenceSet>();
+    // package visibility for testing
+    final AtomicReference<SequenceSet> recoveredFreeList = new AtomicReference<>();
+    private final AtomicReference<SequenceSet> trackingFreeDuringRecovery = new AtomicReference<>();
 
     private final AtomicLong nextTxid = new AtomicLong();
 
@@ -149,6 +151,33 @@ public class PageFile {
 
     private boolean useLFRUEviction = false;
     private float LFUEvictionFactor = 0.2f;
+
+    // Compaction config
+    private PageFileCompactionStrategy compactionStrategy = PageFileCompactionStrategy.NEVER;
+
+    // The ratio of the minimum amount of free pages to keep
+    // The default will keep a minimum of 10% of free pages, relative to the
+    // current size of the file when compaction is started.
+    // For example, if the file has 1000 pages and 500 are free at the end of the file,
+    // 400 would be removed and 100 would be kept (100 is 10% of 1000). This would leave
+    // a file with 500 pages used + 100 free
+    private float minFreePageCompactionRatio = .1F;
+
+    // The ratio of the maximum amount of free pages to allow before triggering compaction.
+    // The default will trigger a compaction attempt if the percentage of free pages
+    // hits 30% of the total file size.
+    // For example, if the file has 1000 pages and 350 are free at the end of the file, this
+    // is greater than 30% so we'd truncate the file. If minFreePageCompactionRatio is kept
+    // as the default of 10%, 250 pages would be removed leaving 100.
+    private float maxFreePageCompactionRatio = .3F;
+
+    public enum PageFileCompactionStrategy {
+        TRUNCATION, NEVER;
+
+        public boolean isNever() {
+            return this == NEVER;
+        }
+    }
 
     /**
      * Use to keep track of updated pages which have not yet been committed.
@@ -372,6 +401,9 @@ public class PageFile {
      */
     public void load() throws IOException, IllegalStateException {
         if (loaded.compareAndSet(false, true)) {
+            if (maxFreePageCompactionRatio < minFreePageCompactionRatio) {
+                throw new IllegalStateException("minFreePageCompactionRatio is greater than maxFreePageCompactionRatio");
+            }
 
             if (enablePageCaching) {
                 if (isUseLFRUEviction()) {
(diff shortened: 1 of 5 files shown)


cshannon:
I added tests here for the PageFile to cover different scenarios, but the test coverage could still probably be expanded. We could add more tests using KahaDB (there's only one now). I am targeting this for 6.4.0 for now, so there is some time as this won't get merged until 6.3.0 is done.

mattrpav:
@cshannon this is a super awesome feature! First glance, I think it should include some metrics for tracking time, pages and/or file size compacted. 

I’m happy to help add those

cshannon:
> @cshannon this is a super awesome feature! First glance, I think it should include some metrics for tracking time, pages and/or file size compacted.
> 
> I’m happy to help add those

Metrics for pages/size are useful but time would not be for truncation. Time would be useful for when defrag is implemented but for truncation it's nearly instant time so there's nothing to really measure. Also, something to discuss is if we enable this by default. For something like defragmentation we could leave it off by default, but for a simple truncation like this PR is doing I think having it on by default could make sense. We would need to decide on default parameter thresholds though. For now I picked .1 for `minFreePageCompactionRatio` (leave 10% free pages relative to the current size after compaction) and .3 for `maxFreePageCompactionRatio` (don't start trying to compact until there's at least 30% pages that are free relative to the current size)

cshannon:
Also, this can be configured using spring as long as the value matches the exact enum value, including case:
```
<persistenceAdapter>
    <kahaDB directory="target/data/kahadb" indexCompactionStrategy="NEVER"/>
</persistenceAdapter>
```

If we wanted to make it case insensitive we could make the setIndexCompactionStrategy() setter take a String and do the conversion from that manually (we do this for journalDiskSyncStrategy)

mattrpav:
Specifics to requested changes (I'm happy to add on a feature branch, let me know your preference).

Metrics:
1. Add a min/max/avg page file size metrics (measures effectiveness of enabling trunaciton)
2. Add currentPageFileSize
3. Add trunactionCount <-- count how many times truncation has occurred

Config/Status:
1. truncationSupported: true|false <-- does the fs support truncation?
2. truncationEnabled: true| false <-- did admin set truncation to be on?
3. truncationStatus: active | disabled <-- the resolved value of (supported && enabled)

cshannon:
I think for the metrics we could do a follow on PR for it. One thing is to keep in mind that we might be expanding (ie defrag) with other compaction options so the naming should be flexible enough. So I agree the naming should probably involve "truncation" because defrag will be different. Defrag would have have its own metrics (like how many pages were moved and how often and how long it took). Ideally we'd run a defrag to move pages, and then future GC cycles could just truncate once the pages got moved.

cshannon:
Here is an example output of this in action. It truncated the file over two GC checks (this has to do with the math and config of max/min free page ratio). According to the logs the first truncation was 5 milliseconds and the second was < 1 milliseconds. There is a 5 second gap because by default KahaDB does a checkpoint every 5 seconds.

```
026-08-21 17:08:19,219 DEBUG (PageFile) - Beginning PageFile compaction check.
2026-08-21 17:08:19,219 DEBUG (PageFile) - Number of free pages to be deleted: 35917
2026-08-21 17:08:19,219 DEBUG (PageFile) - Disk size of end of file free pages: 147116032
2026-08-21 17:08:19,220 DEBUG (PageFile) - Truncating page file to length: 16961536
2026-08-21 17:08:19,224 DEBUG (PageFile) - Page file was compacted, new length: 16961536, old length:164077568
2026-08-21 17:08:19,224 DEBUG (PageFile) - New page count: 4140, free page count: 4006
2026-08-21 17:08:24,229 DEBUG (PageFile) - Beginning PageFile compaction check.
2026-08-21 17:08:24,229 DEBUG (PageFile) - Number of free pages to be deleted: 2555
2026-08-21 17:08:24,229 DEBUG (PageFile) - Disk size of end of file free pages: 10465280
2026-08-21 17:08:24,229 DEBUG (PageFile) - Truncating page file to length: 6496256
2026-08-21 17:08:24,229 DEBUG (PageFile) - Page file was compacted, new length: 6496256, old length:16961536
2026-08-21 17:08:24,229 DEBUG (PageFile) - New page count: 1585, free page count: 1451
```

Here is the before and after with how much data was saved for the index file. This was an extreme example because I purposely created a large backlog without any new metadata (like destinations) added to prevent truncation just to illustrate the potential:

**KahaDB dir before truncation:**
```
-rw-rw----. 1 user user  16M Aug 21 17:08 db-409.log
-rw-rw----. 1 user user 157M Aug 21 17:08 db.data
-rw-rw----. 1 user user 3.2M Aug 21 17:08 db.redo
-rw-rw----. 1 user user    8 Aug 21 17:08 lock
```

**KahaDB dir after truncation:**
```
-rw-rw----. 1 user user  16M Aug 21 17:08 db-409.log
-rw-rw----. 1 user user 6.2M Aug 21 17:08 db.data
-rw-rw----. 1 user user 3.2M Aug 21 17:08 db.redo
-rw-rw----. 1 user user    8 Aug 21 17:08 lock
```

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
```diff
@@ -613,6 +645,150 @@ public void flush() throws IOException {
...
+            // Now that metadata is all updated, we can truncate the file
+            // This should be the last step to make sure there is no issue
+            // updating any of the metadata/tracking before truncation
+            writeFile.getRaf().setLength(newDiskSize);
+            if (enableDiskSyncs) {
+                writeFile.sync();
+            }
+
+            LOG.debug("Page file was compacted, new length: {}, old length:{}", newDiskSize, diskSize);
+            LOG.debug("New page count: {}, free page count: {}", getPageCount(), getFreePageCount());
```
mattrpav, line 788: Can the two debug log statement be combined into one?

activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/disk/page/PageFileCompactionTest.java
```diff
@@ -0,0 +1,385 @@
...
+    // restored back to before compaction - the 20 pages were truncated
+    // and won't come back and are lost without a recovery file
+    assertEquals(pf2.getFile().length(), pf2.getDiskSize());
+    assertEquals(pf2.toOffset(80), pf2.getDiskSize());
+    assertEquals(0, pf2.getFreePageCount());
+    assertEquals(80, pf2.getPageCount());
+
+
+
+    // TODO we also want to test with recovery file still enabled, but
```
mattrpav, line 202: Address the TODO now or in a follow on?

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
```diff
@@ -613,6 +645,150 @@ public void flush() throws IOException {
...
+        // we are updating the file and truncating
+        synchronized (writes) {
+            // Generally this should be empty but need to verify
+            if (!writes.isEmpty()) {
+                LOG.warn("Skipping compaction, writes are in flight.");
+                return;
+            }
+
+            LOG.debug("Number of free pages to be deleted: {}", freePagesToDelete.range());
+            LOG.debug("Disk size of end of file free pages: {}", pageSize * freePagesToDelete.range());
```
mattrpav, line 762: Thoughts on consolidating to one debug log line here?

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
```diff
@@ -613,6 +645,150 @@ public void flush() throws IOException {
...
+            var last = lastFreeSeq.getLast();
+            var updatedFirst = (last - maxPagesToTruncate) + 1;
+            freePagesToDelete = new Sequence(updatedFirst, last);
+        } else {
+            freePagesToDelete = lastFreeSeq;
+        }
+
+        // sync on writes to block the async thread from trying to write while
+        // we are updating the file and truncating
+        synchronized (writes) {
```
mattrpav, line 754: What is the interplay here b/w locking on the writes and the compact() already running within the index lock? 

If it is in the indexLock, then do the writes need to be synchronized here? If so, thoughts on it moving up sooner since the threshold checks could rapidly change (ie async producer)?

mattrpav (commented):
Review comments in-line

activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/disk/page/PageFileCompactionTest.java
cshannon: This is an old TODO that should have been removed, there's a test for recovery and no recovery file now

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
cshannon: There is a separate thread that can handle writing async (optionally turned on) and that is independent of the index lock as it's internal to the PageFile. The writes are queued up in a map so this needs to lock to block the other thread from trying to read that map. The checkpoint calls flush to flush any pending writes first, so there is also an extra sanity check to verify there's nothing pending to write.

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
```diff
@@ -613,6 +645,150 @@ public void flush() throws IOException {
...
+        long totalPageCount = getPageCount();
+
+        // Percentage of pages that are free vs in use
+        double freePageRatio = Math.round((double)freePageCount / totalPageCount * 100d) / 100d;
+
+        // Only attempt to compact if we have reached the maximum ratio of free pages
+        // that is configured.
+        var tolerance = .001;
+        if (freePageRatio < maxFreePageCompactionRatio - tolerance) {
+            var formatted = Math.round((double)freePageCount / totalPageCount * 100d) / 100d;
```
mattrpav: This calculation is duplicate of 'freePageRatio' above.

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
```diff
@@ -613,6 +645,150 @@ public void flush() throws IOException {
...
+        if (toOffset(lastFreeSeq.getLast()) + pageSize != diskSize) {
+            LOG.debug("Unable to compact, last free page block is not at the end of the file");
+            return;
+        }
+
+        long minFreePages = Math.round(totalPageCount * minFreePageCompactionRatio);
+
+        // we must keep a minimum number of free pages so find the point
+        // where we can truncate without removing too many pages
+        long maxPagesToTruncate = freePageCount - minFreePages;
```
mattrpav, line 736: minFreePages is rounded -- prolly wouldn't hurt to add a guard here to return when maxPagesToTruncate <=0

activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/disk/page/PageFile.java
```diff
@@ -613,6 +645,150 @@ public void flush() throws IOException {
...
+        long freePageCount = getFreePageCount();
+        long totalPageCount = getPageCount();
+
+        // Percentage of pages that are free vs in use
+        double freePageRatio = Math.round((double)freePageCount / totalPageCount * 100d) / 100d;
+
+        // Only attempt to compact if we have reached the maximum ratio of free pages
+        // that is configured.
+        var tolerance = .001;
+        if (freePageRatio < maxFreePageCompactionRatio - tolerance) {
```
mattrpav: Feels like the rounding could result in fuzzy numbers. Thoughts on using something simplified:

```
if (freePageCount < totalPageCount * maxFreePageCompactionRatio)
```

cshannon: Your suggestion to simplify makes sense and is fine, but we can't really avoid the fuzzy math because we are trying to compare a long and a float/double so it makes sense to just truncate the decimal or round. It's just an estimate and at the end of the day it doesn't matter that much.

We could cast to a long but I think something like this makes sense:

`if (freePageCount < Math.round(totalPageCount * maxFreePageCompactionRatio)) `

It's an estimate and so I think rounding to the near long value and comparing makes a lot of sense.

mattrpav (approved):
LGTM

PR #2246, opened 2026-07-24
Fix reactor build for sources (used by release plugin)

jbonofre:


cshannon (approved):
I suppose we should start doing a dry run of the release plugin earlier so we catch issues like this before release day :smile:

Issue #2187, opened 2026-07-04
Modernize ActiveMQMessageAudit to be lock-free


mattrpav:
The current ActiveMQMessageAudit is implemented using synchronized blocks around an LRUCache implemented with a LinkedHashMap as an extension to the sync-free ActiveMQMessageAuditNoSync class. 

Modern Java language collection (ConcurrentHashMap) and lock-free programming approach can improve performance (especially when there are concurrent producers). Both classes can be replaced with a single, modern implementation.

A modernized MessageAudit:

- [ ] Provide exact same functional capabilities as current ActiveMQMessageAudit/ActiveMQMessageAuditNoSync
- [ ] Allow for shrinking/growing the audit size without completely clearing the list

PR #2252, opened 2026-07-24
WIP: [#2187] MessageAudit modernization and performance optimization

mattrpav:
- [x] AtomicBitArrayBin provides lock-free bin array for storing messageIds
 - [x] ConcurrentMessageAudit demonstrates a MessageAudit using JDK-only concurrent hashmap
 - [x] CaffeineMessageAudit demonstrates a MessageAudit using Caffeine cache library
 - [x] MessageAuditPerformanceTest compares results between the implementations
 - [x] activmeq-client/src/test/resources/benchmark-results.html provides results of a test run on MacBookPro M3


cshannon (commented):
CaffeineMessageAudit is not going to work without some tweaks as it isn't entirely thread safe. The map updates (adding/removing) from the cache would be ok but it doesn't protect another thread from removing or update after you get an object from the cache.

For example in your rollback method:

``` java
    var bab = cache.getIfPresent(pid.toString());
    // At this point another thread could remove/update the entry in the cache
    if (bab != null) {
        bab.setBit(id.getProducerSequenceId(), false);
    }
```

cshannon:
Also, from first glance the `AtomicBitArrayBin` class looks incredibly complex, did you use something to generate that? I have a lot of hesitation with introducing something that complicated vs simply locking without a real benefit. simple locking may be better under high contention anyways (that's where the benchmarking comes into play)

The main issue here is verifying it's actually correct and truly thread safe and won't introduce weird bugs

cshannon:
Have you looked into using an [async](https://github.com/ben-manes/caffeine/blob/master/caffeine/src/main/java/com/github/benmanes/caffeine/cache/AsyncCache.java) cache? With everything being done with computeIfPresent and under lock of the cache there could be a boost by using that as entries for different keys can be operated on independently. You can configure a thread pool and in the future Caffeine may support virtual threads (it doesn't yet). I haven't looked at your benchmark stuff yet to see if it could compare sync vs async effectively

cshannon:
So to be clear, if using the async cache we still need to wait on the future returned to complete but it might increase concurrency across different message ids (keys). We can always stick to the sync version for now as the computation for storing in the cache should be quick, I was just curious if there was any difference. The underlying cache delegates to ConcurrentHashMap which will block some operations if compute() is slow so the async version offloads to another thread. 

If there isn't much difference the standard sync cache is simpler.

Issue #2238, opened 2026-07-23
Flaky test cleanup following on v6.3.0 release


mattrpav:
In Progress (@mattrpav)

- [ ] (2x) AMQ4083Test.testExpiredMsgsBeforeNonExpiredWithTX:192 Inflight count should reach zero, currently: 7
- [ ] NetworkAdvancedStatisticsTest.testNetworkAdvancedStatisticsErrors:143->assertNetworkStats:406 Timed out waiting for connector exception counts local>=1 remote>=0
- [ ] NetworkBridgeReconnectOnHandshakeFailureTest.testBridgeReconnectsAfterRemoteBrokerRestart:263 Bridge should go down

Backlog fixes

- [ ] AbortSlowAckConsumer0Test.testSlowConsumerIsAbortedViaJmx:99->AbortSlowConsumer0Test.testSlowConsumerIsAbortedViaJmx:123 no slow consumers left expected:<0> but was:<1>
- [ ] VirtualConsumerDemandTest.testToTopicWithDurable:1358
- [ ] MQTTSubscriptionRecoveryTest.testDurableSubscriptionsAreRecovered:91->MQTTTestSupport.restartBroker:133 » IO Transport Connector could not be registered in JMX: java.io.IOException: Failed to bind to server socket: mqtt://0.0.0.0:46365?transport.subscriptionStrategy=mqtt-default-subscriptions due to: java.net.BindException: Address already in use
- [ ] RequestReplyToTopicViaThreeNetworkHopsTest.runWithTempTopicReplyTo:458 no errors
- [ ] InMemoryJobSchedulerTest.testAddCronAndByteSequence:92 » TestTimedOut test timed out after 60000 
- [ ] UsageBlockedDispatchTest>CombinationTestSupport.runBare:113->CombinationTestSupport.runBare:107->testFillMemToBlockConsumer:167 did not get a message
- [ ] CachedLDAPAuthorizationModuleLegacyTest>AbstractCachedLDAPAuthorizationMapLegacyTest.testSynchronousUpdate:123 expected null, but was:<[]>
- [ ] JMSClientNioPlusSslTest>JMSClientTest.testDurableConsumerSync:718->JMSClientTestSupport.createConnection:140->createConnection:40 » JmsConnectionRemotelyClosed org.apache.qpid.proton.engine.TransportException: AMQP header mismatch value 0, expecting 41. In state: HEADER0 [condition = amqp:connection:framing-error]
[...]

PR #2256, opened 2026-07-27
Fix AMQP Object message decompression

cshannon:
This fixes the AmqpMessageSupport utility to correctly decompress the body of Object Message's in all cases.

Previously the decompression could stop early and not decompress the entire stream of data. This was due to the loop incorrectly casting the read int from the inflater stream as a byte before comparing to -1 to look for end of stream.

This is incorrect because the read() method can return a value between -1 and 255. Java uses Two's Complement to represent signed ints, so if the returned value int is 255 and is cast to a byte it becomes -1. This meant that reading 255 would return -1 leading to the code to exit thinking end of stream has been reached.

This was discovered when I as was looking into the "flaky" test error for https://github.com/apache/activemq/issues/2238, `JMSInteroperabilityTest.testOpenWireToQpidObjectMessageWithOpenWireCompression:493 » MessageFormat Failed to read object: null`

Turns out, the test was not flaky but showed a real error. The test uses a random UUID, so by the test would intermittently fail depending on if the UUID decompression ran into a 255 int or not.

### DIFF
--- a/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/message/AmqpMessageSupport.java
+++ b/activemq-amqp/src/main/java/org/apache/activemq/transport/amqp/message/AmqpMessageSupport.java
@@ -256,12 +256,6 @@ public final class AmqpMessageSupport {
                      InputStream iis = MarshallingSupport.createInflaterInputStream(
                             message.getMaxInflatedDataSize(), is)) {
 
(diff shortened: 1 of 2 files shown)


tabish121 (commented):
LGTM

tabish121 (approved):

PR #2270, opened 2026-07-31
fix(jmx): avoid NPE in ConnectionView.getWireFormatInfo

### LINKED ISSUE #2203: NPE at ConnectionView.java > getWireFormatInfo()
target source:
https://raw.githubusercontent.com/apache/activemq/refs/tags/activemq-5.19.8/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectionView.java

issue:
At getWireFormatInfo() ,
If the "connection" is null, there would be some notice like this.

@Override
public String getWireFormatInfo() {
    if (connection instanceof TransportConnection) {
        WireFormatInfo info = ((TransportConnection) connection).getRemoteWireFormatInfo();  # << mod
        return (info != null) ? info.toString() : "WireFormatInfo not available"; # << mod
    }
    return "WireFormatInfo not available";
}

arimu1:
## Summary

`ConnectionView.getWireFormatInfo()` called `getRemoteWireFormatInfo().toString()` without a null check. `TransportConnection.wireFormatInfo` is only set after wire-format negotiation, so reading the JMX attribute earlier (or for transports that never set it) threw a `NullPointerException`.

This change returns the existing `"WireFormatInfo not available"` placeholder when the remote `WireFormatInfo` is null, matching the non-`TransportConnection` path.

Fixes #2203

## Changes

- `ConnectionView.getWireFormatInfo()`: null-safe remote `WireFormatInfo` handling
- `ConnectionViewTest`: regression coverage for present and null remote info

## Test plan

- [x] `mvn -pl activemq-broker test -Dtest=ConnectionViewTest` (JDK 17)
  - `testGetWireFormatInfoWhenPresent`
  - `testGetWireFormatInfoWhenRemoteInfoNull`

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectionView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectionView.java
@@ -23,6 +23,7 @@ import javax.management.ObjectName;
 
 import org.apache.activemq.broker.Connection;
 import org.apache.activemq.broker.TransportConnection;
+import org.apache.activemq.command.WireFormatInfo;
 import org.apache.activemq.util.IOExceptionSupport;
 
 public class ConnectionView implements ConnectionViewMBean {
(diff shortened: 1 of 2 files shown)


mattrpav (approved):
LGTM

cshannon:
This can be merged if CI is good

Issue #2271, opened 2026-07-31
Missing feature Delivery Delay (#1717) causes messaginghub/pooled-jms library to fail with ActiveMQ 6.3.0


graben:
ActiveMQ 6.3.0 specified itself as 3.1 compliant but it isn't, yet! Thats causing Exceptions at least with messaginghub/pooled-jms.

```
java.lang.UnsupportedOperationException: getDeliveryDelay() is not supported
	at org.apache.activemq.ActiveMQMessageProducerSupport.getDeliveryDelay(ActiveMQMessageProducerSupport.java:71) ~[activemq-client-6.3.0.jar:6.3.0]
	at org.messaginghub.pooled.jms.JmsPoolMessageProducer.<init>(JmsPoolMessageProducer.java:64) ~[pooled-jms-3.2.2.jar:?]
	at org.messaginghub.pooled.jms.pool.PooledSessionHolder.getOrCreateProducer(PooledSessionHolder.java:187) ~[pooled-jms-3.2.2.jar:?]
	at org.messaginghub.pooled.jms.JmsPoolSession.createProducer(JmsPoolSession.java:454) ~[pooled-jms-3.2.2.jar:?]
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104) ~[?:?]
	at java.base/java.lang.reflect.Method.invoke(Method.java:565) ~[?:?]
```

jbonofre:
@graben if you take a look on https://activemq.apache.org/components/classic/documentation/jms2 you can see that we clearly say that ActiveMQ 6.3.0 is not fully 3.1 compliant (still a work in progress).
The full 3.1 supports is planned for 6.5.0.

DeliveryDelay and receiveBody are already a work in progress.

graben:
Well, but the truth is in the code [ActiveMQConnectionMetaData.java](https://github.com/apache/activemq/blob/main/activemq-client/src/main/java/org/apache/activemq/ActiveMQConnectionMetaData.java). This should not have been updated until really 3.1 compliant IMHO.

mattrpav:
@graben the path to full Jakarta 3.1 is within view. The 6.3.0 release was critical as it unblocks a large dependency set (Jetty and Spring).

[...]

Issue #1713, opened 2026-02-25
Setup Jakarta Messaging 3.1 TCK


jeanouii:
We need to setup Jakarta 3.1 TCK on Apache ActiveMQ
It needs to be activated only on demand (specific profile because it takes a bit of time to run). We'll probably create a dedicated run on Github to run those.

Quoting JB on the email thread discussion.

> It is great to have this for two main reasons:
>
> 1. It provides a clear assessment of our current Jakarta Messaging 3.1 support and identifies the remaining gaps.
> 2. It gives us the opportunity to be listed as a compatible implementation on the Jakarta website.

jeanouii:
## Known Failures

The majority of failures are caused by unimplemented JMS 2.0 features in ActiveMQ:

| Category | Root Cause | Failures |
|---|---|---|
| Async send | `CompletionListener` APIs throw `UnsupportedOperationException` | 36 |
| Shared subscriptions | `createSharedConsumer` / `createSharedDurableConsumer` not implemented | 12 |
| `receiveBody()` | `JMSConsumer.receiveBody(Class<T>)` not implemented | 12 |
| Delivery delay | `setDeliveryDelay` / `getDeliveryDelay` not implemented | 10 |
| JMS version metadata | `ConnectionMetaData` reports version 1.1 instead of 3.1 | 4 |
| Exception types | Wrong exception class thrown in several edge cases | ~15 |
| Other | Temp destination lifecycle, property handling, security exceptions | ~12 |

PR #2272, opened 2026-07-31
[#2271] Update Producer getDeliveryDelay to always return zero

mattrpav:
### Change

This change modifies the in-progress behavior of deliveryDelay support by allowing all zero (0) values to continue to be processed. Zero (0) indicates no deliveryDelay, so the fact that this is hard-coded is no problem, as no value is going across the wire.

The UnsupportedOperationException is now thrown only when a non-zero value is specified on the setDeliveryDelay.

### Driving factor

The messagehub jms pool queries the value, even when unused to retain state between instances of being borrowed and is gated around the ConnectionMetaData major and minor version. Since we changed those values, accessing of setDeliveryDelay is ungated and all Spring applications are unable to send JMS messages using ActiveMQ v6.3.0.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducerSupport.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducerSupport.java
@@ -68,7 +68,7 @@ public abstract class ActiveMQMessageProducerSupport implements MessageProducer,
      */
     @Override
     public long getDeliveryDelay() throws JMSException {
-        throw new UnsupportedOperationException("getDeliveryDelay() is not supported");
+        return 0L;
     }
     
     /**
--- a/activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
+++ b/activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
@@ -251,7 +251,7 @@ public class ActiveMQProducer implements JMSProducer {
 
     @Override
     public long getDeliveryDelay() {
-        throw new UnsupportedOperationException("getDeliveryDelay() is not supported");
+        return 0L;
     }
 
     @Override
--- a/activemq-jms-pool/src/main/java/org/apache/activemq/jms/pool/PooledProducer.java
+++ b/activemq-jms-pool/src/main/java/org/apache/activemq/jms/pool/PooledProducer.java
@@ -156,7 +156,7 @@ public class PooledProducer implements MessageProducer {
      */
     @Override
     public long getDeliveryDelay() throws JMSException {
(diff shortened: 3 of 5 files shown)


activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
```diff
@@ -246,12 +246,16 @@ public long getTimeToLive() {
...
+            throw new UnsupportedOperationException("setDeliveryDelay(long) is not supported");
+        }
+
+        return this;
     }
 
     @Override
     public long getDeliveryDelay() {
-        throw new UnsupportedOperationException("getDeliveryDelay() is not supported");
+	return 0L;
```
cshannon: indentation looks off

cshannon (commented):
Do we also need to change the setter? If it's only querying is the getter enough?

cshannon:
Also, why don't we just roll back the version that is reporter? This fix is a mostly a workaround that shouldn't be necessary, if we just roll back the major version to 1 like it should be then that also fixes it.

cshannon:
@jeanouii - What do you think of the fix here vs rolling back the version to 1 until it's finished? You had made the change originally to ActiveMQConnectionMetaData.java in #1713 so I'm not sure if we can roll it back easily.

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducerSupport.java
```diff
@@ -56,7 +56,9 @@ public ActiveMQMessageProducerSupport(ActiveMQSession session) {
      */
     @Override
     public void setDeliveryDelay(long deliveryDelay) throws JMSException {
-        throw new UnsupportedOperationException("setDeliveryDelay() is not supported");
+        if (deliveryDelay != 0) {
```
cshannon: My preference would be to leave the setter alone. I don't think we should allow using it as it's not supported, the getter can just return 0 for now, meaning there's just no delay.

cshannon (changes requested):

activemq-client/src/main/java/org/apache/activemq/ActiveMQMessageProducerSupport.java
mattrpav: PR updated

activemq-client/src/main/java/org/apache/activemq/ActiveMQProducer.java
mattrpav: Fixed

mattrpav:
Changes applied

cshannon (approved):

Issue #2275, opened 2026-08-02
Refactor Usage to address contention


mattrpav:
While performing benchmarks against 6.3.0 with virtual threads, I observed an unexpected degradation in parallelized throughput. 

I honed the test down to everything stripped back (no audit, no expiry, async send, etc) and found that since memoryUsage is shared across all destinations the locking here accounted for the majority of wait time. Moving to a lock-free data structure for this metric removes the bottleneck. 

To see the effect, run the benchmark from the main and then from the patched commit. 

Measured on an 11-core Apple M3 Pro, JDK 25, direct dispatch, topic-per-producer:
    
      threads   stock (main)              [#2275] AtomicLong
         1         5.19M msgs/sec           5.76M
         4         1.49M                    6.21M   (+317%)
         8         1.26M                    5.66M   (+348%)
        22         1.07M                    6.05M   (+465%)

PR #2276, opened 2026-08-02
[#2275] Refactor Usage to address contention

mattrpav:
Reviewer notes:

1. Checkout the first commit to run the test and see the performance
2. Checkout the second commit to see LongAdder performance
3. Checkout the third commit to see AtomicLong performance

I'm leaning towards AtomicLong as its more commonly understood and would make the change less brittle to maintain going forward.

edit: Design note: Consider calculating max limit and doing comparisons vs calculations

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/usage/StoreUsage.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/usage/StoreUsage.java
@@ -68,18 +68,6 @@ public class StoreUsage extends PercentLimitUsage<StoreUsage> {
     }
 
     @Override
-    public int getPercentUsage() {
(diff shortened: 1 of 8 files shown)


cshannon (commented):
Thanks for getting this started, I agree AtomicLong is probably the way to go for now. 

It would be nice if we can get rid of the lock and calculate usage. Maybe we could compute the actual value (and not use percentage) on first use or startup (ie compute 70% of whatever the limit is) and just do long compare for checking if we hit the limit and then don't need to lock. I think we'd only need to lock if updating that computed value on change which would be rare (like JMX)

mattrpav:
Updated to eagerly compute percentage

PR #2283, opened 2026-08-04
[#2282] Drop redundant sequence generator monitor in RegionBroker

mattrpav:
LongSequenceGenerator is private final on RegionBroker and is already thread-safe. The synchronized is redundant.


cshannon (approved):
I have no idea why this is protected by synchronized, it's already using the atomic sequencer as you pointed out so i don't see any reason to not do this.

PR #2399, opened 2026-08-04
[#2398] FutureBrokerInfoTimeout is missed due to logic operator error

mattrpav:


activemq-broker/src/test/java/org/apache/activemq/network/FutureBrokerInfoTimeoutTest.java
```diff
@@ -0,0 +1,160 @@
...
+     */
+    @Test(timeout = 10000)
+    public void testGetTimedThrowsTimeoutExceptionWithinTimeout() throws Exception {
+        AtomicBoolean disposed = new AtomicBoolean(false);
+        FutureBrokerInfo future = new FutureBrokerInfo(null, disposed);
+
+        AtomicBoolean timedOut = new AtomicBoolean(false);
+        AtomicLong elapsed = new AtomicLong(-1);
+
+        Thread t = new Thread(() -> {
```
cshannon: I would re-write these tests and use an executor, it's simpler/cleaner. Here's an example for this test but it applies to all the tests.

```java
@Test(timeout = 10000)
public void testGetTimedThrowsTimeoutExceptionWithinTimeout() throws Exception {
    AtomicBoolean disposed = new AtomicBoolean(false);
    FutureBrokerInfo future = new FutureBrokerInfo(null, disposed);

    AtomicBoolean timedOut = new AtomicBoolean(false);
    AtomicLong elapsed = new AtomicLong(-1);

    ExecutorService executor = Executors.newSingleThreadExecutor();

    executor.submit(() -> {
        long start = System.currentTimeMillis();
        try {
            future.get(200, TimeUnit.MILLISECONDS);
        } catch (TimeoutException e) {
            timedOut.set(true);
        } catch (Exception ignored) {
        } finally {
            elapsed.set(System.currentTimeMillis() - start);
        }
    });
    executor.shutdown();
    assertTrue("get(200ms) should have returned within 3s but is still blocked "
            + "- the timeout is being ignored",
            executor.awaitTermination(3, TimeUnit.SECONDS));
    
    assertTrue("Expected TimeoutException from get(200ms)", timedOut.get());
    assertTrue("Timed out too early: " + elapsed.get() + "ms", elapsed.get() >= 180);
}
```

cshannon (changes requested):

activemq-broker/src/test/java/org/apache/activemq/network/FutureBrokerInfoTimeoutTest.java
cshannon: also it might be worth wrapping the assertion in a try/finally just in case it get stuck:

```java
        try {
            assertTrue("get(200ms) should have returned within 3s but is still blocked "
                            + "- the timeout is being ignored",
                    executor.awaitTermination(3, TimeUnit.SECONDS));
        } finally {
            executor.shutdownNow();
        }
```

cshannon (approved):
LGTM assuming tests pass

PR #2402, opened 2026-08-05
feat(prometheus): Add native /metrics endpoint

hai-ben:
New activemq-prometheus module that adds a `/metrics` endpoint to the broker that vends metrics in the Prometheus format without requiring any sort of open JMX port. The plugin relies on the existing Jetty auth systems and only scrapes MBean during a request, so it doesn't put any additional load on the broker unless someone is actually using it and users define their own Prometheus scrapers and monitoring setup.

It depends only on the jakarta stuff jetty already does.

I've included an example grafana dashboard in this PR just as a way to get up an running quickly (./activemq-prometheus/src/main/resources/example-grafana-dashboard.json). This wouldn't be in the final PR, instead something like it would be uploaded to grafana.com/dashboards.

See Discussion: https://github.com/apache/activemq/discussions/2226

Questions for the community:
- What permissions should be required to reach this endpoint? admins? everyone?
- What do you think about the chosen metrics and names?
- Because it's no load without use, what are your thoughts about having it on by default?


graben:
Maybe an idea for improvement for broker with many queues an topics. Collect metrics periodically in the background and answer http request with the last collected result. This avoids a lot of preasure on the broker and keep getting results back to prometheus "quickly" available. In my experience is better to not not the most actual data but have at most actual data by providing broker stability.

hai-ben:
I like the idea @graben. I know for RabbitMQ, there's an internal metrics store that both their console and their prometheus plugin scrape from, so the plugin still has 0 load on idle, does that pattern fix what you're getting at?

A couple of other parallel additions I might also just add in this vein:
- A simple cache with some configurable TTL, additional callers inside the TTL would represent only marginal load 
- Adding support to stream the response back to callers

hai-ben:
Looking into the source a little more, it looks like everything is already just describing MBeans for metrics anyways. The reason RabbitMQ has that pattern is that those metrics need to be actually aggregated, ActiveMQ MBeans are already ready to use. So I'm not sure I see the savings that pre-caching metrics has over lazily updating a cache in response to a caller.

I'll still add the streaming response and the lazy TTL cache, as those are definitely worthwhile (imo).

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,195 @@
...
+        new MetricDefinition("memory_percent_usage", "Percent (0-100) of destination memory limit used", "MemoryPercentUsage", MetricType.GAUGE),
+        new MetricDefinition("memory_limit_bytes", "Memory limit for this destination in bytes", "MemoryLimit", MetricType.GAUGE),
+        new MetricDefinition("memory_usage_bytes", "Memory used by this destination in bytes", "MemoryUsageByteCount", MetricType.GAUGE),
+        new MetricDefinition("store_message_size_bytes", "Store message size in bytes", "StoreMessageSize", MetricType.GAUGE),
+        new MetricDefinition("average_enqueue_time_milliseconds", "Average time (since last start) messages waited before dispatch", "AverageEnqueueTime", MetricType.GAUGE)
+    };
+
+    @Override
+    protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+        boolean perObject = request != null && "true".equalsIgnoreCase(request.getParameter("per_object"));
```
sergio-d-lemos: I prefer to use `final` as much as possible

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,195 @@
...
+
+        StringWriter output = new StringWriter();
+        PrintWriter writer = new PrintWriter(output);
+
+        try {
+            writeMetrics(ManagementFactory.getPlatformMBeanServer(), writer, perObject);
+        } catch (Exception exception) {
```
sergio-d-lemos: I think we should log this exception too, not just silently drop. So we can know when and what failed

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,195 @@
...
+            writeDestinationMetrics(mBeanServer, writer, "Topic");
+        }
+        writer.flush();
+    }
+
+    private void writeBrokerMetrics(MBeanServer mBeanServer, PrintWriter writer) throws Exception {
+        ObjectName pattern = new ObjectName("org.apache.activemq:type=Broker,brokerName=*");
+        Set<ObjectName> brokers = mBeanServer.queryNames(pattern, null);
+
+        for (MetricDefinition metric : BROKER_METRICS) {
```
sergio-d-lemos: Just as a curiosity, is there any practical difference between iterating the brokers inside the metrics loop or doing the other way around ? 


```
for (ObjectName broker : brokers) {  
   for (MetricDefinition metric : BROKER_METRICS) {
   ...
   }
}
```

Do we need to have the metrics grouped together ?

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    private double getNumber(MBeanServer mBeanServer, ObjectName name, String attribute) {
+        try {
+            Object value = mBeanServer.getAttribute(name, attribute);
+            if (value instanceof Number) {
+                return ((Number) value).doubleValue();
+            }
+        } catch (Exception ignored) {
+            // Some attributes are not available on every ActiveMQ deployment (eg: bridge metrics)
```
sergio-d-lemos: Should we log a warning?

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,195 @@
...
+        } else {
+            writer.println(metric + "{" + labels + "} " + value);
+        }
+    }
+
+    static String sanitizeLabel(String value) {
+        if (value == null) {
+            return "unknown";
+        }
+        return value.replace("\\", "\\\\").replace("\"", "\\\"").replace("\n", "\\n");
```
sergio-d-lemos: Again, just my curiosity. If I have broker called "# TYPE" (or anything else that is meaningful for the prometheus format), it won't mess things up right?

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    private static final class MetricDefinition {
```
sergio-d-lemos, line 234: +1 final classes, final everything!

sergio-d-lemos (commented):
Nice work! I left some simple comments, I understand this is WIP.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: To be honest, not really. Though if we're implementing streaming responses then this makes it so metrics are grouped per object, which seems intuitive to me.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: The name itself is just put blindly into quotes like `activemq_broker_connections{broker="# TYPE"} 42`. If someone has the`"` char in the broker name it won't affect this exporter, but scrapers might struggle. This is why I added the guard for the character `"`, so with a broker name like `\"My evil \n# TYPE Broker` won't cause the scraper trouble.

activemq-prometheus/src/main/webapp/META-INF/LICENSE
mattrpav, line 1: LICENSE file should be added to the archive from the central file, not stored redundantly

mattrpav:
Excited to get a modern metrics endpoint into the broker. I think there are a couple architecture points that need to be sorted out.

Use of hand-crafted JSON output-- to avoid a security attack vector his needs serious character escape handling, or (my preference) leverage Jackson JSON processing library.

hai-ben:
> Excited to get a modern metrics endpoint into the broker. I think there are a couple architecture points that need to be sorted out.
> 
> Use of hand-crafted JSON output-- to avoid a security attack vector his needs serious character escape handling, or (my preference) leverage Jackson JSON processing library.

@mattrpav I'm not sure I understand the security risk JSON _output_ has or the kind of attack you're foreseeing here. Could you tell me more? Maybe link some previous CVEs on this?

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: Updated a bunch of stuff to use final, lmk if you see anything that could be but isn't.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: Agreed.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: Yes. Done.

hai-ben:
@graben on the lazy cache I did some reading and some tests. [Prometheus recommends](https://prometheus.io/docs/instrumenting/writing_exporters/#scheduling):

> Metrics should only be pulled from the application when Prometheus scrapes them, exporters should not perform scrapes based on their own timers. That is, all scrapes should be synchronous.

The logic being that it's whoever's scraping Prometheus's responsibility to manage time-stamps.

I also did some perf testing to see what affect this would actually have on a real broker using worst-practices. I spun up an m5.xlarge broker and put some queues on it (1_000, 5_000, and 25_0000). Callers would scrape Prometheus every 15s. I tried with 1, 5, and 20 callers. I also ran this under no-load and a load of ~2k msg/s. The best practice is to only touch this endpoint from one caller once every 30-60s and only touch the per-objects when you really need them. Here are the results:

PrometheusTest

The CPU load is noticeable, but even with 20 concurrent scrapers on 25k queues with load, scrape times were well under 10s. 

Importantly, the CPU spikes on 20 concurrent scrapers weren't much worse than 1 or 5 scrapers. There are also a lot of edge cases to handle with a lazy cache (deleted queues, race conditions with concurrent scrapes, etc) and it would not provide much benefit against worst practices. And a scheduled cache also puts this non-zero load on the broker, even if it's not being used.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,243 @@
...
+import org.slf4j.LoggerFactory;
+
+public class PrometheusMetricsServlet extends HttpServlet {
+
+    private static final long serialVersionUID = 1L;
+    private static final Logger LOG = LoggerFactory.getLogger(PrometheusMetricsServlet.class);
+    private static final String CONTENT_TYPE = "text/plain; version=0.0.4; charset=utf-8";
+
+    // Metrics can be easily extended by adding them here
+    private static final MetricDefinition[] BROKER_METRICS = {
```
Jerald: Is there an opportunity for this metric list to be automatically generated from the registered MBeans on the broker? I can see how a system could deterministically turn the PascalCase name into a snake_case name for Prometheus (although the names would not be the same as right now).

mattrpav:
> @mattrpav I'm not sure I understand the security risk JSON _output_ has or the kind of attack you're foreseeing here. Could you tell me more? Maybe link some previous CVEs on this?

The write take data from the broker unsanitized. So things like brokerName, queue/topic names, and clientIds are all user-defined strings that create attack surface for malformed json.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,243 @@
...
+
+    @Override
+    protected void doGet(final HttpServletRequest request, final HttpServletResponse response) throws IOException {
+        final boolean perObject = request != null && "true".equalsIgnoreCase(request.getParameter("per_object"));
+        final MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
+
+        // If the mBeanServer is unavailable nothing useful can be produced, so return a 500
+        final Set<ObjectName> brokers;
+        final Set<ObjectName> queues;
+        final Set<ObjectName> topics;
```
Jerald: What's the reason to split queue and topic information like this? Could this be a combined "destinations" set without much impact?

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,243 @@
...
+        try {
+            brokers = mBeanServer.queryNames(new ObjectName("org.apache.activemq:type=Broker,brokerName=*"), null);
+            if (perObject) {
+                // Scraping destinations on brokers with many queues or topics can be expensive.
+                queues = mBeanServer.queryNames(new ObjectName(
+                        "org.apache.activemq:type=Broker,brokerName=*,destinationType=Queue,destinationName=*"), null);
+                topics = mBeanServer.queryNames(new ObjectName(
+                        "org.apache.activemq:type=Broker,brokerName=*,destinationType=Topic,destinationName=*"), null);
```
Jerald: Are there any other `destinationType` values used in ActiveMQ? (Notably I'm unsure what composite/virtual destinations look like)

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,243 @@
...
+        final Map<ObjectName, String> identified = new LinkedHashMap<>();
+        for (final ObjectName broker : brokers) {
+            final String brokerName = resolveStringAttribute(mBeanServer, broker, "BrokerName");
+            if (brokerName != null) {
+                identified.put(broker, "broker=\"" + sanitizeLabel(brokerName) + "\"");
+            }
+        }
+
+        for (final MetricDefinition metric : BROKER_METRICS) {
+            final String metricName = "activemq_broker_" + metric.name;
```
Jerald, line 146: Is prefixes like this the conventional pattern for "namespacing" in the prometheus metrics world?

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,243 @@
...
+        try {
+            final Object value = mBeanServer.getAttribute(name, attribute);
+            if (value instanceof String) {
+                return (String) value;
+            }
+        } catch (final Exception exception) {
+            LOG.debug("Skipping object {}: identity attribute {} unavailable", name, attribute, exception);
+        }
+        return null;
+    }
```
Jerald, line 219: Is there an opportunity to downcast the MBeans to their concrete class as a way to remove some of these string attribute checks?

I'm usually weary of those being brittle in the face of attribute changes, though it's very unlikely to be a concern here.

Jerald (commented):
Long-time ActiveMQ lurker here. I have some interest in Prometheus support myself, very excited to see it happening! 

I have a few opportunities for improvement I noticed, not sure what others think, but none of them seem to impact correctness, just future improvements.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
mattrpav: Yes, they must be split. Queues and Topics are separate namespace regions, so it is legit (and spec compliant) to have a queue and topic with the same name.

Jerald:
@hai-ben would you have time/interest in running another load test in a different scenario?

I know that in some CPU constrained environments with lots of thread churn, ActiveMQ can quickly buckle due to the impact of starving threads of CPU cycles. Given that Prometheus scrapes look to create a relatively-large single-threaded workload, I'm very interested to understand how (if at all) Prometheus scrapes impact ActiveMQ when everything is competing for CPU cycles.

As I saw you're using EC2 instances for testing (m5.xlarge), a comparable instance type with CPU constraints would be something like t3.micro. Running a load test using short-lived clients that open/close dedicated connections should easily induce the sort of issues I'm talking about.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
Jerald: Ah! That's a good point, the destination name is not globally unique. I believe it would be the tuple of `(destination_type, destination_name)` that's unique, does that sound right? (Effectively encoding the same info as the destination URI.)

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: In my mind one of the great things about Prometheus is the ability for people to standardize on dashboards. I'm hesitant to do this because it might create churn for dashboards if MBean names ever change.

Though maybe I'm not understanding something. What's the use case you envision for this?

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,243 @@
...
+
+        for (final MetricDefinition metric : DESTINATION_METRICS) {
+            final String metricName = "activemq_" + typeLower + "_" + metric.name;
+            writeMetadata(writer, metricName, metric.withFormattedHelp(typeLower));
+            for (final ObjectName destination : destinations) {
+                final String brokerName = sanitizeLabel(destination.getKeyProperty("brokerName"));
+                final String destinationName = sanitizeLabel(destination.getKeyProperty("destinationName"));
+                final String labels = String.format("broker=\"%s\",destination=\"%s\"", brokerName, destinationName);
+                writeSample(writer, metricName, labels, getNumber(mBeanServer, destination, metric.attribute));
+            }
```
Jerald, line 192: I'm realizing there's an alternate way to encode these metrics that could be used. Right now, the destination type (queue/topic) is encoded in the metric name itself, despite all destinations sharing a set of emitted metrics (though queues and topics do have a few metrics unique to themselves). But because the underlying metrics are mostly overlapping, the destination type can also be disambiguated by a label instead of a unique metric name.

For example, the current format for a destination metric name is something like this: `activemq_$destType_$metricName` (`activemq_queue_connections`). And it would have labels like this: `broker=$brokerName`, `destination=$destName` (`broker="foo",destination="FOO.BAR"`).

But it should be near equivalent to use a destination name like `activemq_destination_$metricName` and then have labels `broker=$brokerName`, `destinationType=$destType`, and `destination=$destName`. 

Both these approaches are valid (as far as I'm aware), so it's mostly a matter of taste. One upside that I see is having destination type as a label more closely aligns with how the MBeans look. If someone were to convert a tool from using MBeans for metrics and move to Prometheus, I'd expect this approach would be an easier transition (but whether or not that matters is a separate question).

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
```diff
@@ -0,0 +1,243 @@
...
+        }
+        writer.println(metric + "{" + labels + "} " + rendered);
+    }
+
+    static String sanitizeLabel(final String value) {
+        if (value == null) {
+            return "unknown";
+        }
+        // See: https://prometheus.io/docs/instrumenting/exposition_formats/
+        return value.replace("\\", "\\\\").replace("\"", "\\\"").replace("\n", "\\n").replace("\r", "\\r");
```
Jerald: I was just reviewing the prometheus docs on this format, but I realized it doesn't state anything about the CR (`\r`) needing escaping, only the LF (`\n`).

Do you know if this is an undocumented assumption in Prometheus? (I've not run into it before myself.) Or should this escaping be removed?

Either way, it would be great (but I totally understand if impractical) to have a real Prometheus parser used as a control in some sort of test here. That will help validate that ActiveMQ is not just emitting the metrics that are expected, but also emitting them in a format that can be consumed by Prometheus correctly. An adversarial test with `\r` would be a perfect starting point if there's a test suite like this.

mattrpav:
> ActiveMQ can quickly buckle due to the impact of starving threads of CPU cycles. 

@Jerald I suggest you run your scenarios against ActiveMQ v6.3.1+ with Virtual Threads enabled. In addition to Virtual Threads, a number of hotspots have been removed.

I'd be interested if you find hot spots and I'll remove them.

blog: https://medium.com/javarevisited/how-activemq-v6-3-0-boosts-topic-throughput-up-to-93-6b00006f3b82?sharedUserId=mattrpav

hai-ben:
@mattrpav

> The write take data from the broker unsanitized. So things like brokerName, queue/topic names, and clientIds are all user-defined strings that create attack surface for malformed json.

What I was missing was that users that create destinations on the broker would be different than users consuming metrics. And a bad sanitizer could let the former break the latter.

Prometheus output isn't JSON, but I'll swap to the official Prometheus parser/output library to address the concern.

hai-ben:
> @hai-ben would you have time/interest in running another load test in a different scenario?
> 
> I know that in some CPU constrained environments with lots of thread churn, ActiveMQ can quickly buckle due to the impact of starving threads of CPU cycles. Given that Prometheus scrapes look to create a relatively-large single-threaded workload, I'm very interested to understand how (if at all) Prometheus scrapes impact ActiveMQ when everything is competing for CPU cycles.
> 
> As I saw you're using EC2 instances for testing (m5.xlarge), a comparable instance type with CPU constraints would be something like t3.micro. Running a load test using short-lived clients that open/close dedicated connections should easily induce the sort of issues I'm talking about.

Sure @Jerald. I'll try those tests with @mattrpav's suggestion of using virtual threads.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: I'm still relatively new to ActiveMQ so I'm not totally sure. Maybe this @sergio-d-lemos could answer this one?

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
Jerald: My mental model (but contributors please chime in) is that the MBean names (and their attributes) are part of the public API, in the sense that ActiveMQ wouldn't break them without documenting it and a Semver change.

Which is to say, MBeans changing is already a breaking change for anyone using them for monitoring, there's no difference in the impact here for Prometheus.

The use-case I see is that this prevents the MBean metrics and Prometheus metrics from diverging over time. Right now, you took some editorial liberties with translating the MBean attribute names to Prometheus metric names. While I don't have an issue with this per se, it does mean the two ways of monitoring an ActiveMQ broker will necessarily diverge a bit. (This also reduces the effort to expose new metrics to Prometheus, because there simply would be no extra effort.)

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
Jerald: I just did a little digging around, but struggled to find the exact spot defining this `destinationType` field. I think this might be the transitive source though, showing that the temp destination types are also valid: https://github.com/apache/activemq/blob/main/activemq-client/src/main/java/org/apache/activemq/command/ActiveMQDestination.java#L382-L395

Either way, this leads me to believe that you should assume temp destinations have a different type unless you can prove that to not be true. Missing them here would lead to incomplete metric scraping, possibly being very confusing to some people.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: Yes the standard is `namespace_subsystem_name`. So you have `activemq_broker_connections` and `activemq_queue_messages`.

activemq-prometheus/src/main/java/org/apache/activemq/prometheus/PrometheusMetricsServlet.java
hai-ben: No clue. I think we want to shift this to a library anyways which will handle all of this.

Jerald:
> I suggest you run your scenarios against ActiveMQ v6.3.1+ with Virtual Threads enabled. In addition to Virtual Threads, a number of hotspots have been removed.

Oh that's great to hear! I'm still stuck using older versions of ActiveMQ in production, with virtual threads likely a ways out, so that's where my thoughts are coming from. I'm happy for this to be a worry of the past!

Is there an opportunity for this (EDIT: Prometheus) to be backported to the 5.19 branch? Or would this only be expected to be available in 6.x? (Feel free to leave these questions until later if they detract from the review here.)

mattrpav:
> Oh that's great to hear! I'm still stuck using older versions of ActiveMQ in production, with virtual threads likely a ways out, so that's where my thoughts are coming from. I'm happy for this to be a worry of the past!

Keep in mind-- the broker and clients are wire-compatible across versions. You can run 6.3.x broker and 5.19.x clients no problem.

> Is there an opportunity for this (EDIT: Prometheus) to be backported to the 5.19 branch? Or would this only be expected to be available in 6.x? (Feel free to leave these questions until later if they detract from the review here.)

Unfortunately, no. Virtual Threads need JDK 25 support and that is only technically possible in v6.3.x

hai-ben:
@mattrpav I've addressed your comments about using a custom parser and users for prometheus. Latest commit now uses the 3rd the party prometheus library for parsing/serialization and contains a user group for the metrics endpoint.

@Jerald I've also addressed your comments about making sure destination types are handled correctly. Still need to do to that CPU constrained testing we talked about.

cshannon (commented):
I haven't dug into this in depth but at first glance I am wondering why this is necessary to build a native metrics endpoint and do this work ourselves. is there a reason someone can't just use https://github.com/prometheus/jmx_exporter ? That's the standard way to export metrics from JMX for a Java process

jbonofre:
Can you elaborate the rationale here?

My take on that:
1. We can leverage OTLP (I have a PR adding spans, etc, see https://github.com/apache/activemq/pull/1769)
2. The OTLP exporter can be consume by any tool, Prometheus included (but thanks to that we can also support Honeycomb, zipkin, etc)

I'm not against adding `/metrics` endpoint (as a servlet in Jetty), but it sounds very specific, and I don't see why not leveraging OTLP.
If the activemq-prometheus is optional and not included by default, that's totally fine.

hai-ben:
Thanks folks for letting me know the rationale isn't clear. Stepping back, the underlying problem I'm trying to fix is that ActiveMQ exposes MBean attributes but does not provide an out of the box way to map those to make to metrics, which makes each monitoring stack unique deployment specific.

@cshannon, the way I see jmx_exporter, is that it provides a way to make and use a mapping of MBean attributes to metrics, but doesn't provide a mapping, so nothing works out of the box. This is the pain point I'm trying to solve, I just see this Prometheus servlet as the easiest path to fixing that problem. I'm committed to maintaining it.

@jbonofre and for OTel, it's sort of the same problem, it's a way to expose metrics, but it's mapping agnostic with steps needed to get it running out of the box. So this is complementary, OTel could be set up to scrape and export this endpoint.

My goal is to provide end users some batteries included way to get started monitoring their broker right away without needing to have an opinion on MBean attributes.

I'm updating the PR to make this off by default with a toggle in the config and cleaning up the example grafana dashboard.

cshannon:
> @cshannon, the way I see jmx_exporter, is that it provides a way to make and use a mapping of MBean attributes to metrics, but doesn't provide a mapping, so nothing works out of the box. This is the pain point I'm trying to solve, I just see this Prometheus servlet as the easiest path to fixing that problem. I'm committed to maintaining it.
> 

@hai-ben - I guess I am not understanding how this solves a pain point and is the easiest path forward because at the end of the day, the MBean attributes have to be mapped for prometheus in some fashion, whether by writing code like this PR or by writing yaml for the JMX exporter.

Besides requiring writing a lot of code to do the mapping and export manually, this PR also hard codes all the attributes so if new attributes are needed to be added it requires a code change and is not flexible.

With the JMX exporter, a configuration file is provided and that's it. It's very flexible and uses regular expressions so new metrics can be mapped and picked up automatically or just require a simple config change. It's easy to maintain and update and most people are very familiar with it as it's very common in the Java ecosystem. There's probably already plenty of examples online if we google for ActiveMQ. Honestly you could probably just use an AI model to generate one in a few minutes.

So, really I'm just trying to see the benefit of this approach vs the JMX exporter because I still don't see one but maybe I am missing something

sergio-d-lemos:
Hi folks, to share my two cents here:

I agree that this approach does not offer the same flexibility as, for example, using Prometheus JMX exporter. What we gain from this approach is that the complexity is minimal, there are no new dependencies and it just works. I can see the value on that. At least for our use-case, bringing in new dependencies always comes with explicit and implicit costs.

I think there is a tradeoff between hardcoding the MBean attributes here vs using a configuration based approach like JMX Exporter. Here we do it once, with JMX exporter people maintaining their configuration will have to update them. 

In the end of the day this might not be the right solution for everyone. I still think it would benefit many users.

jbonofre:
I'm fine adding a `/metrics` endpoint if it's not enabled by default: the user has to explicitly enabled it by configuration and the `ServletInit` should be no-op if not enabled.

I'm not super concerned by the new dependencies introduced by this PR (it's prometheus related and well scoped).

PR #2405, opened 2026-08-05
[#2404] WIP: Network advisory clean-ups

### LINKED ISSUE #2404: Network Connector dispatch usage of integer creates race

mattrpav:
- [x] dispatch counter race fix
 - [x] timed task to ack advisory messages instead of relying on count-based ack percentage 

This hardens network consumers from appearing slow or hung due to outstanding ack's and being unintentionally closed due to slow consumer strategy.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
@@ -37,7 +37,6 @@ import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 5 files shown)


cshannon:
Responding to: https://github.com/apache/activemq/issues/2404#issuecomment-5199536169

Flushing outstanding acks probably makes sense anyways, but the motivation seems weird to me. (trying to work around stale connections) Anytime you do batching it usually makes sense to have a time limit to flush so you don't wait forever. In fact, I've had to configure bridges to lower the percentage for acking advisories for bridges that are not as busy just so we don't show pending acks for a long time in metrics.

I am mostly just curious why you need to use this as a work around for bad connections? Does the inactivity monitor not work for your use case? That's the tool to detect bad connections and to cause resetting if there are no heart beats, so it seems strange to rely on this for that reason. I would think if the connection is bad the broker would stop getting heart beats.

PR #2408, opened 2026-08-05
[#2418] Improve performance of IntSequenceGenerator

### LINKED ISSUE #2418: Improve performance of IntSequenceGenerator

cshannon:
This adopts the same approach as [LongSequenceGenerator](https://github.com/apache/activemq/blob/cd1dfa298c11dd614fdc16bad0b694639f185991/activemq-client/src/main/java/org/apache/activemq/util/LongSequenceGenerator.java#L25) and uses an Atomic field updater to avoid using synchronized.

The sequence generator is used by ResponseCorrelator so the existing tests will catch any errors.

See https://issues.apache.org/jira/browse/AMQ-7300


mattrpav (approved):
Nice find! LGTM

PR #2432, opened 2026-08-09
[#2431] Skip topic sendLock for non-persistent sends

### LINKED ISSUE #2431: Skip persistent sendLock for non-persistent topic messages
Non-persistent messages can skip the sendLock. We already have a flag for optimizing out persistence. 

This reduces the locks required for non-persistent and Advisory messages to send.

```
canOptimizeOutPersistence()
```

mattrpav:
Non-persistent messages can skip the sendLock. We already have a flag for optimizing out persistence. 

This reduces the locks required for non-persistent and Advisory messages to send.

```
canOptimizeOutPersistence()
```


cshannon (approved):
LGTM, makes sense to optimize by first checking if the message is persistent and if there are consumers before we acquire the sendLock, otherwise we will often be locking for no reason if the persistent check is false later.

Issue #2434, opened 2026-08-09
Reduce MDC add/remove on TransportConnection


mattrpav:
MDC entries are copy-on-write operations and currently every service command sets and removes the connectorURIString as an entry.

Optimizing this measured a 3-6% throughput improvement on M3

The proposed fix solves for this and accounts for nio scenarios where underlying thread services multiple connections by only swapping the value as needed.

PR #2435, opened 2026-08-09
[#2434] Avoid per-command MDC add+remove in TransportConnection.service

mattrpav:
MDC entries are copy-on-write maps in log4j2 — a put per command is a
full thread-context map copy per message. Put only when this thread's
tag actually changes and leave it sticky between commands (same
lifecycle-scoped pattern as the activemq.broker MDC entry), so the
steady-state per-command cost is a single allocation-free MDC.get.

In benchmarks this yields 3-6%

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnection.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnection.java
@@ -128,6 +128,8 @@ public class TransportConnection implements Connection, Task, CommandVisitor {
     private final Transport transport;
     private MessageAuthorizationPolicy messageAuthorizationPolicy;
     private WireFormatInfo wireFormatInfo;
+    // Cached connector URI for the MDC tag; connector.getUri() is stable once started.
+    private String connectorUriString = "unset";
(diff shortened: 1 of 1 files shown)


cshannon (changes requested):
Nice find.

I was thinking this could be simplified and improved even more because we only need to set the value once. We could set it inside the TransportConection start() method after the the transport connection starts, maybe after [this](https://github.com/apache/activemq/blob/9c3add11f743ed97b438f2b630a11644e6772443/activemq-broker/src/main/java/org/apache/activemq/broker/TransportConnection.java#L1151) line.

The connector is passed in the constructor and is final, so this should never change. If we just set this after start I think you could get rid of the caching of the connectionUri entirely and skip the conditional check, just set it once and be done.

mattrpav:
@cshannon I applied the change, but went ahead and moved it up ahead of the brokerInfo being sendAsync(), since there is a chance a service(Command) could come in before it is fully set (ie vm:// transport).

cshannon (approved):
LGTM, could use a description in the comments on the PR instead of just the code itself to help explain why this change was done

PR #2470, opened 2026-08-19
Update to Jetty 12.1.12

cshannon:
Dependabot is throwing an error on the Jetty update so this is a manual update to see if this will fix that error.


mattrpav (approved):
LGTM, tests passed

PR #2474, opened 2026-08-20
Harden soTimeout vs InactivityMonitor

mattrpav:
Harden soTimeout to not conflict with InactivityMonitor. The current TcpTransport read check doesn't count bytes received, only if it looped without blocking or throwing an exception.

### DIFF
--- a/activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransport.java
+++ b/activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransport.java
@@ -613,26 +613,39 @@ public class TcpTransport extends TransportThreadSupport implements Transport, S
     }
 
     protected void initializeStreams() throws Exception {
+        // receiveCounter feeds AbstractInactivityMonitor.readCheck() so only increment when there is data read
         TcpBufferedInputStream buffIn = new TcpBufferedInputStream(socket.getInputStream(), ioBufferSize) {
             @Override
             public int read() throws IOException {
-                receiveCounter.incrementAndGet();
-                return super.read();
+                int result = super.read();
+                if (result >= 0) {
+                    receiveCounter.incrementAndGet();
+                }
+                return result;
             }
             @Override
             public int read(byte[] b, int off, int len) throws IOException {
-                receiveCounter.incrementAndGet();
-                return super.read(b, off, len);
+                int result = super.read(b, off, len);
+                if (result > 0) {
+                    receiveCounter.incrementAndGet();
+                }
+                return result;
             }
             @Override
             public long skip(long n) throws IOException {
-                receiveCounter.incrementAndGet();
-                return super.skip(n);
+                long result = super.skip(n);
+                if (result > 0) {
+                    receiveCounter.incrementAndGet();
+                }
+                return result;
             }
             @Override
-            protected void fill() throws IOException {
-                receiveCounter.incrementAndGet();
-                super.fill();
+            protected int fill() throws IOException {
+                int read = super.fill();
+                if (read > 0) {
+                    receiveCounter.incrementAndGet();
+                }
+                return read;
             }
         };
         //Unread the initBuffer that was used for protocol detection if it exists
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/network/SoTimeoutInactivityMonitorTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/network/SoTimeoutInactivityMonitorTest.java
@@ -1,0 +1,162 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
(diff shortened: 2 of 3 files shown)


activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransport.java
```diff
@@ -613,26 +613,36 @@ public void stop() throws Exception {
...
 
     protected void initializeStreams() throws Exception {
+        // receiveCounter feeds AbstractInactivityMonitor.readCheck() so only increment when there is data read
         TcpBufferedInputStream buffIn = new TcpBufferedInputStream(socket.getInputStream(), ioBufferSize) {
             @Override
             public int read() throws IOException {
-                receiveCounter.incrementAndGet();
-                return super.read();
+                int result = super.read();
+                if (result >= 0) {
```
cshannon, line 621: This should be > 0

activemq-unit-tests/src/test/java/org/apache/activemq/network/SoTimeoutInactivityMonitorTest.java
```diff
@@ -0,0 +1,162 @@
...
+            remoteBroker.stop();
+            remoteBroker.waitUntilStopped();
+        }
+    }
+
+    @Test(timeout = 120_000)
+    public void testInactivityMonitorDetectsDeadLinkWithSoTimeout() throws Exception {
+        var detectMs = runDeadLinkScenario("?wireFormat.maxInactivityDuration=" + MAX_INACTIVITY
+                + "&soTimeout=1000&soWriteTimeout=1000&keepAlive=true");
+        assertTrue("the inactivity monitor must detect a daed link and tear the bridge down even when soTimeout is "
```
cshannon: dead is misspelled here

cshannon (changes requested):

activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransport.java
cshannon: Nevermind, this returns the byte so 0 is valid so >= 0 is correct

activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransport.java
```diff
@@ -613,26 +613,36 @@ public void stop() throws Exception {
...
+                }
+                return result;
             }
             @Override
             public int read(byte[] b, int off, int len) throws IOException {
-                receiveCounter.incrementAndGet();
-                return super.read(b, off, len);
+                int result = super.read(b, off, len);
+                if (result > 0) {
+                    receiveCounter.incrementAndGet();
```
cshannon, line 630: So it would be "more" correct here to add the actual bytes received, which the other NIO transports do. We could change this but it doesn't matter much either way. This value is only used to detect if data is received and the inactivity monitor just does a != check to see if the old value is different than the new value so the exact bytes don't matter. 

In fact, this value is actually an integer so it likely overflows if there's a lot of data but going negative is not an issue in this case and it just wraps around and is fine.

activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransport.java
```diff
@@ -613,26 +613,36 @@ public void stop() throws Exception {
...
+                }
+                return result;
             }
             @Override
             protected void fill() throws IOException {
-                receiveCounter.incrementAndGet();
                 super.fill();
+                receiveCounter.incrementAndGet();
```
cshannon: Looks like this is the main fix here, the fill() method gets called over and over when the soTimeout happens so this prevents incrementing the counter unless it finishes.

What if we update the fill() method to return the number of bytes read? That seems more correct.

```java
@Override
protected int fill() throws IOException {
    int read = super.fill();
    if (read > 0) {
        receiveCounter.incrementAndGet();
    }
    return read;
}
```

The [super method](https://github.com/apache/activemq/blob/e0dca51c19d9c39e5e63915ff8ceb5ad1a2277cf/activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpBufferedInputStream.java#L46) would just become:

```java
protected int fill() throws IOException {
    byte[] buffer = internalBuffer;
    count = 0;
    position = 0;
    int n = in.read(buffer, position, buffer.length - position);
    if (n > 0) {
        count = n + position;
    }
    return n;
}
```

cshannon (changes requested):

activemq-client/src/main/java/org/apache/activemq/transport/tcp/TcpTransport.java
mattrpav: That's an improvement. The method signature change is to the internal TcpBufferedInputStream class, so that is safe as its not a public API or SPI used by folks to extend the broker.

cshannon (approved):
LGTM assuming tests are good

PR #2475, opened 2026-08-20
Fix network bridge local-side close

mattrpav:
Currently, a local-side connection close is treated as the same as a broker shutdown. The bridge is stopped, but not removed from activeBridges or restarted.

This fix differentiates between an administrative close (inactivity monitor, connection.close(), etc.) to ensure the bridge is closed, cleared and recreated.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
@@ -98,6 +98,7 @@ import org.apache.activemq.transport.TransportDisposedIOException;
 import org.apache.activemq.transport.TransportFilter;
 import org.apache.activemq.transport.failover.FailoverTransport;
 import org.apache.activemq.transport.tcp.TcpTransport;
+import org.apache.activemq.util.IOExceptionSupport;
 import org.apache.activemq.util.IdGenerator;
 import org.apache.activemq.util.IntrospectionSupport;
 import org.apache.activemq.util.LongSequenceGenerator;
@@ -1328,12 +1329,13 @@ public abstract class DemandForwardingBridgeSupport implements NetworkBridge, Br
                 } else if (command.isBrokerInfo()) {
                     futureLocalBrokerInfo.set((BrokerInfo) command);
                 } else if (command.isShutdownInfo()) {
-                    if (brokerService.isStopping() || brokerService.isStopped()) {
+                    ShutdownInfo info =  (ShutdownInfo) command;
+                    if (brokerService.isStopping() || brokerService.isStopped() || info.getError() == null) {
                         LOG.info("{} Shutting down {}", configuration.getBrokerName(), configuration.getName());
                         stop();
                     } else {
                         // Administrative shutdown via .stop() or SlowConsumerStrategy needs lifecycle clean-up
-                        serviceLocalException(new IOException("Local network connection was shut down while the broker is still running"));
+                        serviceLocalException(IOExceptionSupport.create(info.getError()));
                     }
                 } else if (command.getClass() == ConnectionError.class) {
                     ConnectionError ce = (ConnectionError) command;
--- a/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
@@ -1328,8 +1328,13 @@ public abstract class DemandForwardingBridgeSupport implements NetworkBridge, Br
                 } else if (command.isBrokerInfo()) {
                     futureLocalBrokerInfo.set((BrokerInfo) command);
                 } else if (command.isShutdownInfo()) {
-                    LOG.info("{} Shutting down {}", configuration.getBrokerName(), configuration.getName());
-                    stop();
+                    if (brokerService.isStopping() || brokerService.isStopped()) {
+                        LOG.info("{} Shutting down {}", configuration.getBrokerName(), configuration.getName());
+                        stop();
+                    } else {
+                        // Administrative shutdown via .stop() or SlowConsumerStrategy needs lifecycle clean-up
+                        serviceLocalException(new IOException("Local network connection was shut down while the broker is still running"));
+                    }
                 } else if (command.getClass() == ConnectionError.class) {
                     ConnectionError ce = (ConnectionError) command;
                     serviceLocalException(ce.getException());
--- a/activemq-unit-tests/src/test/java/org/apache/activemq/network/NetworkBridgeLocalConnectionCloseReconnectTest.java
+++ b/activemq-unit-tests/src/test/java/org/apache/activemq/network/NetworkBridgeLocalConnectionCloseReconnectTest.java
@@ -19,6 +19,7 @@ package org.apache.activemq.network;
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
 
+import java.io.IOException;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import jakarta.jms.Connection;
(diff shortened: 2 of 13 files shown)


activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
```diff
@@ -1328,8 +1328,13 @@ public void onCompletion(FutureResponse future) {
                 } else if (command.isBrokerInfo()) {
                     futureLocalBrokerInfo.set((BrokerInfo) command);
                 } else if (command.isShutdownInfo()) {
-                    LOG.info("{} Shutting down {}", configuration.getBrokerName(), configuration.getName());
-                    stop();
+                    if (brokerService.isStopping() || brokerService.isStopped()) {
```
cshannon: Shouldn't this be checking the started/disposed flags on the bridge and not the broker?

This PR doesn't make sense with how it is currently constructed (checking if the broker is shut down) because it is just a work around to solve the issue with SlowConsumerStrategy but breaks normal shutdown cases.

Bridges can be stopped administratively as you point out while the broker is still running so If someone shuts down a bridge, it _should not_ be restarted. Arbitrarily processing an IOException causes it to get restarted. Even if it didn't, processing an exception doesn't make sense for a normal condition.

One solution for the SlowConsumerStrategy is to just disable it for bridges, which is already an option and I pointed out previously. 

Or if you really need to use it, then figure out a way for it to cause the exception during shut down to restart but not do it during the normal shutdown case. During a normal bridge stop there should be flags set (which I pointed out in the first sentence) so maybe you could check the started or disposed flags

cshannon (changes requested):

activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java
cshannon: I took a closer look and a couple of things:

1. It looks like serviceLocalCommand() [doesn't](https://github.com/mattrpav/activemq/blob/3a420863659ac00d6435e23e446586a5d76fe2ca/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L1252) process commands if the bridge is stopped and marked as disposed, so checking for that flag isn't necessary. This is also good as my main concern here is triggering a restart of the bridge during proper shut down and this wouldn't get executed if stopped. However, it's still interesting that the method will handle ShutdownInfo and do a stop so I am wondering what other legitimate cases it could be used so I am still looking and not sure it's a good idea to do this. The bridge stuff is confusing and there are 2 or 3 transports (if duplex) running.
2. The second thing I noticed is that the abort slow consumer strategy on connection close should trigger an IOException on the connection which I would think should cause the bridge to detect the error and reconnect. It is done [here](https://github.com/apache/activemq/blob/e0dca51c19d9c39e5e63915ff8ceb5ad1a2277cf/activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/AbortSlowConsumerStrategy.java#L165). I'm assuming that doesn't do that? I'm wondering if we could improve this somehow.

mattrpav: See: the serviceRemoteCommand handling. There, it also eventually calls to fireBridgeFailed(). This change is bringing the serviceLocalCommand to parity with the serviceRemoteCommand. No new behavior introduced.

activemq-unit-tests/src/test/java/org/apache/activemq/network/NetworkBridgeLocalConnectionCloseReconnectTest.java
```diff
@@ -0,0 +1,166 @@
...
+        produce(5);
+        assertTrue("bridge should forward before the close",
+                Wait.waitFor(() -> remoteReceived.get() >= 5, 20_000, 10));
+
+        // The bridge's local side is a vm:// client of the local broker.
+        var oldBridge = nc.activeBridges().iterator().next();
+        var bridgeLocalConnection = findVmConnection(localBroker);
+        assertNotNull("expected the bridge's local vm:// connection on the local broker", bridgeLocalConnection);
+        LOG.info("stopping the bridge's local connection server-side: {}", bridgeLocalConnection);
+        bridgeLocalConnection.stop();
```
cshannon: This should probably simulate passing an IOException just like the slow consumer strategy does

cshannon (commented):
@mattrpav - Ok, I played around with this a bunch and I think it will work after all because of the **disposed** flag. That flag gets checked before anything happens in both [serviceLocalCommand](https://github.com/apache/activemq/blob/e0dca51c19d9c39e5e63915ff8ceb5ad1a2277cf/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L1252) and [serviceRemoteCommand](https://github.com/apache/activemq/blob/e0dca51c19d9c39e5e63915ff8ceb5ad1a2277cf/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L752). The disposed flag gets set to true when a bridge is stopped on purpose.

My main concern here was making sure that if the ShutDownInfo command is received that we only process and trigger the restart if there was an unexpected close. On a normal bridge shtudown the command won't get processed because disposed is set to true so all of that is skipped.

Also, you pointed out that the serviceRemoteCommand does the same thing on ShutdownInfo by [calling](https://github.com/apache/activemq/blob/e0dca51c19d9c39e5e63915ff8ceb5ad1a2277cf/activemq-broker/src/main/java/org/apache/activemq/network/DemandForwardingBridgeSupport.java#L856) the exception handler which triggers the bridge failed and reconnect as well so that further verifies this should be ok.

I made one comment inline that I think the test should simulate how the connection is stopped by the slow consumer strategy with an exception [here](https://github.com/apache/activemq/blob/e0dca51c19d9c39e5e63915ff8ceb5ad1a2277cf/activemq-broker/src/main/java/org/apache/activemq/broker/region/policy/AbortSlowConsumerStrategy.java#L165)

Also it would be good to expand on the tests:
1. I would make sure duplex bridges are also handled correctly with reconnect
2. You could try and test that the bridge doesn't reconnect if you stopped and and connection.serviceException() is called (this might already be tested by other bridge tests not sure)

activemq-unit-tests/src/test/java/org/apache/activemq/network/NetworkBridgeLocalConnectionCloseReconnectTest.java
mattrpav: There is now a test for both Abort Slow Consumer strategy modes

cshannon (commented):
@mattrpav - I've been testing this in detail and I still think there are issues. 

The biggest problem I noticed is the serviceLocalException() is firing now when it doesn't need to which was my fear here from the beginning.

When a remote broker is shut down or can't be connected to, the bridge gets notified fo the remote exception and that triggers the bridge failure callback and  the discovery listener triggers reconnect. This tries to connect and fails, so it then calls close on the local connection because it needs stop. There's no error on the local broker connect as we are trying to stop legitimately. Unfortunately because it calls stop, the serviceLocalException() now gets called and tries to retrigger another connection and spits out a wrong error. it is now causing a duplicate attempt when we should only be trying to connect by the discovery listener.

To fix this I played around with a few solutions and I came up with https://github.com/mattrpav/activemq/pull/7

It updates the transport to allow passing any transport exception to the transport stop() call and transports can optionally decide to use it (by default the original stop() is called and we ignore it). The VM transport will take that and pass it to ShutdownInfo() which has been updated with a transient field to capture an error. 

Now when we process the ShutdownInfo we can check if there is an error set and know whether or not we can stop normally or need to call sericeLocalException(). The other nice side effect of this change is we can pass the original exception to sericeLocalException(). So you will now notice in the test output for slow consumer you will see the original exception which is helpful:

```
INFO  DemandForwardingBridgeSupport  - Network connection between vm://local#0 and tcp://localhost/127.0.0.1<ports> shutdown due to a local error: java.io.IOException: 1 Consumers was slow too often (>-1) or too long (>2000): 
```

cshannon (approved):
LGTM, I ran through a bunch of tests and things seem to work correctly 

I think we are good to merge and backport to 6.3.x as long as CI doesn't find anything

Issue #2446, opened 2026-08-11
connecting to amq with websockets (port 61614) throws java.lang.NoClassDefFoundError: org/eclipse/jetty/websocket/server/internal/CompletedUpgradeRequest  on server


nathandeklepper:
On 6.3.x websocket connections fail with the exception pasted below.

I've tested this against 6.1.x, and 6.2.x where it works correctly.

Steps to reproduce...
Download activemq 6.3.1 tar.gz from amq classic downloads page.
Uncomment ws and stomp connectors in conf/activemq.xml
Connect with a ws/stomp client on port 61614 (should be reproducible with any protocol over ws).

The issue can also be produced through a local reverse proxy terminating tls, in which case the client talks wss on port 443.

java.lang.NoClassDefFoundError: org/eclipse/jetty/websocket/server/internal/CompletedUpgradeRequest
	at org.eclipse.jetty.ee11.websocket.server.internal.JettyServerFrameHandlerFactory.newFrameHandler(JettyServerFrameHandlerFactory.java:45)
	at org.eclipse.jetty.websocket.core.server.internal.CreatorNegotiator.negotiate(CreatorNegotiator.java:70)
	at org.eclipse.jetty.websocket.core.server.internal.AbstractHandshaker.upgradeRequest(AbstractHandshaker.java:70)
	at org.eclipse.jetty.websocket.core.server.internal.HandshakerSelector.upgradeRequest(HandshakerSelector.java:45)
	at org.eclipse.jetty.websocket.core.server.WebSocketMappings.upgrade(WebSocketMappings.java:296)
	at org.eclipse.jetty.websocket.core.server.WebSocketMappings.upgrade(WebSocketMappings.java:264)
	at org.eclipse.jetty.ee11.websocket.server.JettyWebSocketServlet.service(JettyWebSocketServlet.java:210)
	at jakarta.servlet.http.HttpServlet.service(HttpServlet.java:723)
	at org.eclipse.jetty.ee11.servlet.ServletHolder.handle(ServletHolder.java:754)
	at org.eclipse.jetty.ee11.servlet.ServletHandler$ChainEnd.doFilter(ServletHandler.java:1688)
	at org.eclipse.jetty.ee11.servlet.ServletHandler$MappedServlet.handle(ServletHandler.java:1602)
	at org.eclipse.jetty.ee11.servlet.ServletChannel.dispatch(ServletChannel.java:871)
[...]

PR #2482, opened 2026-08-21
[BAD PR] [#2446] Add missing jetty dependency to assembly for websockets

mattrpav:


mattrpav:
BAD PR -- accidentally made to point back to 'main' vs 'activemq-6.3.x'

PR #2504, opened 2026-08-28
[6.3.x] Fix: Assembly startup logs harmless WARN about schemas (#2501)

mattrpav:
The schemas are included internally to tomcat's jasper support.

These are harmeless for running in Jetty

(cherry picked from commit aa6c07ecab9b7166016aec25bdc3616683a1ef62)


jbonofre (approved):
I don't think it's a fix, it's more an improvement to me.

jbonofre:
Should the same apply on `main`?

mattrpav:
@jbonofre this is a fix for spurious WARN logs as part of the Jetty 12.1 migration. This is a clean-up so users don't get WARN messages that are not actionable.

The initial PR is already merged into main.

PR #2505, opened 2026-08-31
Improve GC checkpoint algorithm for empty durables

### LINKED ISSUE #2438: Improve KahaDB GC algorithm for rewriting empty durable subscriptions
The garbage collection algorithm for cleaning up journal files will re-write durable subscriptions that are empty to prevent blocking a file from being cleaned up. However, the current algorithm only moves along subscriptions that are in the first file in the GC, vs any file in the set which can prevent cleaning up files in certain scenarios.

**Todo:** Fill in with more details on the nature of the problem (when/how it happens) and how the fix will work

cshannon:
This improves the garbage collection algorithm by re-writing empty durables that exist in any file that is eligible for garbage collection. This allows a single GC cycle to remove any files that have no other commands that need to be kept and free up space quicker.

The previous algorithm would only advance an empty durable subscription command if it existed in the first file in the set. This would cause empty durable subs that were written to other files from being moved in some cases (such as another durable that was offline with pending messages in the same file). In the worst case it could prevent any files from being GC'd, and even in the best case it also meant it would take several GC cycles to actually move the durables which meant keeping files around longer than needed.

Closes #2438

### DIFF
--- a/activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/JournalDurableSubCheckpointTest.java
+++ b/activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/JournalDurableSubCheckpointTest.java
@@ -1,0 +1,237 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/JournalDurableSubCheckpointTest.java
```diff
@@ -0,0 +1,237 @@
...
+import org.apache.activemq.command.ActiveMQQueue;
+import org.apache.activemq.command.ActiveMQTopic;
+import org.apache.activemq.store.kahadb.disk.journal.Journal.JournalDiskSyncStrategy;
+import org.junit.After;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
```
mattrpav, line 42: Can this test be @Category(Parallel)?

mattrpav: Can't do parallel test yet, this is in the kahadb-store module

mattrpav (approved):
LGTM

PR #2506, opened 2026-08-31
[6.3.x] Improve GC checkpoint algorithm for empty durables (#2505)

cshannon:
This improves the garbage collection algorithm by re-writing empty durables that exist in any file that is eligible for garbage collection. This allows a single GC cycle to remove any files that have no other commands that need to be kept and free up space quicker.

The previous algorithm would only advance an empty durable subscription command if it existed in the first file in the set. This would cause empty durable subs that were written to other files from being moved in some cases (such as another durable that was offline with pending messages in the same file). In the worst case it could prevent any files from being GC'd, and even in the best case it also meant it would take several GC cycles to actually move the durables which meant keeping files around longer than needed.

(cherry picked from commit c1bb9d6613b7713449e588056cf6bfbbfa92988c)


cshannon:
This is targeted for 6.3.3 so this should wait for merger until 6.3.2 is finalized and the version is bumped

PR #2515, opened 2026-09-03
[6.3.x] fix(jmx): avoid NPE in ConnectionView.getWireFormatInfo (#2270)

### LINKED ISSUE #2203: NPE at ConnectionView.java > getWireFormatInfo()

cshannon:
getRemoteWireFormatInfo() can be null before wire-format negotiation completes (or for transports that never set it). Calling toString() on that null caused an NPE when the JMX attribute was read.

Return the existing "WireFormatInfo not available" placeholder when the remote WireFormatInfo is missing.

Fixes #2203

(cherry picked from commit b3f861c41b52a07a54e62b3f79052735221b74bd)

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectionView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/ConnectionView.java
@@ -23,6 +23,7 @@ import javax.management.ObjectName;
 
 import org.apache.activemq.broker.Connection;
 import org.apache.activemq.broker.TransportConnection;
+import org.apache.activemq.command.WireFormatInfo;
 import org.apache.activemq.util.IOExceptionSupport;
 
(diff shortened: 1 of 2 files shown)


mattrpav (approved):
LGTM

PR #2517, opened 2026-09-03
fix: MQTTWireFormat rejecting valid 4-byte remaining lengths

hai-ben:
Increment in the check loop happens before the check, causing an error. Additionally this code is [practically duplicated here](https://github.com/apache/activemq/blob/main/activemq-mqtt/src/main/java/org/apache/activemq/transport/mqtt/MQTTCodec.java#L165-L167C18). That code has it correct, but it looks tricky to abstract into a shared class.

See: https://github.com/apache/activemq/issues/2518

### DIFF
--- a/activemq-mqtt/src/test/java/org/apache/activemq/transport/mqtt/MQTTCodecTest.java
+++ b/activemq-mqtt/src/test/java/org/apache/activemq/transport/mqtt/MQTTCodecTest.java
@@ -378,12 +378,15 @@ public class MQTTCodecTest {
         publish.topicName(new UTF8Buffer(topic));
         publish.payload(new Buffer(new byte[payloadSize]));
 
-        DataByteArrayOutputStream output = new DataByteArrayOutputStream();
-        wireFormat.marshal(publish.encode(), output);
-        Buffer marshalled = output.toBuffer();
+        Buffer marshalled;
+        try (DataByteArrayOutputStream output = new DataByteArrayOutputStream()) {
+            wireFormat.marshal(publish.encode(), output);
+            marshalled = output.toBuffer();
+        }
 
-        DataByteArrayInputStream input = new DataByteArrayInputStream(marshalled);
-        codec.parse(input, marshalled.length());
+        try (DataByteArrayInputStream input = new DataByteArrayInputStream(marshalled)) {
+            codec.parse(input, marshalled.length());
+        }
 
         assertEquals("Expected one frame from a valid 4-byte Remaining Length", 1, frames.size());
         PUBLISH decoded = new PUBLISH().decode(frames.get(0));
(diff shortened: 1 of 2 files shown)


sergio-d-lemos:
LGTM

mattrpav:
Do you know which MQTT client and library was used?

hai-ben:
@mattrpav I've put an issue here: https://github.com/apache/activemq/issues/2518

I was able to reproduce with paho-mqtt and MQTT.js. I don't see a reason other clients wouldn't experience this.

mattrpav:
This is the correct fix.

activemq-mqtt/src/test/java/org/apache/activemq/transport/mqtt/MQTTCodecTest.java
```diff
@@ -364,4 +364,51 @@ public void testUnmarshalInvalidRemainingLengthField() {
...
+        final int overhead = 2 + topic.length() + 2;
+        final int payloadSize = 2_097_152 - overhead;
+
+        PUBLISH publish = new PUBLISH();
+        publish.qos(QoS.AT_LEAST_ONCE);
+        publish.messageId((short) 1);
+        publish.topicName(new UTF8Buffer(topic));
+        publish.payload(new Buffer(new byte[payloadSize]));
+
+        DataByteArrayOutputStream output = new DataByteArrayOutputStream();
```
mattrpav: try-with-resources here

activemq-mqtt/src/test/java/org/apache/activemq/transport/mqtt/MQTTCodecTest.java
```diff
@@ -364,4 +364,51 @@ public void testUnmarshalInvalidRemainingLengthField() {
...
+        wireFormat.marshal(publish.encode(), output);
+        Buffer marshalled = output.toBuffer();
+
+        DataByteArrayInputStream input = new DataByteArrayInputStream(marshalled);
```
mattrpav: try-with-resources here

hai-ben:
@mattrpav added the try blocks

cshannon (approved):
> Increment in the check loop happens before the check, causing an error. Additionally this code is [practically duplicated here](https://github.com/apache/activemq/blob/main/activemq-mqtt/src/main/java/org/apache/activemq/transport/mqtt/MQTTCodec.java#L165-L167C18). That code has it correct, but it looks tricky to abstract into a shared class.

Nice catch, thanks for the PR and tests. As you also noted, we do have duplicate logic which is a bit annoying (this is just because of the differences between the NIO and blocking transports). We can look at a refactor in a future update to try and consolidate the logic because it doesn't make a lot of sense to be duplicating it.

cshannon:
I think this is good to merge as long as CI passes

hai-ben:
Can we also backport this to 5.19?

cshannon:
> Can we also backport this to 5.19?

It already was and is targeted for 5.19.12: https://github.com/apache/activemq/pull/2522

Issue #1770, opened 2026-03-14
Support Client IP addresses allowlist/denylist on the TransportConnectors


jbonofre:
Currently, the only way to filter IP addresses for the users is to create a custom broker plugin to intercept and filter.

For security reason and integration in the infrastructure, it's a classic approach to filter (allowlist/denylist) clients IP addresses (or CIDR).

This can be added at `TransportConnector` level, configured directly on the URI.

anmol-saxena-14:
Hi @jbonofre ,
Can I work on this ?
Thanks

jbonofre:
@anmol-saxena-14 I already have something on this one. I can create the PR and you can review.

PR #2539, opened 2026-09-07
[#1770] Feature: IP address allow/deny support

mattrpav:
This feature adds IP allow/deny support at the transport layer, before any protocol negotiation. On deny, connections are quietly closed and a counter is incremented.

- [x] Addresses may be entered in CIDR format or exact IP value.
- [x] Allow/Deny list supports string file: uri support to load from a properties file (one CIDR per-line) with default DoS/memory bloat protections in place (10MB and 100,000 entries). 

Notes:
1. Failover will retry indefinitely
2. SSl and non-SSL see an EOF or connection closed by peer error.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/Connector.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/Connector.java
@@ -86,6 +86,44 @@ public interface Connector extends Service {
 
     long getMaxConnectionExceededCount();
 
+    /** @return the configured remote address allow list, comma separated CIDRs or a file: URI, or null */
+    String getAllowList();
+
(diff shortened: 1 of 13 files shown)


jbonofre:
As you can see in the issue, I had something about that. I would appreciate you ping me before doing work (not the first time it happens).

mattrpav:
@jbonofre that's my bad. I saw this branch sitting in my local since the AMQ-9800 days, so I rebased it to get it up. I'll be sure to check in next time!

jbonofre:
@mattrpav that's all right. Next time, just ping on issue or PR to avoid dual work 😄

PR #2540, opened 2026-09-08
Fix concurrent advisory message evaluation (#2519)

### LINKED ISSUE #2519: 6.2.7 and above (6.3.2 inclusive): "Failed to remove consumer" NPE (ConditionalNetworkBridgeFilterFactory) during broker shutdown - stale consumers left behind
# Issue body (paste below the title field)

## Bug Report

**Describe the bug**

After upgrading to ActiveMQ Classic 6.2.7 (also 6.2.8), brokers in a broker
network log `Failed to remove consumer: ID:...` with a `NullPointerException`
when a broker is stopped/restarted or a network bridge drops under load. The
consumer/connection cleanup is interrupted, so consumers are left registered
(stale connections) until the broker is restarted.

The NPE comes from the advisory path introduced by "Send advisory messages
using Broker connection context" (#2071 / #2075):

```
WARN  | Failed to remove consumer: ID:... | TransportConnection
java.lang.NullPointerException: Cannot invoke "...isQueue()" because the
return value of "...MessageEvaluationContext.getDestination()" is null
	at ConditionalNetworkBridgeFilterFactory$ConditionalNetworkBridgeFilter
	   .matchesForwardingFilter(ConditionalNetworkBridgeFilterFactory.java:116)
	...
	at AdvisoryBroker.removeConsumer(AdvisoryBroker.java:400)
	at TransportConnection.processRemoveConsumer(TransportConnection.java:739)
```

We believe the shared broker-admin `ConnectionContext` (one mutable
`MessageEvaluationContext`) is cleared by one concurrent advisory dispatch
while another is evaluating a conditional network bridge filter.

**Expected behavior per documentation**

This should work correctly per ActiveMQ's documented behavior:

- Broker networks with demand forwarding and replay-when-no-consumers
  (`ConditionalNetworkBridgeFilterFactory`) are a supported, documented
  topology; the [6.2.7 release notes](https://activemq.apache.org/components/classic/download/classic-06-02-07)
  only describe the advisory change as a
  permission improvement (the broker now publishes advisories under its own
  context, so users no longer need write access to advisory topics) - no
  change in dispatch or cleanup semantics is documented.
- Consumers and connections are expected to be removed cleanly when clients
  disconnect and during broker shutdown/reconnect; no stale consumers should
  remain.
- Advisory messages are broker-internal bookkeeping and should not be able to
  break consumer cleanup.

Reality on 6.2.7/6.2.8 with an unchanged 6.2.6 configuration: the NPE above
appears during storms and cleanup is left incomplete - which is why we treat
this as a regression.

**To Reproduce**

- A docker-compose reproduction (3 brokers, loadgen, one command) will be
  attached to this issue; steps to run it are included in the attachment.
- In short: a broker network with `ConditionalNetworkBridgeFilterFactory`
  (`replayWhenNoConsumers`) on the destination policies and a
  network-mirrored wildcard topic subscription (a `>` topic consumer).
  With many concurrent consumer create/remove operations, gracefully stop
  one broker; clients reconnect -> the NPE below appears.
- Expected on 6.2.7 in the first storm round; on 6.2.6 the same setup stays
  clean.

**Expected behavior**

Consumer/connection removal completes cleanly during shutdown/reconnect
storms; no NPE, no stale consumers.

**Your Environment**

- Version: 6.2.7 (also reproduced on 6.2.8; not present in 6.2.6)
- Configuration: static network connectors (duplex/dynamic demand),
  `conditionalNetworkBridgeFilterFactory` on queue/topic policies
- OS: Linux (docker)

**Additional context**

Introducing change: commit `15e78425`, PRs #2071/#2075. A null guard at
`ConditionalNetworkBridgeFilterFactory:116` stops this exact NPE, but giving
each advisory dispatch its own evaluation context would fix the underlying
race.

[docker.zip](https://github.com/user-attachments/files/31827824/docker.zip)

jjj-n:
Concurrent advisory sends share the broker admin context's mutable message evaluation state. When one topic dispatch clears that state while another evaluates a conditional network bridge filter, consumer removal can fail with a null destination and leave cleanup incomplete.

Give each advisory producer exchange a copied connection context with its own `NonCachedMessageEvaluationContext`. Add a `ConnectionContext.copy(MessageEvaluationContext)` overload to preserve the existing connection settings and broker permissions; the no-argument `copy()` retains its existing behavior.

The regression test pauses one consumer-removal advisory while another completes, then verifies that both removals succeed, both removal advisories arrive, and the destinations have no remaining consumers. It exercises the real conditional network bridge filter and also checks that broker identity and disabled advisory flow control are preserved.

Validation:

- `mvn -o -B -pl activemq-broker -Dsurefire.enableProcessChecker=ping test`: 88 tests passed on JDK 21.
- The two new tests passed in 10 consecutive JDK 17 runs. Replacing `AdvisoryBroker` with the version from upstream commit `5d3c979` reproduces the reported NPE.
- Existing `NetworkRemovesSubscriptionsTest`: all 5 tests passed on JDK 17, compiled from source and run directly with JUnitCore against the broker test classpath.
- `mvn -o -B -pl activemq-broker apache-rat:check`: passed.

The `ping` option follows the repository's Windows 11 guidance for systems without `wmic`. The full reactor suite and the reporter's three-broker Docker workload were not run.

Fixes #2519.

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/advisory/AdvisoryBroker.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/advisory/AdvisoryBroker.java
@@ -63,6 +63,7 @@ import org.apache.activemq.command.ProducerInfo;
 import org.apache.activemq.command.RemoveSubscriptionInfo;
 import org.apache.activemq.command.SessionId;
 import org.apache.activemq.filter.DestinationPath;
+import org.apache.activemq.filter.NonCachedMessageEvaluationContext;
 import org.apache.activemq.state.ProducerState;
 import org.apache.activemq.usage.Usage;
 import org.apache.activemq.util.IdGenerator;
@@ -1095,8 +1096,9 @@ public class AdvisoryBroker extends BrokerFilter {
 
     protected ProducerBrokerExchange newAdvisoryProducerExchange() {
         final ProducerBrokerExchange producerExchange = new ProducerBrokerExchange();
+        // Concurrent advisory sends must not share mutable message evaluation state.
         producerExchange.setConnectionContext(Objects.requireNonNull(advisoryConnectionContext.get(),
-                "Advisory ConnectionContext must not be null"));
+                "Advisory ConnectionContext must not be null").copy(new NonCachedMessageEvaluationContext()));
         producerExchange.setMutable(true);
         producerExchange.setProducerState(new ProducerState(new ProducerInfo()));
         return producerExchange;
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/ConnectionContext.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/ConnectionContext.java
@@ -80,7 +80,18 @@ public class ConnectionContext {
     }
 
     public ConnectionContext copy() {
-        ConnectionContext rc = new ConnectionContext(this.messageEvaluationContext);
+        return copy(this.messageEvaluationContext);
+    }
+
+    /**
+     * Copies this connection context using the supplied message evaluation context.
+     * A separate evaluation context allows the copy to evaluate messages independently.
+     *
+     * @param messageEvaluationContext the evaluation context to use in the copy
+     * @return a copy with the supplied message evaluation context
+     */
+    public ConnectionContext copy(MessageEvaluationContext messageEvaluationContext) {
+        ConnectionContext rc = new ConnectionContext(messageEvaluationContext);
         rc.connection = this.connection;
         rc.connector = this.connector;
         rc.broker = this.broker;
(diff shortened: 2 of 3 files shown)


cshannon (approved):
Thanks for the fix! 

This was broken by #2071, which made the change to always use the broker context to send advisories which allowed getting rid of the requirement for all users to have permissions on advisory topics to publish. I had cached the ConnectionContext to avoid having to create a new one for each send but I forgot that the evaluation context is mutable.

It still makes sense to cache the advisory the advisory context because we can configure it how we want (such as flow control false) but now it's just treated as a template and copied each time which makes sense. The context is small (mostly null fields) and we already have to create a new ProducerBrokerExchange for every send, so the impact here will be very minimal and should fix the issue.

I'll merge and backport, assuming the tests are all good.

PR #2543, opened 2026-09-08
Flaky test and execution time improvements

mattrpav:
- [x] Reduce activemq-http test execution time from 25min to ~10min
 - [x] Reduce activemq-runtimeconfig test execution from 8min to ~1min


cshannon (changes requested):
I would not group together a bunch of unrelated changes like this, I would split it up into logical PRs with the same related tests or fixes so it's easier to review and roll back if there is an issue

mattrpav:
Replaced by [#2544] [#2545] [#2546]

PR #2544, opened 2026-09-08
Fix flaky ExceptionListenerTest: wait for transport-failed before createSession

mattrpav:
The security exception fires the ExceptionListener and marks the transport failed on two separate async tasks in ActiveMQConnection.onException. Waiting only for the listener let createSession race a disposing transport and get JMSException("Stopped.") instead of the expected ConnectionFailedException.


activemq-unit-tests/src/test/java/org/apache/activemq/usecases/ExceptionListenerTest.java
```diff
@@ -107,6 +108,20 @@ public boolean isSatisified() throws Exception {
...
         assertTrue("expected exception: " + expected, expected.getCause().getCause() instanceof SecurityException);
 
+        // ActiveMQConnection.onException delivers the security failure to the
+        // ExceptionListener and marks the transport failed on two independent
+        // async tasks. The listener firing above does not guarantee the transport
+        // is failed yet, so wait for the failed state; otherwise createSession can
+        // race a mid-dispose transport and see JMSException("Stopped.") instead of
+        // the expected ConnectionFailedException.
+        final ActiveMQConnection amqConnection = (ActiveMQConnection) connection;
+        Wait.waitFor(new Wait.Condition() {
```
cshannon:
```suggestion
        assertTrue(Wait.waitFor((Condition) amqConnection::isTransportFailed, 5000, 10));
```

It makes sense to speed this up, the default wait can go 30 seconds and checks every second (we should probably change the defaults in another PR). Also you can make it a lot smaller using lambdas

cshannon (changes requested):

activemq-unit-tests/src/test/java/org/apache/activemq/usecases/ExceptionListenerTest.java
cshannon: Also should assert true that and not just continue if it fails

PR #2549, opened 2026-09-08
Hosted alternative to qdox plugin for maven-xbean-plugin

mattrpav:
This came about b/c dependabot found a qdox update that can't be merged

1. QDox is a plugin to maven-xbean-plugin and is used to scan our classes to generate the Spring handlers from java bean to Spring xsd. 
2. QDox has gone EOL altogether and xbean hasn't made plans to update to QDox v2.x (we are using 1.x).

The intent of this self-hosted module is to get what we need working without qdox while we move towards a permanent (possibly Spring-free) solution. The next step might be to move to a XSD-first approach, manage the Spring namespace handler in order to remove usage of the maven-xbean-plugin altogether (it is not widely maintained either).

QDox EOL statement:
ref: https://github.com/paul-hammant/qdox

Thoughts on the available options?


cshannon:
Are there changes we actually need to make that require us to copy and host it vs just using the exiting version? The alternative (as recommended by the link you posted) would be to migrate to something else. That could be a better solution than simply copying something that we now have to maintain, especially if we don't plan to maintain it.

mattrpav:
@cshannon good question. I was not able to find any alternatives to using qdox. XBean is a novel way to manage config and wiring all-at-once while using Spring for activation, but it is not widely utilized any longer.

I see this as a short-term transition step as we could then move off of xbean-plugin altogether, but it requires a little coordination while maintaining spring xml backwards compatibility.

Issue #2140, opened 2026-06-23
Improve KahaDB extensibility


mattrpav:
Currently, KahaDB lacks the ability to be extended due to usage of package-private. 

By modifying to using protected, users will be able to extend KahaDB in key areas.

PR #2550, opened 2026-09-09
[#2140] Widen MessageDatabase index-update points for subclasses

mattrpav:
This change opens the possibility to extend KahaDB for performance improvements. Making architecture changes to KahaDB directly is considered risky, as KahaDB is very stable.

By supporting a subclassed KahaDB, developers and users may gradually rollout new KahaDB features and capabilities. Specifically, this relates to improving KahaDB throughput by being able to leverage modern JDK capabilities -- specifically Virtual Threads and aio.

### DIFF
--- a/activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/MessageDatabaseExtensibilityTest.java
+++ b/activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/MessageDatabaseExtensibilityTest.java
@@ -1,0 +1,127 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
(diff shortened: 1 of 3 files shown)


cshannon:
I am trying to think if there is any clever/better way to make sure that subclasses are calling methods correctly after calling hte index lock but I don't really think so. If we are going to open up things like this more then extending classes will have to be sure to correctly grab the locks

activemq-kahadb-store/src/test/java/org/apache/activemq/store/kahadb/MessageDatabaseExtensibilityTest.java
```diff
@@ -0,0 +1,127 @@
...
+
+        @Override
+        protected KahaDBStore createStore() {
+            store = new CountingStore();
+            return store;
+        }
+    }
+
+    @Test
+    public void adapterSuppliedStoreSubclassObservesIndexUpdates() throws Exception {
```
cshannon, line 75: nit: I'd rename this to `testAdapterSuppliedStoreSubclassObservesIndexUpdates` to match convention

mattrpav: Done

PR #2553, opened 2026-09-09
Return safe fallback from BrokerView read-only getters instead of throwing when broker not yet started

### LINKED ISSUE #2551: Jolokia/JMX queries fail with IllegalStateException on slave broker when startAsync=true
## Problem

When a broker is configured with `startAsync=true` and is running as a slave (blocked waiting on the shared/exclusive lock), Jolokia queries against the Broker MBean fail with:

```
java.lang.IllegalStateException: Broker is not yet started.
```

Browsing the same MBean directly in JConsole shows all attributes correctly, including `Slave=true`.

## Root cause

In `BrokerView.java`, most attribute getters are safe to call before the `ManagedRegionBroker` object exists — for example `isSlave()` reads directly from `BrokerService` and never touches it. However, a few getters — `getBrokerId()`, `getBrokerName()`, plus several destination/statistics getters — call `safeGetBroker()`, which throws `IllegalStateException("Broker is not yet started.")` while the slave is still waiting on the lock.

Since Jolokia reads all attributes on an MBean in a single batch request, that one exception aborts the entire response, so attributes that would have succeeded independently (such as `Slave`) never come back either.

## Environment

- `startAsync=true`
- Broker running as slave, blocked on lock acquisition

## Suggested fix

Have the read-only getters in `BrokerView` return a safe fallback instead of throwing when the `ManagedRegionBroker` isn't set yet:
- broker name falls back to the configured name on `BrokerService`
- broker ID returns `null`
- counts return `0`
- lists return empty

Getters that actually mutate broker state (add/remove destinations, durable subscriptions, stats reset) should keep throwing, since those genuinely can't run yet.

jbonofre:
## Summary

- Fixes the Jolokia/JMX `IllegalStateException` reported in #2551: on a slave broker running with `startAsync=true`, querying the Broker MBean over Jolokia failed even though attributes like `Slave` were perfectly readable, because Jolokia reads all attributes in one batch and a single throwing getter aborted the whole response.
- `BrokerView`'s read-only attribute getters (`getBrokerId()`, `getBrokerName()`, destination/subscriber/producer lists, and the destination statistics counters) now return a safe fallback (`null`/`0`/empty array, broker name falls back to the configured name on `BrokerService`) instead of throwing via `safeGetBroker()` when `ManagedRegionBroker` hasn't been set yet.
- Getters that mutate broker state (`resetStatistics()`, `enableStatistics()`, `disableStatistics()`, destination add/remove, durable subscriber create/destroy) are unchanged and still throw `IllegalStateException` through `safeGetBroker()`, since they genuinely can't run before the broker is up.

## Test plan

- [x] `mvn -pl activemq-broker -am compile` succeeds
- [ ] Manual verification: start a slave broker with `startAsync=true` and confirm a Jolokia bulk attribute read on the Broker MBean succeeds (including `Slave=true`) while the broker is still waiting on the lock

Fixes #2551

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
@@ -44,6 +44,8 @@ public class BrokerView implements BrokerViewMBean {
 
     private static final Logger LOG = LoggerFactory.getLogger(BrokerView.class);
 
+    private static final ObjectName[] NO_OBJECT_NAMES = new ObjectName[0];
+
     ManagedRegionBroker broker;
 
     private final BrokerService brokerService;
@@ -65,12 +67,18 @@ public class BrokerView implements BrokerViewMBean {
 
     @Override
     public String getBrokerId() {
-        return safeGetBroker().getBrokerId().toString();
+        if (broker == null) {
+            return null;
+        }
+        return broker.getBrokerId().toString();
     }
 
     @Override
     public String getBrokerName() {
-        return safeGetBroker().getBrokerName();
+        if (broker == null) {
+            return brokerService.getBrokerName();
+        }
+        return broker.getBrokerName();
     }
 
     @Override
@@ -135,27 +143,42 @@ public class BrokerView implements BrokerViewMBean {
 
     @Override
     public long getTotalEnqueueCount() {
-        return safeGetBroker().getDestinationStatistics().getEnqueues().getCount();
+        if (broker == null) {
+            return 0;
+        }
+        return broker.getDestinationStatistics().getEnqueues().getCount();
     }
 
     @Override
     public long getTotalDequeueCount() {
-        return safeGetBroker().getDestinationStatistics().getDequeues().getCount();
+        if (broker == null) {
+            return 0;
+        }
+        return broker.getDestinationStatistics().getDequeues().getCount();
     }
 
     @Override
     public long getTotalConsumerCount() {
-        return safeGetBroker().getDestinationStatistics().getConsumers().getCount();
+        if (broker == null) {
+            return 0;
+        }
+        return broker.getDestinationStatistics().getConsumers().getCount();
     }
 
     @Override
     public long getTotalProducerCount() {
-        return safeGetBroker().getDestinationStatistics().getProducers().getCount();
+        if (broker == null) {
+            return 0;
+        }
+        return broker.getDestinationStatistics().getProducers().getCount();
     }
 
     @Override
     public long getTotalMessageCount() {
-        return safeGetBroker().getDestinationStatistics().getMessages().getCount();
+        if (broker == null) {
+            return 0;
+        }
+        return broker.getDestinationStatistics().getMessages().getCount();
     }
 
     /**
@@ -163,8 +186,11 @@ public class BrokerView implements BrokerViewMBean {
      */
     @Override
     public long getAverageMessageSize() {
+        if (broker == null) {
+            return 0;
+        }
         // we are okay with the size without decimals so cast to long
-        return (long) safeGetBroker().getDestinationStatistics().getMessageSize().getAverageSize();
+        return (long) broker.getDestinationStatistics().getMessageSize().getAverageSize();
     }
 
     /**
@@ -172,7 +198,10 @@ public class BrokerView implements BrokerViewMBean {
      */
     @Override
     public long getMaxMessageSize() {
-        return safeGetBroker().getDestinationStatistics().getMessageSize().getMaxSize();
+        if (broker == null) {
+            return 0;
+        }
+        return broker.getDestinationStatistics().getMessageSize().getMaxSize();
     }
 
     /**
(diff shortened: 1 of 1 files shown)


cshannon:
Instead of the null check here, it might make more sense to check the state of the broker and see if it's the active broker or the standby. If it's the active broker, and the broker object is null, that would mean a real error and the exception should be thrown.

jbonofre:
@cshannon on slave the broker bean is null. I can do an indirect accessor but I think null check is good enough for the slave case (the purpose here).

mattrpav:
I have a concern that changing the exception handling would be breaking for JMX clients that already handle the exception being thrown.

I’ll do a couple of checks

I prefer this clean/empty data approach over throwing exceptions, but it may need to wait to go into a major version.

jbonofre:
@mattrpav nothing is break, it just bypass the exception when broker is not there, the exceptions are still thrown as before on "active" broker.

jbonofre:
@cshannon the exceptions are still thrown when broker is not null (so on an active broker).

cshannon:
> @cshannon on slave the broker bean is null. I can do an indirect accessor but I think null check is good enough for the slave case (the purpose here).

I was thinking you could do something like this:

```java
@Override
public long getTotalConsumerCount() {
    if (isStandBy()) {
        return 0;
    }
    return safeGetBroker().getDestinationStatistics().getConsumers().getCount();
}

private boolean isStandBy() {
    return broker == null && brokerService.isSlave();
}
```

BrokerService should never be null (even if broker is)

jbonofre:
@cshannon I'm fine to do that but not sure it changes anything.

I've tested the `BrokerService`:
1. `slave` defaults to `true` and is reset to `true` on every stop.
2. It only flips to `false` inside `startAllConnectors()`.
3. `adminView.setBroker(managedBroker)` (this call makes `BrokerView.broker` non-null) happens in `doStartBroker()` which runs before `startAllConnectors()`.

So within a single `BrokerService` instance, `broker == null` implies `isSlave() == true` at every point at startup (there's no case where `broker` is null but `isSlave()` has already gone false.

It means:

```
isStandBy() = (broker == null) && brokerService.isSlave()
```

is correct but `&& isSlave()` term can never actually be false while the first is true, given the current ordering in `BrokerService`.

So:
- It changes no behavior
- It adds a dependency on the `slave` for a "guard" whose only real job is null-safety before calling `safeGetBroker()`.
- If an user ever reorders `startAllConnectors()`/`adminView.setBroker()` in `BrokerService` (unlikely, but possible), this guard behavior would silently change without anyone touching `BrokerView`.

I'm fine to add `isSlave()` check, even if I believe it's useless 😄

jbonofre:
@mattrpav that makes totally sense. Maybe we should target just 6.4.0 for this change (not 7)?

cshannon:
> @mattrpav that makes totally sense. Maybe we should target just 6.4.0 for this change (not 7)?

+1 for 6.4.0

I think this change is probably fine for something like 6.4.0. It also depends on if we want to use the empty data approach instead of error if broker is null no matter what (like the existing PR) or limit it to only if in slave mode. (the broker object is not set until start is called so i'm not sure if it's possible for the broker to be null briefly during initialization). I suggested to do that extra check but maybe we just leave things as is, not sure.

cshannon:
> @cshannon I'm fine to do that but not sure it changes anything.
> 
> I've tested the `BrokerService`:
> 
> 1. `slave` defaults to `true` and is reset to `true` on every stop.
> 2. It only flips to `false` inside `startAllConnectors()`.
> 3. `adminView.setBroker(managedBroker)` (this call makes `BrokerView.broker` non-null) happens in `doStartBroker()` which runs before `startAllConnectors()`.
> 
> So within a single `BrokerService` instance, `broker == null` implies `isSlave() == true` at every point at startup (there's no case where `broker` is null but `isSlave()` has already gone false.

Alright, based on that it's probably no point to add the check. As you stated it doesn't really buy us anything.

cshannon:
So after looking it over and thinking about it, I think the current version is fine for 6.4.0, but not a backport due to the change to empty vs throwing an exception. We can make a note in the release notes because I think this is overall a better way to handle it.

mattrpav:
I'm good with it going in 6.4.0. 

This type of return-instead-of-exception handling is a strict SEMVER/major version violation, but it is minor in impact and isolated to monitoring and not message flow. I think getting these types of things cleaned up in 6.x will allow us to be tighter with SEMVER in 7.x, so I'm good with being looser on SEMVER here.

Good fix!

activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
```diff
@@ -135,56 +143,83 @@ public void stopGracefully(String connectorName, String queueName, long timeout,
 
     @Override
     public long getTotalEnqueueCount() {
-        return safeGetBroker().getDestinationStatistics().getEnqueues().getCount();
+        if (broker == null) {
+            return 0;
```
mattrpav, line 147: How about '0' -> '0L' to tighten up code style?

PR #2554, opened 2026-09-09
StartCommand: keep waiting past the default start timeout unless a start exception was recorded

### LINKED ISSUE #2552: StartCommand exits with no logged error after 10-minute default timeout when a slave broker legitimately waits for the lock under startAsync=true
## Problem

A broker started via `./activemq start`, configured with `startAsync=true` and running as a slave, disappears after roughly 10 minutes with nothing logged.

## Root cause

`BrokerService.DEFAULT_START_TIMEOUT` is 10 minutes. `StartCommand` calls `broker.waitUntilStarted()` with that timeout, then throws if it returns false:

```java
if (!broker.waitUntilStarted()) {
    throw new Exception(broker.getStartException());
}
```

Before `startAsync` existed, this was never an issue, because `start()` itself blocked until the broker became master, so `waitUntilStarted()` returned almost immediately either way.

With `startAsync=true`, `start()` returns right away, and a slave broker can legitimately wait on the lock far longer than 10 minutes — that is expected behavior, not a failure. When the timeout is hit, `waitUntilStarted()` returns `false` with no start exception set, and `StartCommand` treats that as a startup failure and exits the process. Because there's no recorded exception, nothing is logged, which makes the process exit look silent and unexplained.

## Environment

- `startAsync=true`
- Broker running as slave, waiting on the lock beyond the default 10-minute start timeout

## Suggested fix

`StartCommand` should keep waiting as long as `broker.getStartException()` is `null`, and only throw/exit when there is an actual recorded start exception.

jbonofre:
## Summary

- Fixes #2552: a broker started via `./activemq start` with `startAsync=true`, running as a slave, disappeared after roughly 10 minutes with nothing logged.
- `BrokerService.DEFAULT_START_TIMEOUT` is 10 minutes, and `StartCommand` called `broker.waitUntilStarted()` with that timeout, throwing (and exiting the process) whenever it returned `false`. Before `startAsync` existed this was harmless, since `start()` itself blocked until the broker became master. With `startAsync=true`, `start()` returns immediately, and a slave can legitimately wait on the lock far longer than 10 minutes.
- `StartCommand` now loops on `waitUntilStarted()`, only throwing/exiting once `broker.getStartException()` is actually non-null. This can't spin indefinitely: `BrokerService.stop()` always records a start exception (`BrokerStoppedException`) before it ever marks the broker as stopped, so a real stop or startup failure still surfaces immediately.

## Test plan

- [x] `mvn -pl activemq-console -am compile` succeeds
- [ ] Manual verification: start a slave broker with `startAsync=true` and confirm the process stays alive past 10 minutes while waiting on the lock, and exits promptly with a logged error on an actual startup failure

Fixes #2552

### DIFF
--- a/activemq-console/src/main/java/org/apache/activemq/console/command/StartCommand.java
+++ b/activemq-console/src/main/java/org/apache/activemq/console/command/StartCommand.java
@@ -92,8 +92,11 @@ public class StartCommand extends AbstractCommand {
                 throw e;
             }
 
-            if (!broker.waitUntilStarted()) {
-                throw new Exception(broker.getStartException());
+            while (!broker.waitUntilStarted()) {
+                if (broker.getStartException() != null) {
+                    throw new Exception(broker.getStartException());
+                }
+                // still starting (e.g. a slave waiting on the lock under startAsync) - keep waiting
             }
 
             // The broker started up fine.  Now lets wait for it to stop...


activemq-console/src/main/java/org/apache/activemq/console/command/StartCommand.java
```diff
@@ -92,8 +92,11 @@ protected void runTask(List<String> brokerURIs) throws Exception {
                 throw e;
             }
 
-            if (!broker.waitUntilStarted()) {
-                throw new Exception(broker.getStartException());
+            while (!broker.waitUntilStarted()) {
```
cshannon, line 95: Just like my comment on the other PR, we should probably do an broker.isSlave() check here. The exception should be set, but in case something gets broken in the future this would add an extra check just to ensure we always fail for a primary/active broker just like today if it returns false.

Maybe:

```java
while (!broker.waitUntilStarted()) {
    if (broker.getStartException() != null || !broker.isSlave()) {
        throw new Exception(broker.getStartException());  
    }
}
```

cshannon (changes requested):

PR #2555, opened 2026-09-09
[6.3.x] Return safe fallback from BrokerView read-only getters instead of throwing when broker not yet started

### LINKED ISSUE #2551: Jolokia/JMX queries fail with IllegalStateException on slave broker when startAsync=true

jbonofre:
Backport of #2553 to `activemq-6.3.x`.

## Summary

- Fixes the Jolokia/JMX `IllegalStateException` reported in #2551: on a slave broker running with `startAsync=true`, querying the Broker MBean over Jolokia failed even though attributes like `Slave` were perfectly readable, because Jolokia reads all attributes in one batch and a single throwing getter aborted the whole response.
- `BrokerView`'s read-only attribute getters (`getBrokerId()`, `getBrokerName()`, destination/subscriber/producer lists, and the destination statistics counters) now return a safe fallback (`null`/`0`/empty array, broker name falls back to the configured name on `BrokerService`) instead of throwing via `safeGetBroker()` when `ManagedRegionBroker` hasn't been set yet.
- Getters that mutate broker state (`resetStatistics()`, `enableStatistics()`, `disableStatistics()`, destination add/remove, durable subscriber create/destroy) are unchanged and still throw `IllegalStateException` through `safeGetBroker()`, since they genuinely can't run before the broker is up.

Clean cherry-pick of the `main` commit; `BrokerView.java` was identical between `main` and `activemq-6.3.x` before this change.

Fixes #2551

### DIFF
--- a/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
+++ b/activemq-broker/src/main/java/org/apache/activemq/broker/jmx/BrokerView.java
@@ -44,6 +44,8 @@ public class BrokerView implements BrokerViewMBean {
 
     private static final Logger LOG = LoggerFactory.getLogger(BrokerView.class);
 
+    private static final ObjectName[] NO_OBJECT_NAMES = new ObjectName[0];
+
     ManagedRegionBroker broker;
 
     private final BrokerService brokerService;
@@ -65,12 +67,18 @@ public class BrokerView implements BrokerViewMBean {
 
     @Override
     public String getBrokerId() {
-        return safeGetBroker().getBrokerId().toString();
+        if (broker == null) {
+            return null;
+        }
+        return broker.getBrokerId().toString();
     }
 
     @Override
     public String getBrokerName() {
-        return safeGetBroker().getBrokerName();
+        if (broker == null) {
+            return brokerService.getBrokerName();
+        }
+        return broker.getBrokerName();
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


jbonofre:
Target 6.4.0 only.

PR #2567, opened 2026-09-14
JDK 27 / post-quantum TLS support

mattrpav:
Notes:

- M-R jar used to isolate supported features to jdk27
- JDK 27 puts post-quantum algorithms at the front of the list by default
- JDK 27 required for ML-KEM and other post-quantum algorithm support
- JDK 27 EOL's legacy keystore types and denies them (hence the unit tests keystore update required)
- JDK 27 now allows specifying algorithm group names (aka 'NamedGroups'), so all the quantum-ready groups can be added vs having to sift through algorithm names in a list.
- JMX operations added for the specific values related to post-quantum configuration

```
public String[] getNamedGroups();
public String[] getSignatureSchemes();
public boolean isRequirePostQuantumKeyExchange();
```

Note: 27-ea used as identifier until setup-java update for Temurin 27 ga.


cshannon:
As discussed on the dev forum, this seems premature. I think this effort should wait until there is support in an LTS version (java 29), i don't think we need to be trying to support cutting edge features on non LTS java versions at this point as its very unlikely users are going to care and there's plenty of other stuff to work on that is more important (in my opinion)

cshannon:
I will also add i'm -1 in general to add anything that requires a non-LTS java version to build. We should stick to only requiring JDK 25 for the full build until 29 comes out.

mattrpav:
@cshannon I generally agree that we should not prioritize JDK features— esp preview ones that may or may not make it to stable. 

This is different — TLS support for post-quantum algorithms is going to be back-ported all the way back to JDK 8. 

JDK 25 support for PQ is next month.

cshannon:
> @cshannon I generally agree that we should not prioritize JDK features— esp preview ones that may or may not make it to stable.
> 
> This is different — TLS support for post-quantum algorithms is going to be back-ported all the way back to JDK 8.
> 
> JDK 25 support for PQ is next month.

If support is added to JDK 25 then that is fine, but we can just wait until that happens before adding this. I'm -1 on bumping the build to require a non-LTS build.

mattrpav:
Sounds good. I'll split this to two separate PRs-- one for JDK 27 compile validation and JDK25-based support for PQ.

I have PQ support targeting v6.5.0 which should land around when PQ in JDK 25 comes in.

mattrpav:
Replaced by [#2594] and [#2595]

Issue #1997, opened 2026-05-07
jasypt version 1.9.3 is vulnerable to CVE-2022-22965


cloudtracer:
Can this dependency be updated? Its vulnerable to a a number of old CVEs some of which are rated high.

jbonofre:
Already updated: https://github.com/apache/activemq/blob/main/pom.xml#L72

cloudtracer:
Thanks for the fast response @jbonofre but 1.9.3 is a vulnerable version. You can see all the vulnerabilities with this dependency on mvnrepository : https://mvnrepository.com/artifact/org.jasypt/jasypt-spring4/1.9.3

jbonofre:
Oh sorry, I thought you asked to update to 1.9.3.

jbonofre:
Let me do an analyse to see if the CVEs actually applicable in ActiveMQ context.

davhaus:
Does applicability of a given CVE matter? jasypt is EOL as of last fall with the last release published 7 years ago. That alone is a critical finding in some settings and can only be addressed by outright replacement.

https://github.com/jasypt/jasypt:

<img width="676" height="52" alt="Image" src="https://github.com/user-attachments/assets/68d19ab3-469c-438e-97d3-422163847f6e" />
...

<img width="1050" height="193" alt="Image" src="https://github.com/user-attachments/assets/27fea431-a7a7-4e3f-84bf-d01050601e34" />

mattrpav:
I’m in favor of replacing it. jasypt is such a thin wrapper anyway. More work to maintain it than a simple hash utility

davhaus:
Still relevant.

mattrpav:
I'll try to pull this into 6.4.0 and be done w/ jasypt.

PR #2580, opened 2026-09-18
[#1997] jasypt replacement

mattrpav:
- [x] Replace EOL jasypt with JDK-only encryptor/decryptor support
- [x] Support forward-only usage of previous password hashes (log WARN when used)
- [x] Add new stronger crypto by default
- [x] Add new encrypted Spring PropertySourcesPlaceholderConfigurer as Spring's PropertyPlaceholderConfigurer is deprecated


cshannon:
This is a massive PR and given that there has been 0 discussion on it and if we should even do this, there is no way it's going into 6.4.0

mattrpav:
@cshannon no problem kicking this to 6.5.0. Discussion goes back to the end-user reported issue [#1997]

Issue #2583, opened 2026-09-19
Bug: queueSize may show non-zero when queue is empty


mattrpav:
Tricky scenarios to run down that can lead to queue size.

1. A race condition during failover reconnect and the second message hits the store first. The second will log: "POST COMMIT FAILED/Store COMMIT FAILED"
2. During message expiration, an IOException may throw and never retry.3. 
3. A race condition that logs "Duplicate message add attempt rejected"

PR #2584, opened 2026-09-19
[#2583] Duplicate message publish during failover leads to incorrect queue size metric

mattrpav:
Race condition during addMessage, where in a transaction or asyncSend scenario, the second addMessage may 'win', leading the first to fail and lead to incorrect queue size.

This change checks the asyncTaskMap for 'async' adds to catch duplicate before it reaches the store.


activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java
```diff
@@ -526,6 +526,28 @@ public void removeAsyncMessage(ConnectionContext context, MessageAck ack) throws
...
+            // A synchronous add (no future assigned yet) can race a pending
+            // asynchronous add of the same message id (concurrentStoreAndDispatch):
+            // the async task still holds the journal/index update while its message
+            // has already been counted and dispatched from the cursor cache. If the
+            // sync add wins the index it is assigned a real sequence, evades the
+            // duplicate suppression in the cursor add path, and inflates the
+            // destination statistics. Reject it here as the duplicate it is; the
+            // -1 marker is handled by the existing suppression in
+            // Queue.doPendingCursorAdditions.
+            if (isConcurrentStoreAndDispatchQueues()
```
cshannon, line 538: Is it possible for multiple async adds to happen and cause a race condition during failover? 

It looks like the code just calls put() on the asyncTaskMap and never checks if an existing entry is in the map.

If we are adding this check here, don't we also need to add the same check [here](https://github.com/mattrpav/activemq/blob/e3d559b7936d68e923f37926724bbdf890d69392/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java#L321) and [here](https://github.com/mattrpav/activemq/blob/e3d559b7936d68e923f37926724bbdf890d69392/activemq-kahadb-store/src/main/java/org/apache/activemq/store/kahadb/KahaDBStore.java#L335) ?

mattrpav: @cshannon L544 checks for previous entry.

Good call on adding checks in the other two spots, I'll look at those call spots.

edit: Those call sites are already under lock, so I believe it is as easy as checking for presence and then logging the duplicate add attempt and swallowing vs throwing an exception.  

I think the idempotent behavior is a valid approach in this case. Thoughts?

mattrpav: @cshannon I don't think we need to add checks in those other two locations, b/c the cursor audit catches dupe add attempts.

The core scenario in this fix is the async v sync race condition where the cache can get full and it converts the failover send to sync. 

The cursor audit catches async v async races already. New test pushed up with parameterized useCache (T|F) to validate both straight-line paths as well as the async-to-sync path already tested on this PR.

PR #2585, opened 2026-09-19
[#2583] Transacted fire-after-commit first exception exits loop prematurely leading to incorrect queue size

mattrpav:
During a transaction fire-after-commit. The first exception throws and other synchronizations do not get called, leading to incorrect queue size.

This change catches the exception, allows the other synchronizations to complete, then throws the first exception


activemq-broker/src/main/java/org/apache/activemq/transaction/Transaction.java
```diff
@@ -134,24 +134,59 @@ protected void fireBeforeCommit() throws Exception {
...
+        // executing (e.g. the destination synchronization that maintains the
+        // destination statistics) - run them all and rethrow the first failure.
+        Throwable firstException = null;
         synchronized(synchronizations) {
             for (Iterator<Synchronization> iter = synchronizations.iterator(); iter.hasNext();) {
                 Synchronization s = iter.next();
-                s.afterCommit();
+                try {
+                    s.afterCommit();
+                } catch (Throwable t) {
```
cshannon, line 147: Shouldn't we just catch Exception here? I don't think we want to catch Throwable/Errors.

If we get an OutOfMemoryError (for example) that's fatal, we just need to throw and be done. Continuing to execute does not make sense as an Error is going to cause the broker to halt.

cshannon (changes requested):

activemq-broker/src/main/java/org/apache/activemq/transaction/Transaction.java
mattrpav: I don't like catching Throwable either, but leaving other  synchronizations hanging out in space leads to incorrect accounting and preventing them from completing.

cshannon: I think you are missing my point...catching Error makes no sense because the broker is going to be going down anyways. If there is an OutOfMemoryError the broker is dead, we should not be trying to complete anything else so we need to abort and not try to keep processing.

mattrpav: That's the tricky part -- The broker is not guaranteed to go down on IOException -- it may recover. The IOException will got to the IOExceptionHandler and may resume based on configuration.

JDBC for example, can retry and reconnect and the broker go back to being active.

cshannon: > That's the tricky part -- The broker is not guaranteed to go down on IOException -- it may recover. The IOException will got to the IOExceptionHandler and may resume based on configuration.
> 
> JDBC for example, can retry and reconnect and the broker go back to being active.

Did you mean to post this on #2590 ?

cshannon: If you meant to post that on this issue...I'm not saying we can't catch IOException. I said in my first comment we can catch Exception.

I'm saying it's a bad idea to catch Throwable for the reasons I already explained.

mattrpav: Ugh.. yes. My brain is fried.

cshannon: I'm not sure even catching Exception and continuing is ideal but there may not really be a choice. I'm a little concerned there could be some exceptions that happen that we would want to abort the chain, but I'd have to think it through. It's probably fine because the transaction is already durable/committed at that point so it won't be rolled back either way. We would def not want to execute if the transaction was not going to finish but this is already post-commit as pointed out in the comment.

PR #2589, opened 2026-09-21
[6.2.x] Add Spring Schema for 6.2.10

jbonofre:


mattrpav (approved):
LGTM

edit: Should we add a step in the release guide to remind that this needs to be updated for every version?

jbonofre:
@mattrpav that's a good idea. It should be a "post release" step (at same time as updating the version).

mattrpav:
[#2596] provides going forward automation to spring schema version

PR #2590, opened 2026-09-21
[#2583] Temporary expiration failure can lead to incorrect queue metrics accounting

mattrpav:
An IOException when attempting an expiration, incorrectly leaves the message count but registers the expiredCount.

This change does two things:

1. Add a natural retry to message expiration if it fails on IOException
2. Count the expiration and fire the advisory after the expiration-to-store is successful


cshannon:
I don't really understand why we are retrying on IOException. That's usually a pretty fatal exception, retrying seems bad and I could see it getting into a loop. 

What possible IOException could happen that is retriable? The broker will just shut down on IOException in many cases. So having a blanket retry doesn't seem like a great idea.

cshannon:
And yes I understand the failover case being temporary, but we can't just solve that case. This needs to also consider non failover when IOExceptions are fatal.

mattrpav:
@cshannon the accounting is incorrect. If the write fails, the queue size and expiry count are incorrect when this occurs. This change addresses the accounting by moving the expiredCount to after the write is successful, and by allowing for retrying of the expiry, the queue size metric won't be incorrect -- queueSize > 0 when there are no messages to be processed.

The IOException is already being caught today. The new log may be better phrased to say 'expiry _may_ be retried'.

IOExceptionHandler decision to move broker into inactive mode is orthogonal.

PR #2594, opened 2026-09-21
Add JDK 27 compile to the build

mattrpav:
- Updated deprecated and removed legacy keystore types from tests

This change adds JDK 27 compile support to the build. JDK 27 is non-TLS, but will help catch deprecation and any impacting JDK changes before the next LTS release.


mattrpav:
*Pending setup-java support for JDK 27 (non-ea)

